PR #48, opened 2019-09-26
Autolock rail issue. S1P S2P

tatarize:
Autolock uses IS2P to lock rail but did so after commands that called S1P. That should be pointless as ending with S2P should do that automatically.


tatarize:
This does not work

PR #90, opened 2019-12-08
svg elements modification

tatarize:
Current modification towards gettings SVG elements. Some remaining problems which should be readily fixable.


tatarize:
- [ ] Image Raster Step does not properly affect the raster. It should increase the size if native, but if non-native it should remain that size and not be actualized.
- [ ] Tree menu does not notify people of non-actual image sizes or force their actualization.
- [ ] Rotation by slight angles and scales can cause non-center raster operations.
- [ ] Disable Color Define (push to next versions).
- [ ] Move SVG elements as group is not working, moves each individually.

Requires laser retest.

Issue #21, opened 2019-09-07
Proper transform interfacing.


tatarize:
The transforms shouldn't use a bunch of cryptic menu items but rather should just let you adjust the element in place. So you can drag it to the size and rotation you want. Perhaps with a button for flip horizontal or vertical.

tatarize:
This is added and complete in 0.6.0

tatarize:
This issue was corrected in 0.6.0. Which is now in full release.

PR #93, opened 2019-12-22
Exception raised when rotating empty design

jkramarz:
When "Design->Transform->Rotate tau/4" is clicked with empty project, following exception is being raised:
```
  File "meerk40t/MeerK40t.py", line 516, in transform_rotate_right
    self.project.menu_rotate(0.25 * tau)
  File "meerk40t/LaserProject.py", line 332, in menu_rotate
    position = self.selected.center
AttributeError: 'NoneType' object has no attribute 'center'
```
in
https://github.com/meerk40t/meerk40t/blob/57c4c3c957bece764ea27cdab91ed1f0a511a759/LaserProject.py#L327-L339

Is it OK to fix it like that?

### DIFF
--- a/LaserProject.py
+++ b/LaserProject.py
@@ -328,8 +328,10 @@ class LaserProject:
         self.validate()
         if position is not None:
             self.set_selected_by_position(position)
-        else:
+        elif self.selected is not None:
             position = self.selected.center
+        else:
+            return
         if self.selected is not None:
             self.validate()
             for e in self.selected:


tatarize:
That will actually fix that correctly. I would check if self.selected is None earlier and just exit the function up higher rather than throw it in the if elif else structure since if self.selected is None, it basically makes any operation impossible. And it makes the understanding of the flow a bit stranger. Or go back further and either disable that button or hide the button if the selection is None.

But, I'm fine with that. 

I've mostly finished up the svg.elements project that I needed to make this stuff work really well, and it's been a while. But, I'm also working hard on getting rid of the various problems of getting the code to cooperate with the new ways I have to do things, and to test all the functionality out. It's still missing a feature to properly actualize pixels, but once that is in, I should be able to release my newer version. Which coincidentally got rid of that entire menu for rotation.

I'm thinking I should #21 up since that's a much more reasonable way to do all the transforms. Just give it some widgets to do the manipulations. At most it's drawing a bounding box just now, which isn't too bad but isn't that great either.

I'm hoping to finish the update, which basically is a working version with just svg.elements working. In the next day or two. So this fix is going to get wiped out shortly.

But, it's a proper fix and I'm happy to accept it.

PR #97, opened 2019-12-27
Updated MeerK40t UI for Linux

tatarize:


tatarize:
Replacing the scene panel allows linux to scroll and pan correctly, but still no visualization of the objects.

PR #102, opened 2019-12-31
Advanced manipulation

tatarize:


tatarize:
This version currently has the start of:

* widget based scale and rotation manipulations, with positional guides
* translations to other languages
* automatic conversion of text into vector shapes with `freetype-py`

None of which actually work. The positional guides works but it doesn't allow mouse scaling. The translations aren't being invoked, through the dialogs are being run through the translation routine and a messages.po is generated an in `.\locale`.  And the automatic conversion of text creates a blob of not text vector shapes with no semblance a reasonable approximation of text even with the `Vera.ttf` file stored locally, though not included.

None of these failures should get in the way.

tatarize:
Added in Translations to other languages. I don't speak any other languages so my French translation as added is likely really bad from the internet guesses. The main thing is that if somebody *wanted* to do a translation, they totally could do a translation.

tatarize:
- Added a bunch of tree menu changes. 
- Changed the text operation to remove rather than convert. (push to later version).
- Corrected transparency raster bug.  #105.

tatarize:
Still remaining:

* Non-implemented new functions in JobInfo, aren't disabled or implemented.
* Positional guides look like they should work, and they aren't implemented or disabled.
* Text items when selected have zero size and are at the origin.
* Beep function added to elements and operations should take priority and cover the beep when job is done functionality. Spooler no longer needs to force its own beep on empty.

tatarize:
I think this should be it. Needs direct testing. But, back to where it kinda should be.

Assuming there are no absolute fail bugs it should be good enough to push out the version.

joerlane:
My OS X application doesn't home the laser at the end of the job. Sending the K40-Wait Abort had no effect. Home at end of job was working last night; but has been hit or miss during the 0.2.4 development.

The application does quit cleanly now (no longer crashing). I've run a few jobs and they looked good. Then it started failing to open SVG files; and I get two errors (this was working 10 minutes ago).... This issue appears less than consistent. Sometimes it will load a file; other times these errors: Furthermore; if a file is successfully loaded, and I load another file and get errors; the first one disappears from the tree but not the stage drawing.

  File "wx/core.py", line 3360, in <lambda>
  File "meerk40t/MeerK40t.py", line 439, in on_elements_update
  File "meerk40t/MeerK40t.py", line 591, in refresh_tree_elements
  File "meerk40t/MeerK40t.py", line 574, in add_element
  File "meerk40t/LaserRender.py", line 90, in make_raster
  File "meerk40t/LaserRender.py", line 127, in set_brush
wx._core.wxAssertionError: C++ assertion "retval != __null" failed at /Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/osx/carbon/graphics.cpp(116) in wxMacCreateCGColor(): 

  File "meerk40t/LaserRender.py", line 31, in render
AttributeError: 'LaserNode' object has no attribute 'draw'

  File "wx/core.py", line 3360, in <lambda>
  File "meerk40t/MeerK40t.py", line 771, in update_buffer_ui_thread
  File "meerk40t/MeerK40t.py", line 1186, in on_draw_scene
  File "meerk40t/LaserRender.py", line 41, in render
  File "meerk40t/LaserRender.py", line 142, in draw_path
  File "meerk40t/LaserRender.py", line 127, in set_brush
wx._core.wxAssertionError: C++ assertion "retval != __null" failed at /Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/osx/carbon/graphics.cpp(116) in wxMacCreateCGColor(): 


Double click path object on Debian Pi throws error: (Same on Mac OS X)

File "meerk40t/MeerK40t.py", line 534, in on_item_activated
File "meerk40t/Kernel.py", line 343, in open_window
AttributeError: ;ElementProperty' object has no attribute 'set_project'

I think a few of those are absolute fails.

tatarize:
@joerlane Okay,, I think I got the Brush thing. It objects to having None color. So in that even I'll use Transparent brush from the defaults.  And found the Elements thing when I tested it.

tatarize:
Corrected the autohome thing too.

joerlane:
I can't get an SVG load to fail anymore; and I had some files that were pretty consistent before. I can also access Element Properties for both paths and the project (on both Mac and Pi); so that looks good too. 

No beep; and no home on end with my Mac job. Spooler is empty; K40-Wait Abort doesn't kick it. Double checked my preferences for those settings. Those check boxes no longer appear in the burn windows; though it might be a UI glitch on Mac/Linux? It was both homing on end and beeping last night.

joerlane:
I was still typing and waiting for a job to finish there... Pulled again; and now the Beep and Home on end work again.

tatarize:
If you don't find anything. Go ahead and merge the pull request. It'll take fixing the conflict with the readme fix somebody else offered up.

PR #113, opened 2020-01-18
0.3.1

tatarize:
* Should focus more on fixes and features that already exist.


joerlane:
Looks good. Mac OS X now allows the jogging of the head cursor via arrow keys. Number settings work as expected as well.

PR #118, opened 2020-01-19
0.3.2

tatarize:
Saved projects.


tatarize:
* Save, needs to local store a name for the save file or default to save as, clearing that value on "new".
* It should save svgs with text, and with text only (for use with inkscape cheating).
* Should maybe save cut information as its own block assuming that becomes a thing. (can be pushed).

tatarize:
Added a bunch of sorely needed FPS limiting and quality control for framerates.

PR #128, opened 2020-01-26
0.4.0 Branch

tatarize:
Not working now, but will eventually rewrite all the backend code into something more modular and permissive of multiples.


tatarize:
- [ ] The scene reticle no longer moves.
- [ ] Hitting start controller directly crashes missing K40Controller as a key.
- [ ] Last Packet in controller is the same thing as Packet Text.
- [ ] Should criteria based on chip version number. Even if I don't add directly check the GUI for that.
- [ ] If it finds multiple devices that all match the correct criteria it should, if it fails to connect to the first matching device, connect to the second matching device.
- [ ] The USB log currently does not process correctly for the different subdriver elements. While the information of the state in the LibUsb driver may be better, it shouldn't just be broken.
- [ ]

joerlane:
Got this from Mac OS X; can't launch the 0.4.0 pull from about noon on Tues Jan 28: 

python3 MeerK40t.py 
/Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/common/stdpbase.cpp(62): assert ""traits"" failed in Get(): create wxApp before calling this
/Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/common/stdpbase.cpp(62): assert ""traits"" failed in Get(): create wxApp before calling this
/Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/common/stdpbase.cpp(62): assert ""traits"" failed in Get(): create wxApp before calling this
/Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/common/stdpbase.cpp(62): assert ""traits"" failed in Get(): create wxApp before calling this
/Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/common/stdpbase.cpp(62): assert ""traits"" failed in Get(): create wxApp before calling this
/Users/robind/projects/bb2/dist-osx-py37/build/ext/wxWidgets/src/common/stdpbase.cpp(62): assert ""traits"" failed in Get(): create wxApp before calling this
  File "/Users/jlane/Applications/meerk40t040/K40Controller.py", line 154, in __init__
    driver.open()
  File "/Users/jlane/Applications/meerk40t040/CH341LibusbDriver.py", line 388, in open
    raise ConnectionRefusedError
ConnectionRefusedError

  File "MeerK40t.py", line 134, in <module>
    pipe = K40Controller(project)
  File "/Users/jlane/Applications/meerk40t040/K40Controller.py", line 159, in __init__
    driver = CH341Driver(self.driver_index)
  File "/Users/jlane/Applications/meerk40t040/CH341WindllDriver.py", line 11, in __init__
    self.driver = windll.LoadLibrary("CH341DLL.dll")
NameError: name 'windll' is not defined

tatarize:
Ah. Yeah, if neither driver is around it doesn't handle the error correctly. It tried to use the libusb driver, which failed then it tried the windows fallback driver that also failed. And at no point handled the combination of both those methods failing.

tatarize:
For the operation layers note the drag and drop is pretty advanced.
* You can drag items from the elements to the operations.
* You can drag items from the elements to the operation-root for classification.
* Red stroked items get classified as a cut.
* You can right click and delete an operation.
* You can double click the operations to set their settings.

It's all still a bit kludgy.

The forced run of various required items thing is broken. So it won't actualize your image. And it won't rasterize your vector. And there's a few gaping holes. I'll get to fixing some more of the clear gaps then itemizing the bits it's clearly missing and bugs I made by breaking things, and hopefully move towards correcting them.

tatarize:
* Deleting elements needs to kill the operations depending on those elements. 
* Base speed of the objects, if they have non-default values must be given preference in the classification of those objects. 
* Color grouping. Mass set path speeds in operation tree. Etc.

tatarize:
* Spooler has a number of entries that are no longer filled. Type, Speed, Settings, aren't used. Time estimate is still a stub. Device lists `M` but that's not a thing.

tatarize:
* Try a test of the newly reconnected device after a power off without sending the paramode set switch. If it rejects all the packets according to the standard methods it was power cycled and lost registration.  This could also permit jogging the device without position updates to fix position in the case of a deregistration. But, detecting a power cycle is important and doable.

PR #135, opened 2020-02-07
0.5.0

tatarize:
The upgrade to fully integrate the new scheme required a bunch of changes that more or less broke everything. The backend fix for 0.4.0 works breaking the front end feels like another major version. And I broke things at a weird stage.

Fixing the aspects is coming back online rather slowly. And the new sort of hybrid LaserNode/presentation layer, tree thing is a pretty big upgrade in itself. But, getting all the parts to more or less cooperate seems to require it.


tatarize:
Feature freeze and complete. The goal should be to make a minimally working version and publish:

- [ ] Fully replicate the default operations expectations of SVG.
- [ ] Filled shapes should get a raster as well as a vector.
- [ ] Blue vector strokes should get engrave.
- [ ] Red vector strokes should get cut.
- [ ] All other vector strokes should get raster.
- [ ] Should permit rastering a vector object.
- [ ] Requires basic method for setting default engrave, cut, and initial raster speeds.
- [ ] Should respect the override values given in SVG file.
- [ ] Drag and drop should work between operations and elements trees.
- [ ] Dragging operations into each other should insert at placement.
- [ ] Dragging elements into each other should insert at that placement.
- [ ] Dragging elements from one operation to another should move them. 
- [ ] Dragging element from elements to operation should add that element to the ending operation.
- [ ] Deleting element from elements should delete that element from all elements.
- [ ] Deleting operation should delete the operation from operations.
- [ ] Dragging element into "elements" should create a new grouped element.
- [ ] Must permit rebuilding of the operations tree based on the state of the elements tree.
- [ ] Operation preprocessing steps must be properly and fully implemented again. Actualize image, rasterize vector, rotary. Etc.
- [ ] Icons for the operations dialog must work again.
- [ ] Image transformations must work.
- [ ] All menu items should be restored to their previous functional state.
- [ ] Menu Items in scene should work.
- [ ] Elements should move at the group level.
- [ ] Mouse down should select the hit element.
- [ ] Elements Property or equivalent should work. Perhaps only at the operations level. Or activation of element should activate the first relevant operation.

tatarize:
I'm biting a few bullets here.

* I do not have element groups. Rather operations on elements are all done by selection. This will likely be fixed after publishing.
* I do not permit images and vectors to be merged into a raster. This requires some code I haven't yet developed.
* I do not have proper visualization of the subelements within jobbed and spooled operations. This tends to mean actualization and rasterization is kinda done blind.  Fractionally rotated images can sometimes actualize wrong and if this happened it wouldn't be known until rastering.
* I don't have advanced per-object properties dialogs. In fact it all uses ElementProperties which only works on Operations and does not work on Elements except to change colors.

tatarize:
Raster step needs to be two different settings. It needs to be part of images to determine their underlying DPI and keep them at a native resolution. And they need to be part of the raster operation in order to determine what kind of scaling would be required when it makes an image. Since these properties are different and merely related in that if they equal no additional pixels need to be invented. They need two different settings and the relationships between them need to be better mapped out.

* Image, Set to Native.
* Needs actualization applies to SVGImage objects but the 'raster_step' value is applied to the RasterOperation.

And a couple bugs found within the process cannot be corrected easily unless this is finished.

tatarize:
File, New = Crash

tatarize:
* make_image triggers even with empty raster job.

* still seems to assign different speeds for the items in operations (different than the master operation); but those individual operation’s settings aren’t editable. Not sure which it respects; or why it’s assigning a different one in the first place to a regular .svg (not made in Meerk40t).

* I don’t see how to home when running Meerk40t from the command line. It runs the file, and then just sits over the last position. Is there a control command for that?

tatarize:
F8 path addition doesn't refresh the tree.

tatarize:
- [ ] Running image at step 1 when the image is enlarged at step 5 should actualize the image to be larger.

joerlane:
If a file is open; and the bed size in preferences is edited; the tree expands to fill the main window. The scene appears to be inaccessible after that. 

Preferences don't appear to be saving at all.

Rotary settings turn on every time the application is opened.

Unable to tab between elements in preferences, or element properties.

tatarize:
Remove the fstrings from debug fails in python <= 3.5 which isn't quite broad enough. While dropping python 2, this is simply missing in earlier python 3.

tatarize:
- [ ] Remove fstrings from debug code.
- [ ] Running image at step 1 when the image is enlarged at step 5 should actualize the image to be larger.
- [ ] Actualize Image produces incorrect bounds. Bounds are for the preactualized image.
- [ ] Remove delegate to keymap from tree.
- [ ] Truncate the max length of titles to ~100 characters. 
- [ ] Check if settings and especially device settings save in linux.

Issue #114, opened 2020-01-19
Generic Optimization of Laser Pathing and Cut Planner


tatarize:
While most vectorization solutions have an end goal of creating a proper vectorized image from a raster and place significant emphasis on making something that would be presentable as an vector image. There seems to be a simpler solution with regard to laser cutting. If we consider each pixel in a raster image to be a node. And our device can travel in the 8 acceptable octet directions we can merely answer the question: what path can visit all black nodes of the graph exactly 1 time? This is equal to the travelling salesman problem. And while we cannot solve that problem exactly we can get very good approximations by optimizing the naive Eulerian boustrophedonic path, with 2-opt.

This is similar #4 to but would help for other things like  #3. The main difference here is instead of doing raster lines, we do specific individual pixels, permitting non-rastermode connections. Some of these optimizations would already be done in rastering when we skip pointless white edges. Since we could view this as trimming an all white loop by taking the alternative path, of just moving down to the next line. In theoretical terms we'd be including all paths in circuit by, in every case possible, moving down to the next line then back up. If we connected the path start to the path end, we could also optimize the start position. Since we would have a weighting metric declaring that pure white loops cost 0 and having an actual effect of never using them.

We have a grid of white and black states with specific bounds of width and height (basically just a raster), then produce a path that does the naive back-and-forth path. Then the path we take is storable as a series of 3 bits for each node. NW, N, NE, W, E, SW, S, SE, while the state of this pixel is known via the raster, we won't always be firing the laser when we come back to the same point. So this data must also be stored in a 4th bit. The packed path data could also notably be seen as a kind of raw data type since it would pretty directly map to all the moves the laser could make. And optimizing that with 2-opt could be interesting since the Path data could produce the exact pathing commands and optimizing the path against that would permit the optimizing of arbitrary path data when the path merely crossed itself, it could travel the other way. This would only be for data at the same speed within the same block.

Equally this could do things like z-depth rasters where we strike the same place multiple times to create progressive depths without the ability to increase the laser power.

tatarize:
One of the more useful reasons to do this isn't simply optimization but planning. Since the raw formatting would be easy to manipulate like this it would also, make the core elements of a cut planner. You could certainly plan out the cuts being made based on particular criteria. Reducing movements is the best thing but also, reducing direction changes permitting switching of the starting point and there's some common issues with removing overlapping lines while also being able to cut inner shapes first.

PR #160, opened 2020-03-17
0.6.0 (Next Version)

tatarize:
Major changes in this project and criteria for completion:
* svgelements update fixes use of defs, use, and style tags and css classed objects. This should correct some loading errors and the use if Illustrator CSS classed svg data.
    * This needs to be fully tested against the W3C suite of tests as well as against Illustrator data and known failed objects.
* Threaded events and processes were moved into being scheduled. Using the device threads rather than their own processes.
    * Need to test throughput and physical devices to make sure things are properly sending about as fast as possible. And that the switch didn't incur any bugs.
* GRBL Emulator made to more consistent and functional.
    * GRBL Emulator must be tested to work with several different grbl control programs. Universal Gcode Sender. Lightburn. Visicut. LaserWeb.
* Ruida Emulator should do some cursory work and function like a ruida device.
    * Ruida Emulator should be tested with the two known Ruida control software RDWorks and Lightburn.
* Kernel/Device was changed significantly. This change makes devices more independent of the kernel and generally more customizable for each backend.
    * The changes here require a retesting of every feature in MeerK40t, since they constitute a fundamental breaking change. And demanded a change to 0.6.0
* Device Manager was made to launch a different main window for each device. And to launch the device manager if nothing was set to launch at the start.
    * Executive decisions on whether to launch with a default Lhystudios device needs to be made.
    * Testing of the new methods here as a transition from 0.5.2 causes no issues.
* RD Works, RD Files can now be natively opened.
    * This needs to be fully tested and fixed according. There were clear issues of commands not recognized. And other data getting garbled.
* Optimize cuts, inner first.
    * This needs to be tested against a lot of different test files. Some bugs were detected. And this may need to be made optional since the cut order can be defined through other means.
* Modifications of the CLI to delegate some of the functionalities to console. Some commands like -e for sending egv (lhymicro-gl code) directly to the controller was moved into console EGV towards the same end.
* Addition of the console and a large amount of commands. This allows for expert manipulation and testing and diagnostics from texted based commands, for which there are a great many of these.
    * These need to be more strenuously tested for validity of arguments.
    * These may need some better understanding and care and access to other systems.
* Addition of a CLI batch command to execute a file of console commands before starting.
* Link made between the Keymap and the console. So keymap commands properly run console commands and can be made to run anything console has control over, which is most things.
* Alias/Bind functionality was added. Where keymapped keys and command key sequences can be added to and bound. This includes keydown with automatic keyup alias binds with the use of `+` and `-` prefixes. As well as methods to loop, and end particular commands to trigger while a key is being pressed.
* A number of additional keybinds were added including most vector manipulations through numpad keys.
* Default Operations Values Now Save. These will be used for reclassification and persist between sessions.
    * All the save operation values must be tested especially the other values in Raster.
* Added in a proper widget system. So the drawn objects are self-contained and they each deal with their own mouse commands and can do so in different and unique ways. Most stuff is properly organized within those widgets without dumping a lot of unneeded code in the main window.
   * All the widgeting stuff needs to be tested and likely ensure that there is proper updating and reflection of the changes between various aspects.

Truncating features for the following aspects.
* Alternative backend devices should print relevant data based on the code and be changed to clearly note that they do not currently work.

### Feature Freeze
* Fix critical issues.
* Side projects are to be tabled. No Project Moshi. No Project Ruida.
* Features are pushed to next version.
* Devices other than Lhystudios are to be disabled.
* Half finished non-critical aspects are to be disabled. Group / Ungroup. Spooler Button. Control Button, Up/Down buttons in Device.
* Remaining bugs are to be fixed.
* SVG / CSS should be finished.
* GRBL Interpreter should be hooked up correctly.
* Test Everything.
* Correct Bugs.
* Test Everything.
* Correct Bugs. 

### Checklist
Features (Frozen), do next version:
Frozen features moved to #175 

Testing:
- [ ] TEST: Test DXF image with embedded image.
- [ ] TEST: Infinite scroll issue when moving operation-elements on the mac.

Bugs: Should fix.
- [x] Pausing while the controller is running allows resume controller but fails to fix after resume.

Finished:
- [x] GRBL has bug that a G1 command doesn't trigger even without a G0 command after it. But, the G0 is also down and program mode.
- [x] Bounding box for text object often lacks height and width. Maybe wrong for elements with middle or end anchors.
- [x] Dragging one operation on top of another does not reorder those items.
- [x] REMOVED: `camera_snapshot` and `camera_update` console commands should be checked for validity.
- [x] Manually actualizing the image doesn't prevent it from being needed in the jobs menu.
- [x] Initial device selection is still on and the device insists on being selected for first boot. This should default to the correct device type.
- [x] Menu clear all, can't really just call clear() on object anymore. Will clear rather than unregister objects.
- [x] Deleting an item from elements while it still exists in files allows that element to remain semiselected.
- [x] Deleting multiple items fails to remove those items from operations
- [x] Optimize cuts must be optional and faster. Should not require what can be a 20 minute operation on a very complex object.
- [x] Loading a file causes it to build the tree three times. 
- [x] Reverse layer order doesn't work for elements.
- [x] TEST: Robustly test Optimize Cuts. #167. 
- [x] TEST: Verify Device list works correctly on OSX/Linux early reports suggest sketchy.
- [x] TEST: Retest that the move_shift wasn't used to single-command move while the laser is on.
- [x] TEST: Test Version command transfers from 0.5.2 settings to 0.6.0 settings. Delete settings. Run 0.5.2 then run 0.6.0 and make sure there's no issues upgrading.
- [x] TEST: Must test updating between shared elements on two different scenes for two different devices.
- [x] TEST: Verify the interoperability between different device main windows.
- [x] Seems likely that failure to explicitly set d_ratio will fail to over-rule a previously set d_ratio. Push d_ratio_custom stuff to interpreters.
- [x] D_ratio persists do not default to zero wrongly.
- [x] D_ratio doesn't actually engage even when turned on expressly.
- [x] Acceleration doesn't fully work.
- [x] Control F on None Fill causes crash.
- [x] If the controller is in idle then the main button does not do 'hold' as expected.
- [x] Slightly rotated image actualized to wrong position.
- [x] Equal but no identical paths get co-selected. So all copies of the exact same paths are selected together. This should be done by `is` rather than equals. Only duplicate copies should select.
- [x] Classify elements doesn't work.
- [x] Raster classification should accept text objects. While rastering them requires some work, text should just draw on the bitmap and work.
- [x] SVG stroke floor. Beyond minimum amount.
- [x] Raster classification is weird, classifies a bunch of red-stroke and blue-stroke objects.
- [x] SVG/CSS Text SVG needs to be done accurately, even without a text-to-path function.
- [x] Font value needs to correctly parse.
- [x] Fixing font position broke, `text hello`, `scale 50`. Fix it better. Fix it right.
- [x] Device manager allows double booting. And all the crashes that clearly will cause.
- [x] Path property loads on double click but does not apply to the entire selection for the fill properties. Fails to work and fails to notify the user that it will not work. 
- [x] Scene selection click doesn't work.
- [x] Threading shutdown of TCP_RUN for grbl mode never stops causing hang if grbl emulator was used.
- [x] Crash bug loading dxf.image files.
- [x] DISABLED: `group` and `ungroup` in console are not useful or functional without functioning groups.
- [x] DISABLED: Spooler has a Start Job function that has no functionality. Since switching from direct threads.
- [x] DISABLED: Devices. Preferences may require a rethink. Hard-coded board type might be in error. Or each board may require its own preferences. (Turning off all devices other than lhystudios device)
- [x] DISABLED: Devices. Move Up/Down are unconnected to anything.
- [x] Arc/Curve failure TP_earrings test, 
- [x] Loading camera with a fail, stopped the refresh of the main window.
- [x] Without camera, export suffers attribute error. LhystudiosDevice has no last_signal
- [x] Removed Elements do not have their cached paths cleared so without delete they crash after finish.
- [x] When doing matrix operations with the keypad it should update the bounding box.
- [x] Unit changed in Settings doesn't trigger the change in the main window.
- [x] Right Click File, Reload. Fails to reload file.
- [x] Selection from tree doesn't trigger navigation update.
- [x] While not strictly a bug, during rasters it's possible to force jogs to happen. Really that should be disabled.
- [x] CSS stroke floor issue applied to thumbnail icons too.
- [x] OSX objects to fonts at size of zero.
- [x] Crashes within the scheduler causes the scheduler thread to end making all crash bugs critical when they should be caught.
- [x] Controller say USB in upper left.
- [x] Stroke Width doesn't seem to accurately function.
- [x] Home before jobs in preference.
- [x] Font objects are still off by a little amount across the board. They should be closer in position to their expected location.
- [x] CRITICAL! Raster loses steps and packets, fails to raster correctly.
- [x] Add basic `tspan` for svg.
- [x] Retest the wait_buffer_empty commands in adjustment tests, without lock.
- [x] Initial view does not start properly zoomed.
- [x] Lhy backend does not properly hold dataflow.
- [x] svgelements update is incomplete. Scaling is still completely broken.
- [x] svgelements `Use of <defs> tags.`
- [x] svgelements `Use of <use> tags.`
- [x] svgelements `Use of <style> tags.`
- [x] svgelements `CSS class objects,` eg: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text
- [x] svgelements `CSS type objects.`
- [x] Properly deal with the SVG visibility value. `visibility="hidden"` #168 
- [x] 164 test, raster rasters off screen wrongly.
- [x] Explicitly test second issue #164 (Illustrator CSS class)
- [x] reticule does not properly move. Even to the end part of the writing. 
- [x] Devices crashes with keyerror.
- [x] Devices does not display the current device.
- [x] Devices does not allow selection of alternative backends.
- [x] Devices, Add Device, Crash.
- [x] Devices. Delete Device. Crash.
- [x] Initial signals on listen do not seem to send. Signals that are not sent leave blank values.
- [x] Keymap needs to integrate with internal console send system.
- [x] Terminal, "grbl", "?" crashes.
- [x] Closing main-windows does not close all other windows.
- [x] Closing main window does not display shutdown channel in Shutdown dialog.
- [x] Autoclose shutdown does not work.
- [x] Changing languages in settings causes crash.
- [x] Selecting some view property like "hide reticule" caused failure to redraw view.
- [x] Change the about page to be agnostic about Li Huiyu's gender. Joerlane believe Li may be female. No point in misgendering.
- [x] Consider adding a widgeting/animation system. 
- [x] Settings don't seem to save.
- [x] Switching devices needs to launch a second main MeerK40t window.
- [x] CLI refers to kernel.device several times, none of that is valid.
- [x] CLI autorun fails because device threads are daemons. It dies instantly.
- [x] Finish and add #167 inside first optimization. Do this somewhat independent of #114. Since it's kinda mission critical.
- [x] Alias commands should break into multiple lines.
- [x] Text Dialog is faulty and should allow Text Editing.
- [x] Image resized, rotated 90 ccw, and translated was moved on actualize pixel in menu.
- [x] Wrongly saves laser position as a static element of the device. That is a recipe for disaster.
- [x] Invert mouse zoom does not work.
- [x] Keymap should allow add hotkey.
- [x] Keymap should accept keystroke for key entry.
- [x] Click in scene position on element does not select that element. Can only select in tree.
- [x] `channel open console` causes an infinite loop and crashes the terminal as it relays the channel open message to both and thus infinitely.
- [x] View menu hide images, crashed.
- [x] Flipping image with the drag edge selection widget, causes it to be unclickable as the right side is less than the left giving it negative space.
- [x] Mousewheel doesn't persist to the background widget and zoom if done within selectionwidget
- [x] Clicking off the tree to the scene, selects one part, but the failure to update makes the selection unable to be selected without deselecting and selecting again. Must sync between the two.
- [x] Add reset user elements to entire elements menu.
- [x] Rotary always on top. Falls behind main window.
- [x] View Zoom Functions crash.
- [x] Final Final Arms suffers performance issues in the test.
- [x] Click and drag for selection. Selects the object on the mouse-up but, the scene has changed the initial position which does not reflect that.
- [x] Duplicate double applies the current matrix. Causing a desync between the bounds and the actual object. With the bounds being correct but the object getting a double concatted matrix.
- [x] Items that are too small have no region to move, only resize. There are similar issues with very small and large items.
- [x] Correction to duplicate offset, actualize offset caused failure of make raster and makethumbnail.
- [x] 5.2 (holdover) Single capture for tree and scene hotkeys.
- [x] 5.2 (holdover) Do not break the function of the arrow keys in the tree menu.
- [x] 5.2 (holdover) Custom 1-5 hotkeys not to overwrite the hotkeys themselves.
- [x] Navigation Movement Issues.
- [x] Hold finished lockup.
- [x] Emergency Stop does not clear the currently held spool item causing it to run after abort is cleared.
- [x] View Zoom-out does not correct guides. These also zoom out. Zoom must zoom at the scene level not the interface level too.
- [x] Text box is lower right to position. And text is drawn upper_right.
- [x] Issue #171, DXF Bulges are not implemented.
- [x] Reify as a console command gets wonky. Somehow resets things and reshapes bounds wrongly.
- [x] Settings language change, forces the shutdown event and loads mainwindow with a kernel as the device.
- [x] Shutdown text window can suffer wx._core_Assertion error due to being called in a different thread than the gui thread.
- [x] Languages does not change correctly.
- [x] Consider adding a toolbar with a large number of disabled tools. Paralleling what would be expected of similar software. (Raised Issue)


tatarize:
Currently the svgelements code is a bit torn up. I'm trying to process them in a two stage method where the first run will make direct objects that are the correct class but will only have the initial attributes. This will include processing any use tags. Then a second pass will process the CSS, styling and context values. The older current method still exists and is merely duplicated but cannot deal with use contexts being in situ rather than where the use tag originally occurred.

The Terminal is added. Currently it has one command `grbl` which connects to the `grbl emulation` pipe. 

The GRBL emulation is currently in the works. I'm trying to get all the standard commands expected of a GRBL working. This is a work in progress as well. While plenty of naive code worked with the older system some more complex communications like Lightburn's would suffer rather massive gaps in the implementation. 

The terminal also needs to do things like watch the grbl communications even when it's being run through the laser-server. And obviously these commands can be heavily expanded as the need arises.

tatarize:
- [ ] Initial view does not start properly zoomed.
- [ ] Lhy backend does not properly hold dataflow.
- [ ] svgelements update is incomplete. Scaling is still completely broken.
- [ ] reticule does not properly move. Even to the end part of the writing. 
- [ ] Devices crashes with keyerror.
- [ ] Devices does not display the current device.
- [ ] Devices does not allow selection of alternative backends.
- [ ] Devices, Add Device, Crash.
- [ ] Devices. Delete Device. Crash.
- [ ] Devices. Preferences may require a rethink. Hard-coded board type might be in error. Or each board may require its own preferences.
- [ ] Devices. Move Up/Down are unconnected to anything.
- [ ] Initial signals on listen do not seem to send. Signals that are not sent leave blank values.
- [ ] Keymap needs to integrate with internal console send system.
- [ ] Terminal, "grbl", "?" crashes.
- [ ] Closing main-windows does not close all other windows.
- [ ] Closing main window does not display shutdown channel in Shutdown dialog.
- [ ] Autoclose shutdown does not work.
- [ ] Changing languages in settings causes crash.
- [ ] Selecting some view property like "hide reticule" caused failure to redraw view.
- [ ] Change the about page to be agnostic about Li Huiyu's gender. Joerlane believe Li may be female. No point is misgendering.
- [ ] Consider adding a toolbar with a large number of disabled tools. Paralleling what would be expected of similar software.
- [ ] Consider adding a widgeting/animation system. 
- [ ] Consider the backend modifications needed for Moshiboard. Send first, then monitor progress.

Issue #165, opened 2020-05-10
If initial classification doesn't create an engrave operation, it's hard to do that.


tatarize:
See, https://www.reddit.com/r/ChineseLaserCutters/comments/gfh2mv/meerk40t_cant_engrave_only_raster/

Basically a black object imported, and we want to engraved. There should be an easier, more intuitive method to do that. It should provide an easier set of methods to do that. 

https://www.youtube.com/watch?v=ikTlH5z2U-0&feature=youtu.be

Issue #175, opened 2020-06-02
Frozen Features from 0.6.0


tatarize:
Moving feature frozen requests from 0.6.0 PR #160 here.

- [ ] 5.2 (holdover) Jobs elements to be able to rearrange and delete, items.
- [x] Add Delete to Console and a bound button to the delete key. Might be a problem without undo.
- [x] Correct #165 needs an alternative method of classifying as engrave.
- [ ] Ruida Output Mode, between direct controlling the current System and creating elements (current procedure).
- [x] Save Binds and Aliases.
- [ ] delay console command. To send another command after a certain amount of time in the future.
- [ ] Cannot reorder in the spooler or jobs dialog.
- [ ] DXF layouts don't always fall in the bed. Many of these center around the center point which is often the bottom lefthand corner and are not automatically moved into the bed.
- [x] Tab to next element in preferences.

Bugs Unfixed from 0.6.0
- [ ] Unchecking Mock While Running, causes a crash.
- [x] Can close in the middle of a redraw cycle, crashing because object is now deleted.
- [ ] Changing active device in console causes hotkeys to fail so if modifying kernel controls start to fail.
- [ ] Selecting several elements at once and dragging them into a operation only moves the last object with focus.
- [x] Steps in the image settings, and step 4 on the menu, and steps in the raster are unlinked this is notably kinda confusing.

tatarize:
Closing in favor of #264

Issue #174, opened 2020-06-01
UI/UX Corrections and improvements.


tatarize:
Being worked on in the PR. Moved major bullet points there see #185

Modifications to be made.
The scene should have tools added to it, specifically lasso tool and object selection tool. There should be buttons to adjust the ordering of these elements. Move to front, move forward, push back.
The scene lasso tool should let you select either direction with the slightly different functionality.
Control + Click should function to toggle selected elements.
Selection of elements should not simply be the overall bounding box of the object but rather more directly clicking on the stroke or fill of that object. E.G. Regardless that the rectangle is on top, clicking here should select the circle since it's basically right on the stroke.
selection
The use of the elements tree should be less prominent. With a lot of clear methods of doing all the operations you might need directly in the scene itself. This is both because it's more intuitive and because the element tree is significantly less intuitive.
The color-operation 1:1 item selection scheme should be added. Provide a bottom row of color boxes. Clicking one of these boxes should color the selected element with that color, and add it these to similarly colored operation. Adding that element to a different operation with the color-operations will remove it from any operations it currently occupies.
While these aren't perfect are used elsewhere and work with great results. They are easy to apply and utilize and the color-operation selections can be defined and saved at the level of MeerK40t allowing loaded files to apply whatever assigned scheme is desired by the user.

This would largely de-emphasize use of the tree. RDWorks and Lightburn tend to just display the list of colors and what operation that color does.

It is not yet decided what would become of the tree as it stands. The objections @dietervansteenwegen raised in #171. Are pretty quite good. No part of the tree element is actually that useful. Also, it's fundamentally slow and without better underlying code keeps needing to be rebuilt often. Also the 3 different primary objects in there (the operations, the elements and the files) could likely be broken up into 3 different tree and become pretty significantly more usable. If done correctly it might eventually be scrapped altogether.

PR #185, opened 2020-06-25
UX UI Improvements.

tatarize:
This version is to be hyper focused on UI improvements a lot of the low hanging fruit for improved UI that hasn't been implemented should be implemented here.

Moving towards version freeze due to critical issue.

Critical Issues:
- [x] Rotating an image causes it to miscalculate the start point. If the image is large and rotated -90 degrees this can easily fly off the edge. It's adding the width and the offset to put it out of the scope.

Truncate:

Current Work

- [x] Closing while it's running can cause an error in the lhystudios device where the self.device goes away while the thread is running. This is clearly a lifecycle issue where it shutdown without killing the thread correctly.
- [ ] ctrl + click does not toggle selected objects.
- [ ] Selection is often made of only the top most object. There is no method from the scene to select any other object.
- [x] Due to the new classification, selecting everything and typing raster or alt+r does not put them all into a raster.

Push to next version:
- [ ] Colorbar for Assigning Operations
- [ ] Rect selection should allow both left to right (select all elements fully within the bounds) and right-to-left (select any items crossed by the lasso)
- [ ]  The list of tree objects is too long.
- [ ] The list of objects is doubled in the file foldout with little benefit to the user.
- [ ] The images of the elements provided is not very useful either.
- [ ] The generation of these elements is also very slow.
- [ ] All elements are expanded by default. 
- [ ] After any change the entire list is rebuilt, all positions are lost, and all elements are expanded. *again*
- [ ] Selecting multiple objects in the tree is not intuitive or very helpful. Because the labels and icons are not helpful.
- [ ] Dragging items into operations in the tree menu is not intuitive. Defects with the size of the list, causes this methodology to be problematic.
- [ ] Warn about items located outside the bed area.

Features Frozen from 0.6.0
- [ ] 5.2 (holdover) Jobs elements to be able to rearrange and delete, items.
- [ ] Ruida Output Mode, between direct controlling the current System and creating elements (current procedure).
- [ ] Save Binds and Aliases.
- [ ] delay console command. To send another command after a certain amount of time in the future.
- [ ] Cannot reorder in the spooler or jobs dialog.
- [ ] DXF layouts don't always fall in the bed. Many of these center around the center point which is often the bottom lefthand corner and are not automatically moved into the bed.
- [x] Tab to next element in preferences.

Bugs Unfixed from 0.6.0
- [ ] Unchecking Mock While Running, causes a crash.
- [ ] Can close in the middle of a redraw cycle, crashing because object is now deleted.
- [ ] Changing active device in console causes hotkeys to fail so if modifying kernel controls start to fail.
- [ ] Selecting several elements at once and dragging them into a operation only moves the last object with focus.

Done:
- [x] Add Delete to Console and a bound button to the delete key. Might be a problem without undo.
- [x] There is no rect selection ability.
- [x] Rectangle selection
- [x] Float on Windows rather than Always on Top.
- [x] Display Panel must be finished to show the raster information.
- [x] Crosshatch should be removed/implemented.
- [x] Layers at the side should be dropped / hidden / implemented.
- [x] Start position control should be disabled.
- [x] Group Pulses should be implemented for advanced.
- [x] Passes should be disabled.
- [x] Dot Length should be disabled/implemented.
- [x] Operations Menubar should be commented out.
- [x] Operation tree item should get add operation.
- [x] The rectangle selection in a field of a great many objects goes unacceptably slow since it's not caching the bounding box.
- [x] Rotate menu items do not rotate things correctly.
- [x] Lifecycle changes for various windows.
- [x] New Merged Operations Dialog
- [x] Multi-operations loading and saving and altering from main operations window.
- [x] Save/Load colored operation assignments
- [x] SVG Relative Upgrade
- [x] Ensure that cutting an image results in cutting the outer edge of that image.
- [x] It's not clear how to add operations #165 
- [x] Correct #165 needs an alternative method of classifying as engrave.
- [x] Right click on operations, does not allow you to add an operation. 
- [x] It is difficult to rearrange items. 
- [x] Image rastering routine should be made correct. The main difference is step-size is passed through to the image. So the Imaging raster does not step. The steps used are those found on the images themselves.
- [x] TextEntry in bad text object can be None and causes a crash.
- [x] Ability to include language translations in pyinstaller bundles.
- [x] The names of the tree elements are a tiny fraction of their SVG pathd structure. Which is of no use whatsoever.
- [x] The display name for the operations does not display the entire operation.
- [x] Mouseover of the text item doesn't provide the ellipsed text data.  Omitted for length.


tatarize:
#174
#175

joerlane:
Bug list:
Mac OS X has empty color boxes in the Path Properties
Mac has no word wrap in the About dialog window.
Mac has no tab support in operation settings dialogs.
Pi and Mac dialog windows do not open Front. Navigation, Terminal, and Operation Settings in both cases (all others should as well).

joerlane:
No tree icons show up for pure white text.

tatarize:
What? That's wrong. There's clearly a white text object on that white background. Clear as day.

tatarize:
Still needs some design decisions on what the color stroke operations should be. It's not clear if loading them all into bins is right, or if we should automatically create a new operation for any stroke color, or if we should automatically classify things based on whether they fit the criteria.

The black grabs all color stroke thing is a problem. It's not clear what that should actually do. It's kinda needed for classifying operations for rastering, it's not clear it should be used or is useful for other things.

We could also have color buttons that set the operation bin, and just classify that item with that operation. We could also make that mutually exclusive or let you have the same element in several operations.

Automatically creating an operation for an unseen stroke color is kinda nice, but it creates a lot of operations and those can save between different sessions.

The operations delete rather easily and they might need to be more permanent.

It's still an open question. Especially with saving for operations and elements. If these pre-make the operations should they therefore properly classify, or should the files already know which elements get classified where? And that would make the forcing auto-classification a potential problem.

I guess I can paddle water for a while but the design questions need an answer. There's not enough bugs to stall things for too long.

PR #190, opened 2020-07-02
Auto-create operations for any encountered colors & enable context-menu conversion

jaredly:
Currently there's no way to "Add" an operation, and if you have paths that don't match up color-wise with operations, they are just binned as raster. This makes it so you can define different paths as different colors, and each color gets its own operation.

I also implemented the "Convert Operation" context menu, for ease of conversion.

Test plan:
open an SVG with an orange line, a blue line, and a red line.
the blue and red lines should get binned into the preexisting operations, and a new orange operation should be created.

right-click the orange auto-created "engrave" operation, use "Convert Operation" to change it to a `Cut`. Double-click on it to see that in the edit window it also appears as a "Cut".

image

### DIFF
--- a/Kernel.py
+++ b/Kernel.py
@@ -1260,9 +1260,11 @@ class Elemental(Module):
         if elements is None:
             return
         for element in elements:
+            found_color = False
             for op in self.ops():
                 if op.color == element.stroke:
                     op.append(element)
+                    found_color = True
                 elif op.color == 'black':
                     if op.operation == 'Raster' and not isinstance(element, SVGImage):
                         op.append(element)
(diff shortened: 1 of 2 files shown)


tatarize:
Yeah, there's still some active debate between @joerlane  and myself as to how this should eventually work. We kinda want it classifying stuff a bit like RDWorks but it's a little bit up in the air as to whether everything should get a color or whether the blacks for a classify everything is reasonable or if it needs other things. And whether things should automatically classify into the given operations always and right away.

jaredly:
Thanks!

PR #198, opened 2020-07-16
Notes & Improvements

tatarize:
Current Work:
- [x] Save vs. Save_as Issue.
- [x] WASD update should clear movement on jobfinish to prevent jog increasing and firing on completion.
- [x] tests WASD keys personally.
- [ ] Main window losing focus should force an keyup or end loop command.
- [x] Potential multithreading error. AttributeError for Path attribute cache during the laserrender at 157. Only possible if shutting down and cache was just deleted by the other thread. This would also potentially allow creation prior to shutdown leading to a post-close crash.

Unfinished 0.6.2:
- [ ] ctrl + click does not toggle selected objects.
- [ ] Selection is often made of only the top most object. There is no method from the scene to select any other object.

Push 0.6.2:
- [ ] Colorbar for Assigning Operations
- [ ] Rect selection should allow both left to right (select all elements fully within the bounds) and right-to-left (select any items crossed by the lasso)
- [ ]  The list of tree objects is too long.
- [ ] The list of objects is doubled in the file foldout with little benefit to the user.
- [ ] The images of the elements provided is not very useful either.
- [ ] The generation of these elements is also very slow.
- [ ] All elements are expanded by default. 
- [ ] After any change the entire list is rebuilt, all positions are lost, and all elements are expanded. *again*
- [ ] Selecting multiple objects in the tree is not intuitive or very helpful. Because the labels and icons are not helpful.
- [ ] Dragging items into operations in the tree menu is not intuitive. Defects with the size of the list, causes this methodology to be problematic.
- [ ] Warn about items located outside the bed area.

Features Frozen from 0.6.0
- [ ] 5.2 (holdover) Jobs elements to be able to rearrange and delete, items.
- [ ] Ruida Output Mode, between direct controlling the current System and creating elements (current procedure).
- [ ] delay console command. To send another command after a certain amount of time in the future.
- [ ] Cannot reorder in the spooler or jobs dialog.
- [ ] DXF layouts don't always fall in the bed. Many of these center around the center point which is often the bottom lefthand corner and are not automatically moved into the bed.

Bugs Unfixed from 0.6.0
- [ ] Unchecking Mock While Running, causes a crash.
- [x] Can close in the middle of a redraw cycle, crashing because object is now deleted.
- [ ] Changing active device in console causes hotkeys to fail so if modifying kernel controls start to fail.
- [ ] Selecting several elements at once and dragging them into a operation only moves the last object with focus.


tatarize:
-[ ] Save vs. Save_as Issue.

tatarize:
Outstanding points moved to #264

PR #214, opened 2020-08-28
Earlier 0.7.0 Code.

tatarize:
This is now an intermediate step for the 0.7.0 upgrade.


tatarize:
array('B', [255, 238, 111, 148, 19, 255])
array('B', [255, 206, 111, 148, 19, 255])
array('B', [255, 206, 111, 148, 19, 255])
b'\x00aBcMaBbMaBcMaBbMaBbMaBcMaBbMaB\xa3'

This should request acks 1 time rather than twice.

tatarize:
TextProperties fails to open.

tatarize:
PlotPlanner accepts cutcode objects these are processed into usable bunches of cut positions. We return None if the cutcode requires a jog to a new location? We need to flush data from the plot planner on occasion, how does that data get processed?

Cutcode is processed directly by the interpreter as particular commands. Lasercode generated helper commands. These aren't really needed as much, but these modify the local settings.

* Plot planner should solve for the initial directionality of the a cut. Unless in Strict Mode.
* Plot Planner needs to flush data when the cut settings are changed.
* Restore rapids between objects that are not adjacent or close.
* Rewrite fly switch to lazy kick in but only on a change event. Should not do multiple switches.
* EGVPlotter needs to plot cutcode.

PR #230, opened 2020-09-23
Initial -H home command.

tatarize:
- [x] CLI -H command added for home.
- [x] CLI -s changed key=value
- [x] CLI -O command origin return after.
- [x] CLI -p is fixed. -p takes in an SVG path. so -p M1000,1000v1000h1000v-1000h-1000 should work to draw a 1in 1in rectangle at 1in 1in. The paths are native so 1000 = 1in.
- [x] Correct Actualization Crash. 0.6.5 introduced.
- [x] Add svg_ppi to kernel for loading svgs.
- [x] Correct other actualization routines with the 0.6.5 bounds correction.
- [x] Correct meerk40t/svgelements#50 circle decomposition if h or v mirrored.
- [x] Paths renamed to be Path_1 Path_2, etc. The path names are mostly pointless, even I can't read them. 
- [x] Corrected a right click actualization crash.
- [x] Stroke Width scale correction. Reifies as the abs(sqrt(det)) of the matrix.
- [x] Sped up image threshold.
- [x] Allow crop to accept length values.
- [x] Changed some raster classification stuff.
- [x] Added `image path` command.

### DIFF
--- a/Console.py
+++ b/Console.py
@@ -1473,6 +1473,11 @@ class Console(Module, Pipe):
                                 new_data[x, y] = (v, v, v, 255)
                     elements.add_elem(image_element)
             elif args[0] == 'resample':
+                try:
+                    from OperationPreprocessor import OperationPreprocessor
+                except ImportError:
+                    yield "No Render Engine Installed."
+                    return
                 for element in elements.elems(emphasized=True):
                     if isinstance(element, SVGImage):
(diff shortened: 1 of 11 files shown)


joerlane (approved):
Compiles binaries in a number of operating systems. Useful CLI operation. Tested

PR #236, opened 2020-09-30
0.6.8

tatarize:
- Adds Rapid Option to rapid to next location. In Jobs, on by default.
- Adds Jog Standard, Switch, and Finish option to select type. Typically Standard.
- Optimizes the OK queries. Speeds Raster up by 42% before stuttering occurs. 100 mm/s minimum before stutter vs 70 mm/s previously.
- Adds code to do Standard Jogging with N-SE lhymicro-gl sequences.
- Adds in rotaryview and rotaryscale options for use with rotary see issue #249 
-  Added in `control Jog Transition Test`
- Added in option to Force Continue in Controller when the finish failed to advanced.
- Fixed bug that would sometimes prevent the finish from advancing.
- Corrected a console syntax bug 'egv #$'.


joerlane (approved):
Tested the new Jog modes extensively against both intermittent runaway and hang failures. Multiple runs with Standard and Finish ran without issue. I approve.

PR #250, opened 2020-10-24
Minor Dev. 0.6.9

tatarize:
Minor Dev. 0.6.9 fixed a tiny bug


tatarize:
meerk40t69

joerlane (approved):
Tested the rapid Jog features with SVG files containing exploded paths (paths that appear as one object, but are actually made up of individual line segments, and not complex path). The Jog Standard appears to be functioning well. Jog Finish still does some extra moves either before and/or after; but not 100% of the time, and the exploded path issues appear to be resolved. Developed Mac tests to attempt to reproduce the -psn bug (appears to be resolved as well).

PR #251, opened 2020-10-26
Minor Dev 0.6.10

tatarize:
- [x] Corrects a bug of engraving SVGText. This isn't permitted but failed to strip the data.


joerlane (approved):
Checked the element * merge, and optimize travel in regards to an errant cut line. Tested the stripping of text from engrave during Job start. Text is now movable with the mouse after being generated by the Terminal.

PR #254, opened 2020-11-02
0.6.11

tatarize:
- [x] Prevent the impossible addition of Images and Paths.
- [x] Do not add empty paths. #256 
- [x] Correct but with DXF arcs going CW in error. #257


joerlane (approved):
Checked DXF corner issues. Zero width rectangles. Element merge with images. Everything checks out as expected.

PR #260, opened 2020-11-15
0.6.12

tatarize:
- [x] Corrected a minor extra-move error.
- [x] Correct DXF Splines.


joerlane (approved):
Tested changes with grid and stroke size vs. zoom. Added Mac highDPI display support through pyinstaller.

PR #261, opened 2020-11-22
0.6.13

tatarize:
Added CLI -V
Couple versions increase of `svgelements`
Fixed a serious bug where vector would fail to classify if a raster existed in the operations.


joerlane (approved):
Checked out the -V and color classification issues against the broken test file.

PR #318, opened 2021-01-25
use GetPoint and HitTest to determine the node which is dragged

frogmaster:
This fixes dragging within operations multiplying on GTK.


tatarize:
Does this problem exist in the stable branch of 0.6.x too? And would the same fix also correct it there? Or does this maybe also require some my earlier anti-duplicating code that didn't succeed?

PR #356, opened 2021-02-11
0.6.20

tatarize:
Initial Version Bump.

- [x] Settings window needs to be opened up a hair to display text in Mac OS. A minimum amount would be ideal so as not to disturb other OS views. Add 20px in width, and 10px in height to default opening size.
- [x] Lost DXF stroke indications on file load with massive scaling down.
- [x] DXF load with massive scaling down really shouldn't have happened in the first place.
- [ ] Should inform on scale (and display scale); when loading a file that is automatically re-scaled. (Requires too much change)
- [ ] Inkscape generated DXF export, still does not load same scale as drawn. (Can't be fixed, Inkscape doesn't write the scaling there, it only scales the objects).
- [ ] Crash on disconnect controller while running job (sometimes).
- [ ] Might be possible to fix Cursor display for Mac OS based on some Windows glitch testing.


joerlane (approved):
Tested the dxf rescale, center. Grid, plus Step and Repeat. Mac settings changes for pinch zoom, "scroll" pan; and inversions.

PR #384, opened 2021-03-10
0.6.22

tatarize:
- [x] Black background in some icons.
- [x] Offset circle issue


joerlane (approved):
Tested minus the USB long cable issues.

Issue #401, opened 2021-04-18
Option to Tell User DXF was resized


tatarize:
You could turn it off later but if the DXF wasn't just moved but actively resized the user should be informed with an optional popup.

tatarize:
This is corrected in 0.6.23, still unpublished.

https://github.com/meerk40t/meerk40t/pull/403/commits/9e231d9abaf6f400541ba25961f9693228bfaed6

tatarize:
Published.

https://github.com/meerk40t/meerk40t/releases/tag/0.6.23

Issue #393, opened 2021-03-21
Warn about closing app during burn


Sophist-UK:
I am truly an idiot - once is a mistake, twice is stupidity. Twice now I went to close the spooler window and instead closed the main window.

Can we please have an "Are you sure?" confirmation if we are closing Meerk40t whilst a burn is in progress?

tatarize:
There used to be such a setting, but it got lost in one of the updates/conversions. It also has to go kinda out of the way to kill a process in progress. Otherwise the GUI closes down but the thread stays there sending data without fully killing the processes. I'll be sure to add it back.

tatarize:
This is in the 0.6.23 branch.

tatarize:
Published.

https://github.com/meerk40t/meerk40t/releases/tag/0.6.23

PR #403, opened 2021-04-24
0.6.23 Version

tatarize:
- [x] #393 warn about closing during a burn. Restore popup asking that you cancel.
- [x]  Report that raster step can be made to crash. I can't get set raster steps on an image to crash "if you select if from the Image Operation right-click menu."
- [x]  Adding in the USB-retry dialog thing or whatever we settle on. STOP button, Pause Button, and something to say the USB is constantly retrying until you actively tell it to stop trying. Then it'll tell you that controller might still be able to restore the process if we're not hard-aborting.
- [x]  File "RasterScripts.pyc", line 473, in wizard_image, when trying to configure rasterwizard, other modes disabled. "PIL/Image.pyc", line 3011, in blend
- [x]  Add #401 moving the dxf is one thing, but resizing it needs to inform the user.
- [x]   File "wxMeerK40t.py", line 1678, in open_path_dialog /  File "Kernel.py", line 2250, in classify /   File "Kernel.py", line 1306, in classify / AttributeError: 'Move' object has no attribute 'stroke' -- F8 inclusion of a path caused a crash.
- [x] Correct DriverManager Crash. No selection, click Remove.
- [x] Removed Property Button within Devices.
- [x] Issue #416 remaining raster issue hitting left edge for transparency holding 'P' type image.


tatarize:
@meerk40t/meerkittens  Okay. This is ready to review and merge.

The stable branch has no major changes. So generally it should be unfailing but special checks are needed with regard to:

Usb disconnected now gives 5 attempts then loads up the controller and keeps trying until you click the red "retrying" button to stop. After that you'd need to stop and restart the controller. This should mean if you unplug the USB for a long time it'll pop up the controller but will not kill the project in a way to annoy folks.

Closing while it's trying to run something should refuse without an additional check.

Should tell the user when a DXF file was resized.

Crash if you hit remove in DriverManager with nothing selected.

 @Sophist-UK white / transparent background still burned as black  #416 should be correctly fixed.

I've checked these things and the hope is that nothing in the changes causes any additional issue. 0.6.x should remain rock solid even if 0.7.x does a lot of weird things.

tatarize:
Circa the beta release of 0.6.23 in May.

0.6.23 Beta 1 -- Few bug corrections
@tatarize tatarize released this on May 5

RasterWizard, several modes disabled bug.
Add #401 moving the dxf is one thing, but resizing it needs to inform the user.
F8 Add path caused a crash.
Correct DriverManager Crash. No selection, click Remove.
Removed Property Button within Devices.
Removed Step from ImageOp -- Cannot be set.
Right Click menus include the meaning of double-click as first entry.
svgelements update.
Added DPI menu. (Step size without resize)
Corrected "Solitaire Effect" for Linux in RasterWizard.

Sophist-UK:
> I've checked these things and the hope is that nothing in the changes causes any additional issue. 0.6.x should remain rock solid even if 0.7.x does a lot of weird things.

I think this comment on 0.7.0 is a bit unfair. There are a quite a few bugs, a few of which cause crashes rather than e.g. just slowing things down, but 0.7.0 does work pretty well IME.

tatarize:
I should say 0.7.x is allowed to do weird things and add in features etc. 0.6.x isn't. No new features, only bug and workflow fixes. It should remain solid. It's not allowed to do weird things. Editing the code is done *very* conservatively. Also, getting to work takes precedence over getting it coded in a modular improvable way. Which means it's best to make sprawling spaghetti code patches rather rewrite stuff that is kind of poorly done from an API standpoint (that stuff got fixed in 0.7.x etc)

PR #411, opened 2021-05-10
0.7.0 Device-Update

tatarize:
Change devices to a conglomeration of different smaller parts.

* Spoolers are their own realm, and are just queues to put data in.
* Interpreters are their own realm, and read from a spooler and output to a pipe.
    * Interpreters come in a couple different flavors and this is the main point of a device type.
* Pipes are their own realm, and move around like channels but in registered places.
    * Pipes are sometimes simple like writing to disk or to a network connection, but sometimes more complex like a lhymicro-controller set of code.

These devices get hooked together with some basic commands, the links are part of devices. And can be queried, and booted by being registered as a device. 

`interpreter new Lhystudios pipe file results.egv` would register a new Lhystudios interpreter which would pipe to a file called results.egv. The spooler in this case is not defined. And the interpreter would not be producing data. But, something like:

`pipe tcp-listen -p 440 pipe LhyController` would listen at port 440. And when it gets a connection or signal would send that to the lhycontroller pipe. Which would make the localhost a lhymicro-server.

`spool interpret Lhystudios pipe Lhystudios` would hook spooler default to interpreter lhystudios and that to pipe lhystudios. Which makes spool in this case a basic lhystudios device.

At least this is the plan. Then the spooler window gets made more specific by selecting from a list of active spoolers. And the planner needs to specify the spooler getting the data with the spool command which accepts a particular destination.


tatarize:
lhystudios-dialogs

PR #413, opened 2021-05-11
Make xml link to meerk40t svg namespace immutable.

Sophist-UK:
Wiki pages URLs are subject to change by anyone (like me 😄 this evening) and should not be linked to in this way.

We need an immutable URL which contains sufficient HTML to give a brief description and a link to the actual Wiki page.


tatarize (approved):
This seems pretty solid and correct. Namespace doesn't need to change that much but does need to be somewhat locked in.

tatarize:
Patched this into 0.7.0-devices branch as well.

PR #448, opened 2021-06-24
0.7.0 optimize3

tatarize:


tatarize:
Merge Passes:
Combine all passes together.

Merge Ops:
Combine all regular ops.

Cut Inner First:
Cut Objects are broken into inner first values.

---

* `Merge Passes` must be enabled in order to merge ops. `Merge Ops` implies merge passes.
    * If merge passes is 
`Cut Inner First` prevents the merging of cut operations with the exception of the outermost cuts.
`Cut Inner First` independently runs along side of merge passes. If passes are merged the isolated cut operations

Sophist-UK:
@tatarize 

> Merge Passes:
> Combine all passes together.
> 
> Merge Ops:
> Combine all regular ops.
> 
> Cut Inner First:
> Cut Objects are broken into inner first values.
> 
> * `Merge Passes` must be enabled in order to merge ops. `Merge Ops` implies merge passes.

As previously stated, if Merge Passes e.g. does the same cut repeatedly before moving on to the next  cut, then this needs to have a major fire safety warning attached to it!! I am also unclear why you cannot do merge operations without merge passes (which on reflection is pretty much what I have outlined below except that if you have duplicated operations as an alternative way of doing multiple passes, then you may well cut the same path twice in quick succession).

I am guessing that you haven't followed the approach I suggested whereby you merge operations and passes iteratively with an objective of no path being repeated in less than the time taken to do a single pass i.e. You merge all raster operations and at most one of each type of vector operation, but also at most one pass from each of these operations, and you execute that and then repeat until there are no operations left to perform. And you always start each iteration from a virtual home (i.e. you optimise as if the head is starting at Home without actually doing a Home) in order to try to ensure that a second iteration doesn't work in reverse thus cutting the last path of the first iteration as the first path in the second iteration.

This approach almost guarantees that you won't repeat the same path in less time that it would take without any optimisation. And on a pareto basis it will still deliver the vast majority of time savings.

>   * If merge passes is
>     `Cut Inner First` prevents the merging of cut operations with the exception of the outermost cuts.
>     `Cut Inner First` independently runs along side of merge passes. If passes are merged the isolated cut operations

Sorry - but I cannot make head or tail of this. Can you explain further?

tatarize:
The little part I'm stuck at is how these work together. Merge Passes merges passes. But, how does that interact with the state of Merge Ops and Cut Inner First. I made some notes and posted it to my myself. There's issues when I combine passes but not ops or ops and not passes and they go weird. I still need the lowest level cut to divide out and to figure out the bulk of this stuff for what it should do.

* `Merge Passes` and `Merge Ops` and `Cut Inner First`
     * Multiple passes are combined into a single op. Operations are combined into a single cutcode object, except the outer bits of the cut which are put after the main bit of cutcode.
* `Merge Passes` and `Merge Ops` and NOT `Cut Inner First`
     * Passes and Ops are all combined into the same Cutcode. The entire thing is optimized together.
* `Merge Passes` and NOT `Merge Ops` and `Cut Inner First`
     * Passes within ops are merged together. Cut Objects are divided by level. Passes within the cut objects are also divided by level.
* NOT `Merge Passes` and `Merge Ops` and `Cut Inner First`
     * Unknown. This might mean something but it also could just be disallowed.
* NOT `Merge Passes` and `Merge Ops` and NOT `Cut Inner First`
     * Unknown. Passes look like different ops anyway.
* `Merge Passes` and NOT `Merge Ops` and NOT `Cut Inner First`
     * Each object is treated as an independent grouping. All the passes within that group are combined together.
* NOT `Merge Passes` and NOT `Merge Ops` and `Cut Inner First`
     * All passes are different groups. All ops are different groups. Cut objects are divided by levels.
* NOT `Merge Passes` and NOT `Merge Ops` and NOT `Cut Inner First`
     * All passes are different groups. All ops are different groups. Cut objects remain their own group.

The only really weird set are do not merge passes and do merge ops. The outcome of this set of options is not obvious to me. I suppose it could be done pass first in this case. Do all the first passes as a group. Then do all the second passes, as a group. So you could set 4 ops with 3 passes and then if you merge the passes but do not merge the groups, you would get 3 passes with the 4 ops merged together. With Cut Inner First triggering whether the cut objects are divided by levels. However in this case the cut objects are divided apart by passes.

When not merging passes but merging the ops,  the passes are done as the outer level. This might be weirdly more complex that it needs to be. This could be set with a different option to state the priority of the passes vs. the priority of the ops.

tatarize:
Went a slightly different way, and restored most of the previous functionality and implemented the merge values as normal.

I did some mapping out of the requirements and there were going to be some larger problems, dividing out the cut objects from the rest rather than doing them within constrained cutcode.

Issue #459, opened 2021-07-22
(Documentation) After running from source on windows, meerk40t cannot connect to usb


gsboylan:
![image](https://user-images.githubusercontent.com/9797076/126579708-edb18b14-c1bd-4fa7-8e4b-6685b5e73c11.png)

Followed instructions here: https://github.com/meerk40t/meerk40t/wiki/Install:-Running-from-source

pyusb and the libusb k40 driver from ScorchWorks are both installed on my machine.

tatarize:
Hm. It really should have aborted that and tried the whisperer driver if the libusb driver failed. Maybe for long term I should give you a force switch for the config thing to let you manually choose your driver. So in the future if this comes up more, it'll be fixable without too much trouble.

gsboylan:
This works fine for me with the precompiled executable - so I assumed there's some setup missing, somehow. Not 100% sure. For what it's worth, I've never installed the stock driver that came with the k40, only the scorchworks driver.

tatarize:
Yeah, when you plug it in the windows driver should load up. This is the CH341DLL driver which works fine, however, it's not loading. In that case it's *supposed* to try the LibUsb Driver which works with Whisperer. but somehow it's not doing that. It's supposed to switch between drivers depending on what works. The goal is to have it always work, but it looks like the windows driver failed and it failed in a way that prevented it from trying the other driver or simply didn't correctly attempt the fallback.

This needs to be checked before full release, I'll try deleting the dll and doing other things with it to see if I can replicate it.

gsboylan:
Let me know if you need me to test out a patch

Sophist-UK:
> Followed instructions here: https://github.com/meerk40t/meerk40t/wiki/Install:-Running-from-source

I rewrote these when I installed to run from source on my own Windows laptop. I am not sure anyone other than you has tried them since, so there may well be a problem.

The Windows shortcut that is created does **not** create a console for console errors to appear on. You might want to create a small `.cmd` file as follows which you can run:
```
@Echo On
python meerk40t.py
pause
```

@tatarize The error message `Could not find module 'CH341DLL.dll' (or one of its dependencies). Try using the full path with constructor syntax.` is really not very helpful to the user - though it may be helpful to you because it seems to be finding the dll when running from an executable but not when running from source. You might want to ask @gsboylan to check that this dll is in the expected place when running from source.

tatarize:
The error message isn't mine, it's likely python with the call to ctypes.

[...]

PR #460, opened 2021-07-23
#459 Fix incorrect error types loading drivers

gsboylan:
ch341.py expects drivers to raise ImportError when a failure occurs during initialization, and ConnectionRefusedError when the failure occurs during connection.

libusb.py and windll.py both are throwing ConnectionRefusedError during the initialization phase. A failure to load the driver during this phase will not be caught by the calling code and will result in a failure to fall back to any alternate driver present.

### DIFF
--- a/meerk40t/device/ch341/libusb.py
+++ b/meerk40t/device/ch341/libusb.py
@@ -114,7 +114,7 @@ class Handler(CH341Handler):
                     )
                 )
                 connection.close()
-                raise ConnectionRefusedError("Chip Version Wrong.")
+                raise ImportError("Chip Version Wrong.")
         if bus != -1:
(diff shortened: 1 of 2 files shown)


tatarize:
This is a bit more complex. It needs to turn the file not found into an import error. The connection refused stuff is if the driver loaded but did not connect. That's how that should react correctly there. While I don't have the issue myself, I checked the flow and the better solution is to either catch the filenotfounderror in the main routine or to reraise it as an import error (since that's kind of what that is). I made a commit that should fix it

tatarize:
#459 see https://github.com/meerk40t/meerk40t/commit/962ab67230e674754d5d407340707fd6cd73e301

PR #466, opened 2021-08-06
Various operation classification enhancements

Sophist-UK:
1. Use distance_sq to avoid unnecessary sqrt
2. Use a larger distance for red = cut classification
3. Classify black = raster for distance up to 48. Keep Engrave for all other colours
4. Only add fill-only element to Raster Operation of same fill colour not to all Raster Operations
5. Add operation for fill-only element if no existing matching raster op

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -4492,7 +4492,7 @@ class Elemental(Modifier):
                     elif isinstance(element, SVGText):
                         op.add(element)
                         was_classified = True
-                    elif element.fill is not None and element.fill.argb is not None:
+                    elif element.fill is not None and op.color == abs(element.fill):
                         op.add(element, type="opnode")
                         was_classified = True
                 elif (
@@ -4518,12 +4518,18 @@ class Elemental(Modifier):
                     break  # May only classify in Dots.
             if not was_classified:
                 if element.stroke is not None and element.stroke.value is not None:
-                    if Color.distance("red", element.stroke) < 90: # if element.stroke == Color("red"):
+                    if Color.distance_sq("red", element.stroke) <= 18825: # if element.stroke == Color("red"):
                         op = LaserOperation(operation="Cut", color=element.stroke)
+                    elif Color.distance_sq("black", element.stroke) <= 2304: # if element.stroke == Color("black"):
+                        op = LaserOperation(operation="Raster", color=element.stroke)
                     else:
                         op = LaserOperation(operation="Engrave", color=element.stroke)
                     add_op_function(op)
                     op.add(element, type="opnode")
+                    operations.append(op)
+                elif element.stroke is None and element.fill is not None and element.fill.value is not None: # add operation for fill-only elements if not existing
+                    op = LaserOperation(operation="Raster", color=element.fill)
+                    op.add(element, type"opnode")
                     operations.append(op)
 
     def load(self, pathname, **kwargs):


tatarize (approved):
Seems fine. I'm not sure sq_root commands take enough time to matter there, but it's also fine.

Sophist-UK:
There are many places that tiny optimisations could be done - I spotted one where there were two consecutive `if`s where an `elif` could have been used saving a comparison.

The real performance improvements will come from optimising using numpy (or similar) in threads and / or on GPU if available.

tatarize:
Usually I do elif where it matters sometimes they might look similar but if I modified the objects the conditionals are being checked against it matters a lot. Sometimes I do strings of "if object is not None" followed by another if statement of "if object is not None" which seems a bit weird but the code flow sort of required it. Also, I don't hyper-optimize things that don't happen much or for long. Things like rasterwizard relies as heavily on Pillow as I can because it's faster. And the optimize could certainly be optimized rather heavily, especially with some acceleration structures. 

Numpy is not a requirement of meerk40t. It is of wxPython so if you have a gui, you need numpy and of Pillow. The issue arises with the workflow of an install on a headless pi-zero which sits there sending stuff to the laser. It doesn't need that stuff and wouldn't do much of anything. But, there's nothing in an install like that that needs numpy. There's little point in hyper-optimizing code that doesn't take long to run or isn't run that often. 

There are some other optimizations that might be very useful especially in the widget system. It redraws the entire screen too much and serious does not need to do that. It could very much save an bitmap of the screen and update a small section of that bitmap rather than redraw everything. There's a lot of cases where different layers could do wonders. Like redrawing the interface and redrawing the shapes. You could get by without redrawing the interface if you stored that in a layer and kept a bitmap of that then you could compost a few bitmaps together. Rather than fully rerender entire elements. Often the elements themselves store some cache stuff which helps a lot but it might be far superior to do some level-of-detail rendering and redraw some small sections when only one thing changes.

Sophist-UK:
My point was really exactly that - that there are often huge numbers of tiny tweaks like this, but generally they are not worthwhile.

Personally, I am really not sure about the GUI optimisations that you are talking about - the wxPython should maintain its own layers and only redraw what it needs to. (That is what GPUs are for - or at least what they were originally for, to maintain several layers and merge them for display - then they started to hold a 3D model and convert to 2D for display, and then to be used for fast, parallel calculations.) 

My own experience of GUI speed (but admittedly on fast-ish processors - aging I7s) is that 0.6 was slow, and 0.7 is easily fast enough.

I haven't looked at the optimise code, but I expect that it has array calculations (or could be tweaked to use array calculations) that would be much much faster in numpy, and using a numpy-alike could be even faster by using the GPU. (Not sure whether you can substitute a numpy-alike for the wxPython requirement - but if you can, then that could use the GPU as well.)

meerk40t/core/elements.py
```diff
@@ -4518,13 +4518,19 @@ def classify(self, elements, operations=None, add_op_function=None):
...
                         op = LaserOperation(operation="Cut", color=element.stroke)
+                    elif Color.distance_sq("black", element.stroke) <= 2304: # if element.stroke == Color("black"):
+                        op = LaserOperation(operation="Raster", color=element.stroke)
                     else:
                         op = LaserOperation(operation="Engrave", color=element.stroke)
                     add_op_function(op)
                     op.add(element, type="opnode")
                     operations.append(op)
+                elif element.stroke is None and element.fill is not None and element.fill.value is not None: # add operation for fill-only elements if not existing
+                    op = LaserOperation(operation="Raster", color=element.fill)
```
Sophist-UK, line 4531: @tatarize I realise that this code is somewhat simplistic.

It will deal with the reported bug whereby Inkscape Text->Path produces paths with no Stroke but a fill - and these do indeed need to be Raster operations.

But there is a common false-cut - which looks like a cut but is actually a very thin fill surrounded by a closed path without a stroke. These will be misclassified by this code. The options to fix this are either:

1. Use the colour to determine the Operation type exactly as in the code above; or
2. Analyse the path itself to determine whether the fill looks like a stroke (which seems way too difficult to me).

So I guess we should go for option 1. It is late and I am too tired to get my head around how to code this. Not sure if you want to take a shot at it, or whether you would like me to when I have the time to do it when I am fresh.

PR #469, opened 2021-08-07
Prefer forward paths to backward paths in greedy

Sophist-UK:
Amazingly, one character should do it (for perfect matches).

So for closed paths (which are the primary issue I see with this) where path segments are exactly the same points, this should make these always forwards.


tatarize (approved):
Looks like I was giving ties to backwards.

Sophist-UK:
Further tweak to prefer shorter (esp. zero length) segments over longer ones so that zero length segments are not left over then requiring an unnecessary travel move is in #471.

PR #471, opened 2021-08-08
Make greedy prefer shorter paths all other equal

Sophist-UK:
This ensures point burns are done before move burns which would otherwise result in additional travels.


Sophist-UK:
P.S. I have tested this code.

tatarize (approved):
I'd make 1 a variable for the amount of slop. Also I don't know if forward should be preferred if we just went backwards.

PR #472, opened 2021-08-08
Improve svgelements __repr__

Sophist-UK:


Sophist-UK:
Closing because there is also a bug in svgelements that needs to be fixed as part of #471 - so I will incorporate these changes into that.

PR #473, opened 2021-08-08
Improve classification to use fill colours if no stroke

Sophist-UK:
Previous code to handle Shapes with no stroke would work for text->path objects, but not for e.g. things that looked like cut segments (red and thin) but were really thin red fill and no stroke which would be classified as raster rather than cut.

This code simply uses fill colour as backup if there is no stroke and then classifies accordingly. It ends up being simpler as a consequence.

Dot support would not have previously worked because it was on an `elif` that would never be reached. Corrected Dot support has been commented out because there is no support elsewhere for Dot operations.

SVGelements had a bug on calculating colour distance (missing brackets) which I have fixed. During debugging I also fixed a bunch of __repr__ stuff in SVGelements and these are also included.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -4453,12 +4453,14 @@ class Elemental(Modifier):
         """
         Classify does the placement of elements within operations.
 
-        "Image" is the default for images.
+        "Image" is classified as Image.
+        Dots are a special type of Path
+        Text is classified as Raster
 
-        Typically,
-        If element strokes are red they get classed as cut operations
-        If they are otherwise they get classed as engrave.
-        However, this differs based on the ops in question.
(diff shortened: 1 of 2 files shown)


Sophist-UK:
Here is the file I used for testing which contains all main SVG elements except Text and Clones (which are not currently supported anyway).

[Element classification test.zip](https://github.com/meerk40t/meerk40t/files/6950995/Element.classification.test.zip)

tatarize (approved):
The SVGelements code is going to need to run through some checks or it'll get replaced. It's its own project and is used in other code elsewhere. So there's a writing and maintaining checks and pretty strong set of established tests. And a general understanding that new changes also result in new tests.

PR #475, opened 2021-08-08
Various SVGelements __repr__ tweaks

Sophist-UK:


tatarize (changes requested):
```python
        x_str = Length.str(round(self.x,2))
        y_str = Length.str(round(self.y,2))
```
This is a problem. First in repr it should have a lot of detail and not round. It should be as detailed as possible in a technical sense. Also, another reason Length.str() is used is because the self.x could be a `Length` object which means it cannot round. (Though I could actually add some basic math commands there). So 2in rounded is a crash. If there's a need for a rounding level there it should be added to the Length.str() function. Since the input can be either `float` or `Length` object

Sophist-UK:
Ok - I will revert these lines.

tatarize (approved):

PR #476, opened 2021-08-08
Prevent Tree pane from limiting width

Sophist-UK:
Fix situation where when you widen the pane bar holding the tree pane, the bar gets wider, but the tree only widens to a fixed point and then stops


tatarize (approved):
The tree limiting width might need to be checked against the rPi and Mac. I think there were times it went insane. Not sure it's still a thing.

Sophist-UK:
I tested this on Windows before I submitted the PR and it does what it says on the can.

PR #483, opened 2021-08-11
Various minor changes...

Sophist-UK:
1. Revert change to number of digits in Point coordinates.
2. Minor improvements to __repr__

### DIFF
--- a/meerk40t/svgelements.py
+++ b/meerk40t/svgelements.py
@@ -1169,7 +1169,7 @@ class Color(object):
     @staticmethod
     def parse(color_string):
         """Parse SVG color, will return a set value."""
-        if color_string is None or color_string == SVG_VALUE_NONE:
+        if color_string is None or color_string.lower() == SVG_VALUE_NONE.lower():
             return None
         match = REGEX_COLOR_HEX.match(color_string)
         if match:
@@ -3688,9 +3688,9 @@ class Shape(SVGElement, GraphicObject, Transformable):
         if not self.transform.is_identity():
             values.append("transform=%s" % repr(self.transform))
         if self.stroke is not None:
-            values.append("stroke='%s'" % self.stroke)
+            values.append("stroke=%s" % repr(self.stroke))
         if self.fill is not None:
-            values.append("fill='%s'" % self.fill)
+            values.append("fill=%s" % repr(self.fill))
         if self.stroke_width is not None and self.stroke_width != 1.0:
             values.append("stroke_width='%s'" % str(self.stroke_width))
         if self.apply is not None and not self.apply:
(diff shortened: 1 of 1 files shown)


tatarize (changes requested):
While the color string to lowercase is likely fine, and certainly repr within repr_shape is preferred. 

The case insensitivity during the SVG parsing may actually be out of spec. I believe inline style are CSS and therefore case sensitive and non-inline are SVG and case insensitive. I'm pretty sure the specific defined None object there is already required to be lowercase. `currentColor` is required to be like that with the given case the WC3 tests on these things are somewhat extreme.

Style attributes are case insensitive whereas presentation attributes *are* case sensitive. So the word `fill="red"` should set the fill to red. Whereas `fiLl="red"` should *not* set the fill to red. However: `style=fiLl="red"` *should* set the fill to red. I think the same applies here a bit.  

Non-case sensitivity style attribute. Presentation attributes are case sensitive. Style sheets are not. The style attribute is non-case sensitive.

I think `currentcolor` should not actually correctly work. I think there's even a WC3 test to make sure of that.

tatarize:
The correction should be made in the save routine. To do `none` correctly rather than None. The svg code is correct that `None` fails to be a color. Though I'll check the X11 color lists or whatever, but I'm pretty sure `none` is supposed to work and `None` is supposed to fail.

Sophist-UK:
> The correction should be made in the save routine. To do `none` correctly rather than None. The svg code is correct that `None` fails to be a color. Though I'll check the X11 color lists or whatever, but I'm pretty sure `none` is supposed to work and `None` is supposed to fail.

I am torn here. Part of me says svgelements should faithfully respect the SVG specification including rejecting `None` when it should have been `none` (i.e. should be consistent with other implementations) and another part of me says "NO, NO, NO, NO" - don't be so user unfriendly.

In this case, however, because this is a library only called by programs not accessed directly by the user, I think I agree with you. I will make revert this part of the PR tomorrow and fix it in svg-io.

I have another thing to fix - namely svg-io uses str to get points. So I need to create a root method xml:
```
def xml(self):
   return str(self)
```
and for Point override this so that format string is `.12G` as before and then change svg-io to use `xml` to format paths instead of `str`.

Also, I see that svg-io converts non-path Shapes to Paths before saving - I don't think it should do this.

tatarize:
It's really just that svgelements is rather particular. If you could get the same results without messing with it, it would be preferable.  It takes a lot to make sure that file stays correct and it has a pretty large testing library etc.

Sophist-UK:
Code is reverted.

I already decided that the way to fix the long numbers in element names (which is the issue) is to use a regex replace in the code that sets that to shorten the numbers.

Sophist-UK:
And apologies for screwing it up in the first place.

tatarize:
That section of code is it's own library so it just needs more protection than others. The repr fixes are generally perfectly fine and were most of the previous version after I added a tests to make sure they were done mostly correctly.

PR #487, opened 2021-08-12
Further fixes to __repr__ for Viewbox and SVGImage

Sophist-UK:
1. Viewbox creates a correct __repr__ showing a psuedo executable Viewbox call.
2. SVGImage adds missing attributes, puts the href (which is potentially huge) last and (I hope this is correct) puts the href attribute in quotations so that the end is easily parseable.

### DIFF
--- a/meerk40t/svgelements.py
+++ b/meerk40t/svgelements.py
@@ -3071,7 +3071,19 @@ class Viewbox:
         )
 
     def __repr__(self):
-        return "%s('%s')" % (self.__class__.__name__, str(self))
+        values = []
+        if self.x != 0:
+            values.append("x=%s" % Length.str(self.x))
+        if self.y != 0:
+            values.append("y=%s" % Length.str(self.y))
+        if self.width != 0:
+            values.append("width=%s" % Length.str(self.width))
+        if self.height != 0:
+            values.append("height=%s" % Length.str(self.height))
+        if self.preserve_aspect_ratio is not None:
+            values.append("%s=%s" % (SVG_ATTR_PRESERVEASPECTRATIO, self.preserve_aspect_ratio))
+        params = ", ".join(values)
+        return "Viewbox(%s)" % params
 
     def property_by_object(self, obj):
         self.x = obj.x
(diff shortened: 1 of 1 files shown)


tatarize (commented):
I tried the image_width image_height thing and they crashed or something weird.

I'm not sure how I didn't already have that viewbox in there. I swear I did something there, though I think it parsed it normally. Lemme check it.

tatarize:
It failed the test_repr_viewbox. 

```python
    def test_repr_viewbox(self):
        obj = Viewbox("0 0 100 60")
        print(repr(obj))
        self.assertTrue(obj == eval(repr(obj)))
        self.assertFalse(obj != eval(repr(obj)))
```

It also failed the test_Shapes_repr but that's pretty basic because it changed the way fill was put to include the `Color('#ff0000')` bit rather than just `'#ff0000'` which is fine, though it will cast that to a color correctly. The viewbox fail is relevant though. 

```
>   ???
E   TypeError: __init__() got an unexpected keyword argument 'width'

<string>:1: TypeError
```

My guess is that the viewbox doesn't parse the specific width, height, x, y values but rather just the viewbox bit as the 4 values.

tatarize:
Yeah, just takes a bit of a tweak to make `Viewbox(width=100, height=60)` valid, when it currently is not valid. Though if you want to make little tweaks to this why not submit the pull to https://github.com/meerk40t/svgelements ? I really can't imagine this shows up in meerk40t. I use viewboxes but just in a couple bits of code where I need to center things like align or centering the image in the camera window. It's a pretty limited use of converting a viewbox to center something like I want it.

Sophist-UK:
It shows up in MK when elements have no labels and no id. I can submit it to SVGelements repo too if you wish.

Sophist-UK:
> My guess is that the viewbox doesn't parse the specific width, height, x, y values but rather just the viewbox bit as the 4 values.

I will take a look at this and see if I can fix it.

Sophist-UK:
Closing as I have submitted these changes to svgelements repo instead.

PR #497, opened 2021-08-17
Correct Mouse Cursor Issues

tatarize:


tatarize:
I did some more work on this but something outside the code like the redraw routine or something is resetting the mouse. Makes it hard to diagnose. I recall there was a reason I gave up on this before.

Sophist-UK:
I tried again to track this down - but still failed. I am wondering whether it is:

a. Something to do with the Window Refresh/Update buffering
b. A bug in the wxPython code.
c. Another example of a global `except: pass`.

tatarize (commented):
Seems fine now.

Sophist-UK (approved):
I haven't tested it, but a code review looks OK (from my limited understanding of how MK works).

Issue #465, opened 2021-08-06
v0.7.0 RC8: Please report any issues found here


Sophist-UK:
# PLEASE DO ALL FURTHER TESTING ON v0.7.0b33 RC-8
The latest beta release is v0.7.0beta33 Release Candidate 8. This issue is for reporting issues with this specific version of MeerK40t.

Please do all further testing on this release. Please do not report any further bugs with previous beta releases. Do give any other types of feedback, questions, concerns or comments.

https://github.com/meerk40t/meerk40t/releases/tag/0.7.0-beta33

# CAUTION
**CAUTION:** Please note that (depending on your design) if you use Merge Operations after you have used Duplicate Operation, or you use Merge Passes when you have set Operation Passes > 1, these can result in the laser beam burning the same path several times in quick succession without allowing time for the material to cool down in between. This can have unintended consequences. **Please do not use this feature of the beta whilst your K40 is unattended.**

Sophist-UK:
Review using mock-USB or simulation:
# Bugs
1. With Optimise, spurious travel moves to SVG start/end of closed path still evident in simulation (and presumably in real cuts causing significant twitching). ![image](https://user-images.githubusercontent.com/3001893/128505226-67a351ce-1092-46ab-aacb-b9155b21dd79.png) In RC7, in burns like this (with many small circles) twitches accounted for (perhaps) 1/2 of the total time of the burn - some were due to these, but others were likely hardware related.

Sophist-UK:
@tatarize : I have not tracked this down as yet but I have a feeling that the cause of the spurious travel moves may be that when a closed path is flattened, the "Z" that marks the path as closed (which I think is a Close object in SVGelements) is being yielded as some sort of extra node - or possibly that SVGelements is not checking whether the start and end points of the segments are the same before inserting an extra segment to close the path.

Sophist-UK:
@tatarize Just ran it from source with the forward preference code and it no longer has these extra travel moves showing in simulation. But they are showing in RC8 so either a commit since the release has fixed this or it is something about versions being slightly different or a glitch in executables.

Sophist-UK:
# BUG
1. RC7 and RC8 are taking minutes to close when I have an SVG with a reasonable number of elements loaded. RC6 did not have this problem.

tatarize:
[...]

PR #498, opened 2021-08-18
Fix simulation render of Raster op...

Sophist-UK:
... which is step size too small.

### DIFF
--- a/meerk40t/gui/laserrender.py
+++ b/meerk40t/gui/laserrender.py
@@ -258,7 +258,11 @@ class LaserRender:
                     cache = None
                 if cache is None:
                     cut.c_width, cut.c_height = image.image.size
-                    cut.cache = self.make_thumbnail(image.image, dewhite=True)
+                    try:
+                        cut.cache = self.make_thumbnail(image.image, dewhite=True)
+                    except MemoryError:
+                        cut.cache = self.make_thumbnail(image.image, maximum=1000, dewhite=True)
+
                     cut.cache_id = id(image.image)
                 gc.DrawBitmap(cut.cache, 0, 0, cut.c_width, cut.c_height)
                 gc.PopState()
@@ -498,7 +502,7 @@ class LaserRender:
             width = int(round(width * scale))
             height = int(round(height * scale))
         if image_width != width or image_height != height:
-            pil_data = pil_data.copy().resize((width, height))
+            pil_data = pil_data.resize((width, height))
         else:
             pil_data = pil_data.copy()
         if dewhite:
--- a/meerk40t/gui/laserrender.py
+++ b/meerk40t/gui/laserrender.py
@@ -414,7 +414,9 @@ class LaserRender:
     ):
         """
         Make Raster turns an iterable of elements and a bounds into an image of the designated size, taking into account
-        the step size etc.
+        the step size. The physical pixels in the image is reduced by the step size then the matrix for the element is
+        scaled up by the same amount. This makes step size work like inverse dpi and correctly sets the image scale to
+        the step scale for 1:1 sizes independent of the scale.
 
         This function requires both wxPython and Pillow.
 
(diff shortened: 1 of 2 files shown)


Sophist-UK:
Visually this seems to work.

The alternative fix would be to leave these lines in and multiply scale by step a few lines later on. I tried this and it scaled correctly, but clipped the image to the smaller size.

tatarize:
This is in make raster. But, the raster is created correctly. The issue is in the simulation which should be in a different part of the visualization, I think.

Sophist-UK:
I cannot remember what route led me to here. I guess I can get the stack trace to see how we got to this method for the image - and then go back to see how it got called for simulation with step = 3 and see how to avoid calling make_image with the step.

tatarize:
I'll check it, but the simulation was wrong in #465 report. It actually lasered correctly. I think I heard some bug report about that before but couldn't duplicate it. It's like the built in scale isn't accounted for in the simulation. But, it does actually raster correctly. I think the issue is exclusive to sim.

tatarize:
Making it huge for the sim is helping me hit a bunch of memory and assertion errors. I can make it fail a bit more gracefully I think.

tatarize:
Confirmed. This does not solve the bug which is purely in simulation. This actually makes it raster the 4x larger area well outside the bed and consequently require absurd amounts of memory for the simulation. Which let me fix a bunch of those simulation bugs.

Sophist-UK:
Third attempt at commenting - hoping that Brave won't crash again.

Thinking about this I am sure you are right. But it has got me thinking (which is dangerous I know)...

1. For display in the simulation scene we should be able to reuse whatever internal representations for the Paths in the Raster ops that were displayed in the main scene.
2. We may need to actualise the raster for simulation in order to: A) do optimisations, and B) calculate e.g. burn time. But there may be ways of achieving this without actualising the raster. And if we do need to actualise it for simulation, we might be able to do it at a higher step than the op says anyway - only actualising it at the specified step when we do an actual burn.

meerk40t/gui/laserrender.py
```diff
@@ -500,7 +509,7 @@ def make_thumbnail(
             width = int(round(width * scale))
             height = int(round(height * scale))
         if image_width != width or image_height != height:
-            pil_data = pil_data.copy().resize((width, height))
+            pil_data = pil_data.resize((width, height))
```
Sophist-UK, line 512: Unclear why we remove .copy() here and leave the else copy below.

Sophist-UK:
P.S. I did test this fix before submitting it and I didn't see the issues you showed above.

meerk40t/gui/laserrender.py
tatarize: The resize() does the copy. When it was in giant mode it crashed there. The copy() is superfluous, since resize performs a copy into the resized element.

tatarize:
I added in the correct fix. The test file we have is perfect for this since it shows both the ImageOp and the Raster value. It permitted balancing one without messing up the other. Since there were two bugs in the code. First, I did not use the step size to determine the drawn size of the raster cut. And secondly I used the image matrix to determine the size. This means images in an ImageOp would be done correctly but Raster images would be done wrong. And fixing one would break the other. 

So, I got rid of all of that and only used the matrix for the image to give the X and Y values.

tatarize:
To test, verify that the simulation looks like the scene. And run a test (even just in mock mode) to see the rastering lines are actually matched up as well.

meerk40t/gui/laserrender.py
tatarize: The copy below is because a copy is still needed for the pil_data. But, in the case when you're resizing making a full copy then a smaller copy can kill the memory. You rather make a small copy. Then below when nothing was resized you just make a regular copy.

tatarize (approved):
I'm pretty sure my cutcode rendering fix corrects the error and this is fine to merge.

PR #503, opened 2021-08-19
Fix operations labels not being updated...

Sophist-UK:
... plus a couple of extremely minor fixes inc. reusing global MK version with run-from-source suffix for crash reports.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -512,19 +512,21 @@ class Node:
         return str(element)
 
     def set_label(self, name=None):
+        self.label = self.create_label(name)
+
+    def create_label(self, name=None):
         """
-        Set the name of this node to the name given.
+        Create a label for this node.
+        If a name is not specified either use a cascade (primarily for elements) or
+        use the string representation
         :param name: Name to be set for this node.
         :return: label
         """
-        if name is None:
-            if self.object is not None:
-                self.label = self.label_from_source_cascade(self.object)
-            else:
-                self.label = str(self)
-        else:
-            self.label = name
-        return self.label
+        if name is not None:
+            return name
+        if self.object is not None:
+            return self.label_from_source_cascade(self.object)
+        return str(self)
 
     def _flatten(self, node):
         """
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -461,14 +461,11 @@ class Node:
         node.notify_attached(node, pos=pos)
         return node
 
-    def label_from_source_cascade(self) -> str:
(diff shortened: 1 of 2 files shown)


meerk40t/gui/wxmeerk40t.py
```diff
@@ -3511,7 +3511,7 @@ def set_icon(self, node, icon=None):
 
     def update_label(self, node, force=False):
         if node.label is None or force:
-            node.label = node.label_from_source_cascade()
```
tatarize, line 3514: This doesn't seem like it would work. The reason we force this is so that demands an update. It seems like it won't cascade through the various methods without this.

tatarize: Also, why is it using set_label like it gets a label?

tatarize: This seems like it prevents the label updating in situations where "force" would be called. The flag there is by design since sometimes the label would already be set but a change means it needs to update.

Sophist-UK: I agree that the method is a bit confusing. We can easily create:
```
    def set_label(self, name):
        """
        Set the name of this node to the name given.
        :param name: Name to be set for this node.
        :return:
        :return: label
        """
        self.label = self.create_label(name)

     def create_label(self, name=None):
        if name is not None:
            return name
        if self.object is not None:
            return self.label_from_source_cascade(self.object)
        return str(self)
```

Sophist-UK: Then we can use create_label in wxmeerk40t.

tatarize: I also am not sure this fixes the force which is required when update is called to force the label to relabel.

Sophist-UK:
How about this?

meerk40t/core/elements.py
```diff
@@ -461,14 +461,11 @@ def add(self, data_object=None, type=None, label=None, pos=None):
         node.notify_attached(node, pos=pos)
         return node
 
-    def label_from_source_cascade(self) -> str:
+    def label_from_source_cascade(self, element) -> str:
```
tatarize, line 464: I think this will fall into the crashing again. element can sometimes be a operation and that needs to run correctly.

tatarize (approved):
I'm okay with this but I think it might still hit some issues.

That label stuff is more precarious than it should be.

Sophist-UK:
I have looked at the code, and I think there is a fix needed to avoid doing regex on non SVGelements.

But variable `elements` should probably also be changed.

PR #509, opened 2021-08-21
Improve mouse sizing icon area in scene

Sophist-UK:
1. If object is too small to reasonably have both sizing and move icons, then prioritise move.
2. Otherwise increase to size of the border area to up to twice the size but never more than 1/4 of the total.


tatarize (approved):
Should do what the comment suggests.

PR #510, opened 2021-08-21
Improve settings tooltips

Sophist-UK:


tatarize (approved):
This stuff is fine but the use of multiple `_()` commands means those lines translate each in their own call. Which means we need to have these all done before asking translators to finish them up or whatever.

Sophist-UK:
The purpose of doing it sentence by sentence is that length text strings are more likely to change than shorter ones.

There are other issues elsewhere of translation calls of the form: `_("file %s doesn't exist" % filename)` which should be `_("file %s doesn't exist") % filename`

PR #511, opened 2021-08-21
Raster Ops not matched to exact colors as default

Sophist-UK:
Backwards compatibility so that Raster Ops not matched exactly to any elements are treated as default.

Note: svg_io classified elements individually (which is suboptimal anyway) but does not work with this approach - consequently changes were needed with how various file loaders dealt with things - they now return either False or None (failed), True (succeeded but no file / elements in tree, or the file_node.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -4912,6 +4912,11 @@ class Elemental(Modifier):
         default_cut_ops = []
         default_engrave_ops = []
         default_raster_ops = []
+        new_ops = []
+
+        raster_ops = []
+        deferred_raster_elements = []
+
         for op in operations:
             if op.default:
                 if op.operation == "Cut":
@@ -4920,8 +4925,8 @@ class Elemental(Modifier):
                     default_engrave_ops.append(op)
                 if op.operation == "Raster":
                     default_raster_ops.append(op)
-
-        new_ops = []
+            if op.operation == "Raster":
+                raster_ops.append(op)
 
         for element in elements:
             if hasattr(element, "operation"):
@@ -4939,7 +4944,7 @@ class Elemental(Modifier):
 
 
             add_vector = False    # For everything except shapes
-            add_non_vector = True # Text - Dots and Images will also abuse this
+            add_non_vector = True # Text, Images and Dots
             if isinstance(element,Shape) and not is_dot:
                 if element_color.rgb == Color("black").rgb: # Treated as a raster for user convenience
                     add_vector = False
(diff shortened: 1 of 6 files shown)


meerk40t/device/lhystudios/lhystudiosemulator.py
```diff
@@ -327,4 +327,5 @@ def load(kernel, pathname, **kwargs):
                 "module/LhystudiossEmulator", basename
             )
             lhymicroemulator.write_header(f.read())
-            return [lhymicroemulator.cutcode], None, None, pathname, basename
+            kernel.root.close(basename)
+        return True
```
tatarize, line 331: I'm not sure this already added itself to the data tree.

meerk40t/core/elements.py
```diff
@@ -5052,13 +5065,16 @@ def load(self, pathname, **kwargs):
                 if str(pathname).lower().endswith(extensions):
                     try:
                         results = loader.load(self.context, self, pathname, **kwargs)
+                        if results and isinstance(results, Node):
+                            results.focus()
+                            self.classify([elem.object for elem in results.flat(types=("elem"))])
                     except FileNotFoundError:
```
tatarize, line 5071: Results is going to be pretty much always True here, since the loading function is returning that rather than the loaded data. I am not sure what this is otherwise supposed to do since results shouldn't be a node.

tatarize (approved):
It's iffy if the Ruida stuff loads correctly and I don't think that one function can ever return a node since it will return True basically always.

meerk40t/core/elements.py
Sophist-UK: I changed the loaders to return the `file node` if they create it and elements (svg, dxf, image) or `True` if they don't.

I could classify the elements in the loader and just return True like before, but at the time (late at night - judgement may have been poor) it seemed better than having the code in the loader itself. But on reflection I think it would be better to switch back and to simply find a better way to do classify in svg_io (which is not hard).

So I am not going to merge this - but will add more commits to redo it.

meerk40t/device/lhystudios/lhystudiosemulator.py
Sophist-UK: True means successful - a file_node means added to the tree.

Sophist-UK:
I did ask for this NOT to be merged.

tatarize:
Sorry, I checked didn't look to be a problem. Mark it as a draft I guess. Makes it clearer it's still a WIP.

tatarize:
Uppermost right corner. "Still in progress, mark as draft". The mark as draft is clickable.

tatarize:
Also, I sort of wanted to fix the egv thing, which was being modified by you at that very moment. See #514's overlapping changes.

Sophist-UK:
It was not that much of a problem. Was thinking of cancelling this one and starting again anyway rather than fixing this. So it made little difference. And yes, I should have marked it as draft.

PR #514, opened 2021-08-23
Add to and improve EGV functionality within the app.

tatarize:
* Correct EGV loading, crash


tatarize:
Goals:

* EGV blob type objects to be raw egv code to send to the laser.
* EGV convert to cutcode to be sent to any laser.
* Requires RawCut as an cutcode object parsed from the EGV data.

PR #518, opened 2021-08-24
Various inkscape improvements

Sophist-UK:


tatarize:
The thing inkscape needs most is the ability to save the current work and use inkscape to load it. Especially if it would save out the text as it appeared in the loaded file. Though I'm not sure that's very functional or reasonable.

Sophist-UK:
I am not suggesting that it needs to be done for 0.7.0, but we should probably try to make text work correctly for 0.7.1.

tatarize:
The version after this is a mirage. We're in infinite changes land. I wanted to test the moshiboard but my K40 is mostly out of order for a bit. I guess I could do it with just an x-axis, that or get a rotary. We might want to respect the feature freeze and stop treading water and just nail down the few required things for a next version and just publish it. It's way less buggy. In the last bug hunt I ended up making a bunch of old EGV stuff work. It's pretty few and far between for unfinished things. Though we might have added some bugs.

Sophist-UK:
I agree - we should not be adding new features. At this stage I think only bugs (which I think are primarily minor ones in Optimisation - though we need a definitive list) and usability (because you only make a first impression once) and language strings (because we don't want to be making wholesale i18n string changes after full release).

So my recommendation would be to focus on bug fixes first and put out a new RC when they are done, and we should be able to tweak usability and strings right up to full release.

PR #529, opened 2021-08-29
Fix typo in get_elem_node.

Sophist-UK:
`get_elem_node` is not currently used anywhere but this is a typo as it is currently same code as `get_elem`.


tatarize:
Thankfully it had zero uses in the codebase

PR #538, opened 2021-09-02
Re-engineered classification routine...

Sophist-UK:
... which classifies rasters only to existing raster ops if they are all the same colour ...
... or which classifies rasters to ops by colour keeping overlapping rasters grouped and classifying the group if any raster matches op color...
... and which classifies any rasters unmatched by colour into default raster ops if  they exist and if not into empty raster ops of any colour providing that they are all the same colour or if not into a new default raster op.

So you need to have at least two raster ops of different colour that don't match any raster elements before it creates new raster ops and leaves existing ones empty.

I think that this is about as backwards compatible as we can get and still have improved functionality.

And it is about 40 lines of code less than the previous version - and we don't treat white rasters as special case - both of which are IMO good signs of being on the right track.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -5256,12 +5256,9 @@ class Elemental(Modifier):
                     add_vector = element.stroke is not None and element.stroke.rgb is not None
                     add_non_vector = (
                         element.fill is not None
-                        and element.fill.rgb is not None   # Filled
-                        and not is_straight_line           # Cannot raster a straight line segment
-                        and (
-                            element.fill.alpha is None     # Not transparent
-                            or element.fill.alpha !=0
-                        )
+                        and element.fill.rgb is not None  # Filled
+                        and not is_straight_line          # Cannot raster a straight line segment
+                        and element.fill.alpha !=0        # Not transparent  
                     )
                     if add_vector and add_non_vector and element.stroke.rgb == element.fill.rgb: # Stroke same as fill - raster-only
                         add_vector = False
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -5334,6 +5334,8 @@ class Elemental(Modifier):
 
         # End loop "for element in elements"
 
+        print(len(raster_elements))
+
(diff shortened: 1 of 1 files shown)


meerk40t/core/elements.py
```diff
@@ -5140,94 +5140,115 @@ def classify(self, elements, operations=None, add_op_function=None):
...
+        rasters_one_pass = None
 
-        raster_ops = []
-        deferred_raster_elements = []
-        white_raster_elements = []
-        have_non_white_deferred_rasters = False
-
-        raster_ops_single_color = None
         for op in operations:
+            if op.operation in ["Cut", "Engrave"]:
```
tatarize, line 5212: Making this a list() is unneeded when a tuple() will do

tatarize (approved):

meerk40t/core/elements.py
```diff
@@ -5334,22 +5334,24 @@ def classify(self, elements, operations=None, add_op_function=None):
 
         # End loop "for element in elements"
 
+        print(len(raster_elements))
```
tatarize, line 5337: debug code

tatarize (changes requested):

meerk40t/core/elements.py
Sophist-UK: Grrrrr.

meerk40t/core/elements.py
Sophist-UK: Loads of other lists that didn't need to be mutable in the raster grouping code. So I fixed those too.

Sophist-UK:
I think I would like Joe to give this a try before it is merged.

tatarize (approved):

tatarize:
Also, should check if like 5241 crashes still if element.fill is `None` so checking the alpha == 0 causes a crash. Should be if stroke exists but fill is actually `None` then we hit a value error there. Seen live in RC10 crash logs.

tatarize (approved):

Sophist-UK:
I am going to merge this anyway, so we can get people to test it.

PR #542, opened 2021-09-05
Usability improvements...

Sophist-UK:
1. Better LaserOperation defaults depending on operation type
2. Improved Path label showing stroke and fill colours etc.
3. Translation string improvements moving substitutions out of translation
4. New Op Speed submenu
5. Rotate options more user friendly (and full list)
6. Add Operation now a sub-menu listing types
7. Only offer remove single item when a single item is selected.
8. Many other string improvements
9. Many other context menu improvements
10. Fix menu separators before/after submenus
11. Fix checked radio item when no radio item matches existing values
12. Fix expand after reclassification
13. Fix for missing code changes in commit https://github.com/meerk40t/meerk40t/commit/8c6a2126331c7c0a9f2ceb229e03d90618d71c98

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -4870,7 +4870,7 @@ class Elemental(Modifier):
         @self.tree_separator_before()
         @self.tree_operation(
             _("Expand all children"),
-            node_type=("branch elems", "branch ops", "group", "file", "root"),
+            node_type=("op", "branch elems", "branch ops", "group", "file", "root"),
             help="Expand all children of this given node.",
         )
         def expand_all_children(node, **kwgs):
@@ -4878,7 +4878,7 @@ class Elemental(Modifier):
 
         @self.tree_operation(
             _("Collapse all children"),
-            node_type=("branch elems", "branch ops", "group", "file", "root"),
+            node_type=("op", "branch elems", "branch ops", "group", "file", "root"),
             help="Collapse all children of this given node.",
         )
         def collapse_all_children(node, **kwgs):
@@ -5462,9 +5462,9 @@ class Elemental(Modifier):
                     add_vector = element.stroke is not None and element.stroke.rgb is not None
                     add_non_vector = (
                         element.fill is not None
-                        and element.fill.rgb is not None   # Filled
-                        and not is_straight_line           # Cannot raster a straight line segment
-                        and element.fill.alpha !=0         # Not Transparent
+                        and element.fill.rgb is not None  # Filled
+                        and not is_straight_line          # Cannot raster a straight line segment
+                        and element.fill.alpha !=0        # Not Transparent
                     )
                     if add_vector and add_non_vector and element.stroke.rgb == element.fill.rgb: # Stroke same as fill - raster-only
                         add_vector = False
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -66,6 +66,8 @@ def plugin(kernel, lifecycle=None):
 
 MILS_IN_MM = 39.3701
 
+# Regex expressions
+label_truncate_re = re.compile("((:|\().*)")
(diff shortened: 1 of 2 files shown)


meerk40t/core/elements.py
```diff
@@ -4520,18 +4668,18 @@ def reset_user_changes(node, copies=1, **kwgs):
...
         @self.tree_conditional(lambda node: isinstance(node.object, SVGElement))
         @self.tree_conditional_try(lambda node: not node.object.lock)
         @self.tree_submenu(_("Scale"))
-        @self.tree_iterate("scale", 1, 25)
+        @self.tree_iterate("scale", 25, 1, -1)
         @self.tree_calc("scale_percent", lambda i: "%0.f" % (600.0 / float(i)))
         @self.tree_operation(
-            _("Scale {scale_percent}%"), node_type="elem", help="Scale Element"
+            _("Scale %s%%") % "{scale_percent}", node_type="elem", help="Scale Element"
         )
```
tatarize, line 4683: Hm. I thought this was silly, but I see the need to keep that bit of code away from the translators since it's hardcoded functional.

meerk40t/core/elements.py
```diff
@@ -4547,21 +4695,23 @@ def scale_elem_amount(node, scale, **kwgs):
...
+                -1,-2,-3,-4,-5,-6,-7,-8,-9,-10,-15,-20,-30,-45,-60,-90,-120,-135,-150
+            )
         )
-        @self.tree_operation(_(u"Rotate turn/{i}, {angle}°"), node_type="elem", help="")
-        def rotate_elem_amount(node, i, **kwgs):
-            value = 1.0 / float(i)
+        @self.tree_operation(_(u"Rotate %s°") % ("{angle}"), node_type="elem", help="")
+        def rotate_elem_amount(node, angle, **kwgs):
+            turns = float(angle) / 360.0
             child_objects = Group()
```
tatarize, line 4707: My beautiful taus!

meerk40t/gui/wxmeerk40t.py
```diff
@@ -3348,6 +3349,8 @@ def collapse(self, node):
         if not item.IsOk():
             raise ValueError("Bad Item")
         self.wxtree.CollapseAllChildren(item)
+        if self.wxtree.GetItemParent(item) == self.wxtree.GetRootItem():
+            self.wxtree.Expand(item)
```
tatarize, line 3354: Hey, they collapsed the root, they should live with it.

tatarize (approved):
Seems to be quite solid

meerk40t/core/elements.py
Sophist-UK: Actually I think I was wrong in concept - I think the translator would translate "Scale {scale_percent}" to (say) "Size {scale_percent}" and the variable substitution would happen afterwards. (I was thinking of `_("Scale %s%" % scale_percent)` vs. `_("Scale %s%") % scale_percent` when I was thinking that this would be an issue.)

meerk40t/core/elements.py
Sophist-UK: This was A) Horrible to use - if you wanted 38 degrees then finding a combination of rotations to get there was nigh on impossible and B) broken anyway - the list was a fraction of the size it should have been.

meerk40t/gui/wxmeerk40t.py
Sophist-UK: That is why Open Source is so great. You get collaboration between heroic coder geniuses who don't give a hoot about non-geniuses screwing things up, and mere mortals who know that you need to put guard rails up for the less tech savvy users.

Sophist-UK:
P.S. I really wasn't sure about the additional changes for the commit you had previously made - I was only going on consistency of approach and didn't take the time to understand the code.

PR #551, opened 2021-09-09
Operation icons in color

Sophist-UK:
Allows operation color to be applied to the icons, replacing black with the color. Grey scale is handled by transparency.

The idea was to use icon color to reflect the colour of the operation / element rather than coloring the text (because it is hard to read white text on a white background).

However, if the operation color is close to the tree background color (assumed to be white in normal mode and black in dark mode) this will not contrast well.  So we replace the transparency with a white or black background and removing the alpha channel.

### Notes
1. When I tested this, I found that the icons do NOT have the same impact as text color - so we probably need to do something similar with text color i.e. if close to the background color then change the background or use a different means to show the color; and
2. The text color appears to be set in multiple places using SetItemTextColour and in the wxtree and shadow-tree, and I am not sure which of these to turn off.

### DIFF
--- a/meerk40t/gui/wxmeerk40t.py
+++ b/meerk40t/gui/wxmeerk40t.py
@@ -327,9 +327,9 @@ _ = wx.GetTranslation
 supported_languages = (
     ("en", u"English", wx.LANGUAGE_ENGLISH),
     ("it", u"italiano", wx.LANGUAGE_ITALIAN),
-    ("fr", u"français", wx.LANGUAGE_FRENCH),
+    ("fr", u"franÃ§ais", wx.LANGUAGE_FRENCH),
     ("de", u"Deutsch", wx.LANGUAGE_GERMAN),
-    ("es", u"español", wx.LANGUAGE_SPANISH),
+    ("es", u"espaÃ±ol", wx.LANGUAGE_SPANISH),
     ("zh", u"Chinese", wx.LANGUAGE_CHINESE),
 )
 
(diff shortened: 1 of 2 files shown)


meerk40t/gui/wxmeerk40t.py
```diff
@@ -327,9 +327,9 @@ def gui_start(**kwargs):
 supported_languages = (
     ("en", u"English", wx.LANGUAGE_ENGLISH),
     ("it", u"italiano", wx.LANGUAGE_ITALIAN),
-    ("fr", u"fran�ais", wx.LANGUAGE_FRENCH),
+    ("fr", u"français", wx.LANGUAGE_FRENCH),
     ("de", u"Deutsch", wx.LANGUAGE_GERMAN),
-    ("es", u"espa�ol", wx.LANGUAGE_SPANISH),
+    ("es", u"español", wx.LANGUAGE_SPANISH),
     ("zh", u"Chinese", wx.LANGUAGE_CHINESE),
```
tatarize, line 333: What changed here? The non utf-8 characters have a nasty habit of changing to a different broken set of characters so these changes concern me like it's trying to use unicode for these in a breaking fashion.

tatarize (approved):

meerk40t/gui/wxmeerk40t.py
Sophist-UK: Sorry - I didn't change these lines myself. I think Github changed it to UTF when I edited it in a browser - and I don't know any way of stopping this.

I am not unclear why we are using ISO 8859-1 rather than UTF anyway.

Sophist-UK: I will leave you to correct this and merge it.

Sophist-UK: It is on my ToDo list to stop using Github web pages and point my Git at my run-from-source directory. I just haven't got around it it yet.

tatarize: I'm fine with anything that actually shows the letters rather than the messed up thing they turn into. I'll likely be checking this prior to new versions though. Also hit this problem when I tried to use a tau with regard to angles in the rotate.

PR #562, opened 2021-09-14
Update laserspeed.py

dpticar:


tatarize (approved):
This is entirely accurate. The value wasn't used for anything so it wasn't detected sooner. This clearly did miscalculate the multiplied value in there.

tatarize:
@dpticar great catch

PR #571, opened 2021-09-18
0.6.24 (Legacy)

tatarize:
- [X] Disable Windows Computer Sleep
- [ ] Must be tested in Windows


Sophist-UK:
As previously discussed the changes to Classify in 0.6.23 were to:

a. Fix images being classified multiple times
b. Classify everything

These were **bugs** that need fixing on 0.6.

Sophist-UK:
A desk review of the code LGTM. I don't use sleep on my laptop, so I have not tried to test this.

tatarize:
It's somewhat annoying to test correctly. Set sleep to 1 minute, start a raster that takes longer than a minute, make sure it finishes and then the computer quickly goes to sleep after it finishes.

PR #573, opened 2021-09-24
Add Legacy Classify Option, circa RC1

tatarize:
Uses RC1 Classification code and gives a checkbox in settings to activate the legacy code instead of the modern code.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -1715,6 +1715,7 @@ class Elemental(Modifier):
         bed_dim.setting(int, "bed_width", 310)
         bed_dim.setting(int, "bed_height", 210)
         context.root.setting(bool, "classify_reverse", False)
+        context.root.setting(bool, "legacy_classification", False)
 
         # ==========
         # OPERATION BASE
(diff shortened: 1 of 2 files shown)


meerk40t/core/elements.py
```diff
@@ -5540,6 +5541,79 @@ def contains(box, x, y=None):
...
+                add_op_function(element)
+                continue
+            if element is None:
+                continue
+            for op in operations:
+                if op.operation == "Raster":
+                    if element.stroke is not None and op.color == abs(element.stroke):
+                        op.add(element, type="opnode")
+                        was_classified = True
+                    elif isinstance(element, SVGImage):
```
Sophist-UK, line 5578: This `elif` section should be removed. It is a bug that results in images in Raster ops.

meerk40t/core/elements.py
```diff
@@ -5540,6 +5541,79 @@ def contains(box, x, y=None):
...
+                        op.add(element, type="opnode")
+                        was_classified = True
+                    elif isinstance(element, SVGText):
```
Sophist-UK, line 5581: This should be left as-is for compatibility but it adds a text element with no fill and very thin stroke to a raster operation which is not really correct.

meerk40t/core/elements.py
```diff
@@ -5540,6 +5541,79 @@ def contains(box, x, y=None):
...
+                    and element.stroke is not None
+                    and op.color == abs(element.stroke)
+                ):
+                    op.add(element, type="opnode")
+                    was_classified = True
+                elif op.operation == "Image" and isinstance(element, SVGImage):
+                    op.add(element, type="opnode")
+                    was_classified = True
+                    break  # May only classify in one image operation.
+                elif (
```
Sophist-UK, line 5598: This should use the isDots to determine whether they are dots as there is more than one way to get a dot.

meerk40t/core/elements.py
```diff
@@ -5540,6 +5541,79 @@ def contains(box, x, y=None):
...
+                    was_classified = True
+                    break  # May only classify in Dots.
+            if not was_classified:
+                if element.stroke is not None and element.stroke.value is not None:
+                    op = LaserOperation(
+                        operation="Engrave", color=element.stroke, speed=35.0
+                    )
+                    add_op_function(op)
+                    op.add(element, type="opnode")
+                    operations.append(op)
```
Sophist-UK, line 5615: We need to add disabled (output = False) operations for image or Dots or raster if element has not been classified.

meerk40t/core/elements.py
```diff
@@ -5540,6 +5541,79 @@ def contains(box, x, y=None):
...
+                    elif isinstance(element, SVGText):
+                        op.add(element)
+                        was_classified = True
+                    elif element.fill is not None and element.fill.argb is not None:
+                        op.add(element, type="opnode")
+                        was_classified = True
+                elif (
+                    op.operation in ("Engrave", "Cut")
+                    and element.stroke is not None
+                    and op.color == abs(element.stroke)
```
Sophist-UK, line 5590: Should also check `and op.default == False`.

meerk40t/gui/settings.py
```diff
@@ -136,6 +136,14 @@ def __init__(self, *args, **kwds):
                 "classify_reverse",
                 False,
             ),
+            (
+                _("Legacy Classify"),
+                _(
+                    "Use the legacy classification algorithm rather than the modern classification algorithm."
```
Sophist-UK, line 142:
```
_(
    "Use the legacy classification algorithm rather than the modern classification algorithm."
) + "\n\n"
+ _(
    "Legacy classification is compatible with Meerk40t 0.6.22+."
) + "\n"
+ _(
    "Modern classification has better handling of raster paths."
)
```

meerk40t/core/elements.py
Sophist-UK: If there is no existing Dots operation, this will classify dots to an Engrave operation.

meerk40t/core/elements.py
tatarize: That only happens if there is no image op. Which seems like the best choice in that case. You might want the image.

meerk40t/core/elements.py
tatarize: I can't really do much else with them and it's a one text fits all thing. So all text is treated the same way regardless of other factors.

meerk40t/core/elements.py
tatarize: Can't legacy code.

meerk40t/core/elements.py
Sophist-UK: Nope. Put the raster operation first and watch it duplicate. Create a disabled Image op if need be.

meerk40t/core/elements.py
Sophist-UK: I did say it needed to be left. Just pointing out a use case where this legacy approach is incorrect - as I keep saying there are lots of use cases and getting them all correct is impossible, and getting most of them correct in a consistent way is bloody difficult.

meerk40t/core/elements.py
Sophist-UK: Dots is NOT in the legacy code. Anywhere. It is an MK7 only thing. And maintaining compatibility with a barely functional 0.7 RC1 used only by a couple of hard-core testers is completely different to maintaining compatibility with 0.6.2x which was widely used for a long time by the MK community.

PR #572, opened 2021-09-24
Reimplement colour matching

Sophist-UK:
Filled items match exact color vector ops.
Unfilled items match exact color raster ops.

PR #574, opened 2021-09-25
Part revert #572 and fix #573

Sophist-UK:


tatarize (approved):
Yeah, these all check out. Took longer than changing nothing but the modifications even to legacy are fine.

PR #575, opened 2021-09-25
Retrofit 0.7 legacy classification improvements to 0.6.24

Sophist-UK:


tatarize (commented):
The break change must have been added in 0.7.x somewhere. I'm not sure LaserOperations() work the same (they are not nodes in 0.6.x) and I'll have to do a fuller check of this code because of the requirements for compatibility and high need of success. And these improvements might be somewhat features which aren't really to be added.

Sophist-UK:
I didn't just copy and paste the code. I adapted it to mimic existing code and then tested it.

tatarize (approved):
This looks safe, and integration into 0.6.24 is the suspect part, integrating into the 0.6.24 spec version is not a problem

PR #578, opened 2021-09-26
Implement code quality workflow

Sophist-UK:
This PR creates 3 workflows:

1. unittest - the workflow is working, and I have hacked specific tests to stop them hanging part way through the tests, however because unittest still hangs at the end of the tests, I have disabled this workflow until such a time as unit tests have been fixed.
2. Python code checking - a single job that installs all prerequisite software and runs a variety of code checkers against the meerk40t code (but not unittest code). Output from Pylint and flake8 are saved as downloadable artifacts, black and isort are in the console output from the workflow. mypy is disabled because we do not use strict typing yet in MK. These jobs will currently report success in all cases as exit codes are overwritten.
3. CodeQL - Github's code security code checker that Picard uses. This will probably only produce results when this PR has been merged.

## Next steps once this PR is merged
1. Implement same for svgelements
2. Find a way to save black and isort output as artifacts
3. Confirm whether code checking needs MK preqs installed (suspect it does)
4. Examine output and tweak relevant config files to eliminate unwanted (considered false) negatives.
5. Fix remaining code issues.
6. Fix unit tests and re-enable workflow.


Sophist-UK:
Ok - this is ready for merge. :-)

tatarize (approved):

Sophist-UK:
@tatarize Thanks for approving this. Can you please provide me with your black / isort settings so I can tweak the github ones to match.

If you get any time to review the results of Pylint and flake8 and let me have a list of errors codes you want to accept I will add those too. I won't copy the workflows to svgelements until we have done this to save duplicated effort.

CodeQL attempts to install python prerequisites using requirements.txt and wxpython fails. So we need to include the build code from other yml files instead. Since this will not be necessary for svgelements I will do this once I have cribbed the yaml files for that.

Sophist-UK:
P.S. My personal view is that requirements.txt should contain all pre-requisite packages for full functionality running from source. And we should then have alternatives for e.g. cut down without DXF and camera (for e.g. Pi with limited disk space) and for graphicless remote server builds.

tatarize:
No settings. Just black and isort. They are not things with options. They basically always do the same thing.

Sophist-UK:
Actually they do have settings. maximum linelength for example, and isort has a setting for how to reformat import statements.

tatarize:
I think they have some default settings and those will do. It seems like it it works we could change things from that point on.

Sophist-UK:
The workflow certainly works. At present it is set to complete successfully regardless, because until we get a clean run for current code we cannot tell if a PR creates new code quality issues.

So we need to get all checks to run clean so we can get feedback for each PR - for Pylint and flake8 that will mostly be by filtering out checks that we are not interested in but possibly with some code formatting, and for black / isort (where you have these in your IDE) we probably only need to set the max line length to the same values as you use) for them to run clean.

Once we have these running cleanly I will install and document a Pycharm and a VsCode IDE with matching tools so that developers can check the code before they submit it.

tatarize:
> P.S. My personal view is that requirements.txt should contain all pre-requisite packages for full functionality running from source. And we should then have alternatives for e.g. cut down without DXF and camera (for e.g. Pi with limited disk space) and for graphicless remote server builds and for development. What do you think?

It's not just limited disk space, it tries to install that stuff and wxPython and openCV can do weird things like start compiling a thing for days before crashing due to lack of memory. The current system allows `pip install meerk40t[all]` if you want that. it goes not by the requirements file but the setup.cfg file. 

The requirements.txt I think could be expanded and it would do things without messing stuff up, because the cfg is what pip uses to install things. But, there's already some alternative installs. Also, the goal there is to install the needed bits and the smallest set of those is best. Else you have total requirements. If I wrote a neat little script that homed the laser and it worked by loading up meerk40t and running it with `-Ze "home"` it would require all of OpenCV and wxPython to do that, and would need them installed even though they *never* use them.

tatarize:
@Sophist-UK Can it be merged anyway with failing tests such that we can leave the tests on even though we might fail them to get the reports in where it doesn't quite work and then succeed?

Sophist-UK:
It is merged. It is disabled in Github Actions web page - so doesn't need a merge to turn it on.

tatarize:
I fixed the tests with a PR, should pass now, and not hang.

tatarize:
PR might need approval and merging.

PR #581, opened 2021-09-28
Add link to beginners wiki

Sophist-UK:
@tatarize  - I have no idea what is causing it, but this change stops the webhelp functionality from launching a web page.

Works on master branch, doesn't work on this branch.

Can you help?

### DIFF
--- a/meerk40t/core/webhelp.py
+++ b/meerk40t/core/webhelp.py
@@ -1,10 +1,13 @@
-MEERK40T_ISSUES = "https://github.com/meerk40t/meerk40t/issues"
-MEERK40T_HELP = "https://github.com/meerk40t/meerk40t/wiki"
-MEERK40T_WEBSITE = "https://github.com/meerk40t/meerk40t"
-MEERK40T_RELEASES = "https://github.com/meerk40t/meerk40t/releases"
-FACEBOOK_MEERK40t = "https://www.facebook.com/groups/716000085655097"
+import requests
+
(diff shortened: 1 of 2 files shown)


tatarize:
Remove requests from the import there it fails. All the routines that work like mini-plugins do not load if there's an import error. So the dxf-loading stuff will fail if ezdxf fails, and this means that it does not load. So it just silently removes DXF functions. The same is happening here. the `import requests` line that suddenly shows up there is suffering an import error which aborts loading the plugin.

Sophist-UK:
Ah. No error message even when there is an attached console. Usually you run without an attached console so such messages will not be seen by the user.

That was a line I left in for making the Github API requests.

Sophist-UK:
Or at least do so when developer mode is on.

tatarize (approved):
Perfectly fine set of changes. No potential bugs. That formatting will likely get undone by applying Black tho.

Sophist-UK:
No - the black code check did not pick up anything new. Just the existing 3 issues.

Sophist-UK:
Default black line length is 88 and is not overridden. I will need to adjust all other tools to use the same line length,

Also, I am going to look at switching to pyproject.toml as a common configuration file if that is OK with you.

How much checking do you want to happen on the unittest .py files?

tatarize:
I'm mostly concerned with the code and features set, how it's setup on github etc doesn't matter that much to me, so long as it can still upload to pypi, I'm fine with such things.

PR #584, opened 2021-09-29
Implement custom import exception

Sophist-UK:
Rather than ignore all ImportErrors, this implements a custom exception which can be explicitly used for optional imports so that failing imports due e.g. to coding errors are not hidden from the developer.

### DIFF
--- a/meerk40t/core/exceptions.py
+++ b/meerk40t/core/exceptions.py
@@ -1,0 +1,15 @@
+# Define meerk40t specific exceptions
+
+# Base meerk40t exceptions
+class Mk40tError(Exception):
+    pass
+
+class Mk40tImportAbort(ImportError, Mk40tError):
+    """
+    MkImportAbort should be used as follows in plugins that import an optional prerequisite Pypi package:
+
+    try:
+        import wx
+    except ImportError as e:
+        raise Mk40tImportAbort("wx") from e
+    """
--- a/meerk40t/dxf/dxf_io.py
+++ b/meerk40t/dxf/dxf_io.py
@@ -1,6 +1,12 @@
 import os
 
-import ezdxf
+from ..core.exceptions import Mk40tImportAbort
+
+try:
+    import ezdxf
+except ImportError as e:
+    raise Mk40tImportAbort("ezdxf")
+
 from ezdxf import units
 
 try:
--- a/meerk40t/main.py
+++ b/meerk40t/main.py
@@ -3,6 +3,7 @@ import asyncio
 import sys
 
 from .kernel import Kernel
+from .core.exceptions import Mk40tImportAbort
 
 try:
     from math import tau
@@ -95,7 +96,6 @@ parser.add_argument(
     help="Do not load meerk40t.plugins entrypoints",
 )
 
-
 def run():
     argv = sys.argv[1:]
     args = parser.parse_args(argv)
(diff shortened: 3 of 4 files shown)


meerk40t/main.py
```diff
@@ -123,179 +123,182 @@ def run():
     """
     try:
         from . import kernelserver
-
-        kernel.add_plugin(kernelserver.plugin)
-    except ImportError:
+    except Mk40tImportAbort:
         pass
+    else:
+        kernel.add_plugin(kernelserver.plugin)
```
tatarize, line 130: kernelsever plugin will *never* raise MK40tImportAbort that will simply never happens

This looks sort of more polite but if the newer paradigm is crash-harder we might as well just move this out of the try altogether.

tatarize (changes requested):

tatarize:
Yeah reviewing this there's a lot of moving parts I hadn't considered previously.

Currently plugins are a lot cleaner and they don't as a matter of course actually require MeerK40t. They just don't really have any functionality without meerk40t loaded. Though throwing their import errors properly hard is likely preferred and I think it's the default. The PackageDistribution error is different than ImportError.

There's also the frozen imports vs dynamic imports with camera it's loaded twice because frozen needs to hard code that to hardcode which imports get overtly added.

While this is a great start I feel like maybe we'd be better off just removing the `try` from all the non-optional hard plugins and print something if the optional internals are missing or something. And dying hard with anything in the dynamics plugins that doesn't load. There's a weird amount of history that goes into this problem.

1. Dynamic plugins must work. These can go ahead and have import errors that cause failures.
2. Frozen plugins must have a direct import command. This is because these must load in pyinstaller.
3. The camera plugin is both, and gets add-attempted twice. this is because it's an optional plugin but always included in frozen builds.
4. The plugins shouldn't really often require meerk40t things, beyond a copy of kernel object.
5. Dynamic plugins shouldn't actually require a method of conditional failure. They are installed or not installed. They should have their requirements.

Your solution has advantages over the other solution. Namely you only throw the extra optional error in the case that the import of the requirement fails. So if other code has failed like what happened with ezdxf changing their import structure the other week. It will throw a correct import error and the bug will be quickly detected. The ezdxf import fail was within the actual version of ezdxf used being greater than `0.16.4` but the ezdxf import itself was loaded. So the better failure would be as you have it. 

I think maybe your solution is right. But, we shouldn't expect dynamic plugin authors to use the exception. It should be  a pure meerk40t core exception. Since you can just use pip to uninstall the plugins if you don't have the requirements.

There also might be a some need to support windows only plugins. The PR for #583 does this, since it doesn't actually need to run any of this code outside of windows it defines a plugin that only loads within windows. I'm not sure this stuff would need a helper function or methods though.

meerk40t/main.py
```diff
@@ -123,179 +123,182 @@ def run():
...
 
     try:
         from .extra import embroider
-
-        kernel.add_plugin(embroider.plugin)
-    except ImportError:
+    except Mk40tImportAbort:
         pass
+    else:
+        kernel.add_plugin(embroider.plugin)
```
tatarize, line 263: With embroider I'm actually a little more inclined towards the try since I think there's little problem simply deleting that plugin file and it would just be gone. Though the hard-import for frozen would stay and fail regularly. Most of the stuff in extra is supposed somewhat failable. Since really they aren't supposed to do anything mission critical.

meerk40t/core/exceptions.py
```diff
@@ -0,0 +1,15 @@
+# Define meerk40t specific exceptions
+
+# Base meerk40t exceptions
+class Mk40tError(Exception):
+    pass
+
+class Mk40tImportAbort(ImportError, Mk40tError):
+    """
```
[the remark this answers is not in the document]
Sophist-UK: I think that this name is pretty descriptive - it flags an Import issue as an Abort not an error.

meerk40t/dxf/dxf_io.py
```diff
@@ -1,6 +1,12 @@
 import os
 
-import ezdxf
+from ..core.exceptions import Mk40tImportAbort
```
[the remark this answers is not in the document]
Sophist-UK: We could easily move it from `meerk40t/core/` to `meerk40t/`.

meerk40t/main.py
Sophist-UK: You said that you didn't want main.py to know anything about whether plugins are optional (and abort when their prerequisite packages are missing). I have no issues with removing the try/except from all but those frozen plugins known to need it.

Sophist-UK: The new paradigm is "Don't hide genuine error conditions, but instead allow the exceptions to 'crash' MK - only hide highly specific exceptions that are known to be OK."

Sophist-UK:
Unless I hear otherwise, I propose to:

1. Remove the try blocks from frozen plugins known **not** to have Pypi prerequisites.
2. Move exceptions.py up one level to `meerk40t/`.
3. Remove the Mk40tImportAbort from the dynamic plugin loads.

tatarize:
Go ahead and leave the exceptions.py in core. Really moving it up a level would require a rename since in theory the kernel stuff is a bit less meerk40t centric.

tatarize:
This would mostly result is removing the try blocks for nearly everything except those couple things where we know there's a soft requirement. And in that case catch the custom exception that means that that soft requirement is met or not. That seems generally pretty reasonable.

tatarize:
> Why is the camera plugin not simply part of core MK?

Because if it was part of core MK then MK would require OpenCV. While there are ways to get around some of the requirements and maybe just include the camera py stuff into main, that wouldn't be too much of a problem. The stuff that requires an additional requirement can and likely should be spun-off into an addon. The only ones where this could be done were the optional requirements so `dxf`, `gui` and `camera` and it was only actually done with camera.

It was deemed to be too destructive and problematic to do this with the gui and really sort of pointless with something as small as  `dxf` which equally could have been shoehorned into the same sort of methods. And issues with compiling caused a partial retrograde of these changes.

But, the core idea is the same there, camera is its own module rather than simply part of main. This was done after branching into plugins and partially to show the power of plugins and partially to test them. It was also clear that opencv the code that came with that was rather large and without the camera we could compile things at about half the size. So we could publish a camless version that would come in at 25 megs rather than 50 megs, though we never really did that. The only actual difference between the version was that we could uninstall the `meerk40t-camera` plugin.

At that time we also experimented with having all the internal plugins be entry_points within the config. This simplified to main to just the dynamic plugin bit of code. But, made compiling it impossible. It was from this that we realized we needed some hard hooks to compile core plugins with optional plugins. Camera was already spun off because it was the most reasonable to do so, and it still remains spun off to this day. And all the other modules that could be spun off were left in actual modules that loaded the code. So they all switched to plugin style code

Sophist-UK:
I can understand the rationale, however I am wondering whether it should be revisited: 

1. If we get over worried about file size, we are going to have an even larger variety of executables that we would otherwise have.
1. 25MB file size is not a huge difference for most PCs these days. So the existing executable version with camera meets 99%+ of users needs. Given the small number of people who might want it without, installing Python and using the Pypi or Homebrew versions would meet the needs of those whose PCs are too small - and if the camera module were included in MK proper, that will be a small amount of code, and the user could choose when installing through Homebrew or Pypi whether to install opencv or not.
1. Merging it would simplify both the code and the build process.
1. If we are going to have alternative executables, my vote would be to have a cut-down install-and-run remote server version without the GUI - because I think that in the future this could be the second largest use case.

tatarize:
It's not just the file size but the requirements to include these. The number of requirements should be kept as low as possible since that tends to makes things generally easier and better. There was a need to make plugins and camera seemed the most outside the standard functionality of the code.  But, issues with compiling made it a bit of a rethink, and few other bits of code required too much additional code so camera was made it's own thing. Bringing it back into the fold isn't really needed. The only other major aspect was gui but moving that outside the fold would be a pain, since about half the things being edited are going to be part of the gui. And a lot of times those are edited together.

Sophist-UK:
We could have the camera plugin included, and still check for opencv (headed or headless) much as we do for e.g. ezdxf.

But it is working ATM so this could be on the back burner for a suitable opportunity.

PR #585, opened 2021-09-30
Remove camhead option

Sophist-UK:
Headless is sufficient for MK.
Most users these days use a virtual environment so will not need headed for other projects.
If user needs headed they probably have skills to install it.
Giving user a headed option as standard is a choice they don't need IMO.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,5 @@ setup(
         'gui': ['wxPython>=4.0.0', "Pillow>=7.0.0"],
         'cam': ["meerk40t_camera", "opencv-python-headless>=3.4.0.0"],
         'dxf': ["ezdxf>=0.14.0"],
-        'camhead': ["meerk40t_camera", "opencv-python>=3.4.0.0"],
     }
 )


tiger12506:
That option was added for a reason. On some platforms (such as pi) it is not feasible (maybe not possible) to compile headed opencv, so having the option is important. Is there some other reasoning behind removing it?

tatarize:
Yeah, it's a pointless bifurcation of choice but this comes from OpenCV. 

If you already have the headed version installed demanding the headless version is going to require both version which is explicitly what OpenCV said do not do.

Inpain used to use the headed version and I we really can't make that decision for the end user. If we require OpenCV we must accept either the headed or headless version. And while headless is enough that is why it's called cam, the regular version with the gui stuff included gets camhead which is clearly more cryptic but also perfectly allowed.

We clearly prefer headless afterall, `cam` and `all` go with the headless version, but the full version with stuff we don't use is fine and can be included in a limited number of cases where the user has some requirements or a pre-established requirement to use one version rather than a different version. Tiger's also right that there's different support within opencv for the different version and it might be impossible to meet the requirements of one without meeting it for the other.

tatarize:
This should probably closed without merging unless there's a more robust rationale for the change.

PR #586, opened 2021-10-01
Process checksum-fail command.

tatarize:
* Permits the use of `%` in the lhymicro-gl code as a special command for the controller.
* The use of % does the checksum xor with 0xFF
* The rejection of the packet with a 207 error is ignored, and said to be correct.
* No retry attempt is made.

To test this `egv ~IPP%$~` and you can see if you `channel open 0/usb_send` and `channel open 0/recv` that this was just rejected. You can do this a bunch in the middle of actual values rastering processes etc and it won't process any of that stuff. Spaces get a bit weird and it might actually send the data correctly for one packet if you exceed 30 digits and it needs to break that data into different bits.

### DIFF
--- a/meerk40t/device/lhystudios/lhystudiosdevice.py
+++ b/meerk40t/device/lhystudios/lhystudiosdevice.py
@@ -1950,6 +1950,7 @@ class LhystudiosController:
         * : clear the buffers, and abort the thread.
         ! : pause.
         & : resume.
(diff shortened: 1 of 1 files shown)


tiger12506 (approved):
I did not personally test this, but Tat says he did, and I see nothing in the code that looks dangerous.

PR #600, opened 2021-10-08
Improve image actualization.

Sophist-UK:
This code is based on a discussion between Sophist and @Taterize on IRC on how to handle image actualisation.

I have not yet tested this code, but am pushing this to a PR so that @Taterize can take a look at the code concept.

### DIFF
--- a/meerk40t/image/imagetools.py
+++ b/meerk40t/image/imagetools.py
@@ -1667,9 +1667,8 @@ class RasterScripts:
                                     # image = image.convert("L")
                                 empty_mask = image.point(lambda e: 0 if e == 0 else 255)
                                 image = ImageOps.invert(image)
-                            elif alpha_mask:
-                                # When NOT inverting we can use the alpha mask
-                                empty_mask = alpha_mask.point(
+                            else:
+                                empty_mask = image.point(
                                     lambda e: 0 if e == 255 else 255
                                 )
                         except (KeyError, OSError):


Sophist-UK:
Current status:

1. I believe that actualize is complete and working. I am unclear whether there are any issues with converting image.modes like P to RGB or PA to RGBA when using actualize, but my gut reaction is not.

2. Right-click actualize seems to be working great.

3. Raster scripts are still converting transparent pixels to white unnecessarily. 

4. Simulation (or LaserRender) is not calling actualize but using wx transformations instead and (despite using wxGraphicContext) these are still using black non-transparent fill for actualization of rotated matrices, and are also converting transparent to black (where the image has already been actualised). 

5. Burn is probably using actualize - there is a different actualize in `CutPlan` which calls `make_actual` which calls the real `actualize`. This could potentially be simplified - make_actual would be better in imagetools and given the small number of places it is call it could probably be eliminated in favour of the real actualize. Also perhaps sensible to change the name of one to prevent confusion.

Sophist-UK:
> I'm biased but I think we should consider #601 over this one.

I am not biased and I have no issues considering #601.

I will need to try the #601 code against my test cases, but I wonder whether:

a. We should reduce coloured images to greyscale every time - I have no idea either way but I shied away from going this far with my code but if you think it is OK, then it is fine with me;
b. If we are converting to "L", whether the code should mimic the greyscale code in RasterScripts;
c. Whether we should aim for an LA mode which preserves alpha rather than convert transparent to white (which I would personally like to see because you don't end up with white background overlaying other elements);
d. Whether the mask code is correct `255 - e` rather than `255 if e == 0 else 0`
e. Whether #601 addresses some of the stuff that I think still needs fixing as described above.

Perhaps we should combine the best of both.

Issue #599, opened 2021-10-08
Raster wizard introduces black lines at right hand / bottom edge


Sophist-UK:

Sophist-UK:
I tried it on my laptop running from source and I didn't get this - so probably fixed by #597.

But bad image processing is for me getting to the point where 0.7.1 is unusable.

tatarize:
The PR #600 and #601 both seek to address this problem. I think there are some merit to both. But, some clear interference. #601 seeks to make rasters uniform. All the raster stuff is transitioned to using dewhite to make the preview images. So all images loaded and viewed in meerk40t are converted to black/transparent images. Actualize makes the images actually black and transparent. It should also reduce bounding box if needed and allow some other nice changes. But, the most important thing is that it should get rid of any and all whack-a-mole style errors here.

* Has not been extensively checked with raster wizard.
* May avoid applying to an image from execute-job if it matches the old criteria but should actually trim and modify the image.
* Has not been tested for inverted images and inverted images might prefer being white/transparent images
* Does harder alterations of the current errors.

#600 
* fixes the actualize pixel code call, but may be overly harsh in recalculating the bounding box.
* includes a burn value in actualize that is not triggered.
* Burn should not function differently than not-burn. 
* Mode P for pallette might have an alpha. I forget exactly how that works. This would mean new_mode.endswith("A") could fail.
* Comment is wrong, Pillow does allow to fill in new areas created by transform with transparent but only when that mode is RGBA and is the default in that case. The problem seems to be that it's transparent black so converting it to L converts that color to black.
* Probably slower since it holds the alpha and image and applies the same to both, likely bbox is also slower.

Sophist-UK:
As stated above, #597 seems to fix this problem. However responding to these comments:

1. Both #600 and #601 seek to make rasters uniform. I just wasn't brave enough in #600 to go to greyscale "L" on actualise rather than keep the image coloured "RGB". But on balance I think that going to grey scale is a good idea on actualize. But I think my approach to keeping alpha i.e. end up with "LA" is a better UI approach for user Actualization of source elements - whereas for simulation we should ideally switch to black with alpha for greyscale because that visibly simulates burns, and for burns we should switch to greyscale with no alpha because that is simplest for processing.

> Burn should not function differently than not-burn.

If we go to greyscale uniformly, then we don't need to treat burn as different.

[...]

PR #601, opened 2021-10-08
Make Processed Images Uniform

tatarize:
Seeks to make rasters uniform.  Should get rid of any and all whack-a-mole style errors here. Should simplify the understanding. Conceptually this takes alpha into account and uses it to generate a greyscale `L` image. This should be the result of any actualize operation. The core idea behind this technology is displaying the black and white as black and transparent. This is done by using the black and white image as the alpha-mask of a pure black image and displaying that. Then under the hood, all images that with any processing are greyscale. If the image is not `L` to start, we peel off the alpha and combine that with the image to make alpha values entirely white. 

* All the raster stuff is transitioned to using dewhite to make the preview images. This includes images that are full color and are not changed. They are merely displayed in black and transparent. 
* All images are actualized and (should-be) Raster Wizard'ed to black/transparent images. The viewed bitmap is automatically converted regardless of the original mode of the image.
* Black/Transparent images are stored in `mode L` image, which is properly black and white but viewed as black/transparent.
* Actualize makes the images actually black and transparent, and crops the size correctly.
* Has been extensively checked with RasterWizard.
* Has been tested for inverted images and inverted images might prefer being white/transparent images.

### DIFF
--- a/test/test_imagetools_wizard.py
+++ b/test/test_imagetools_wizard.py
@@ -82,32 +82,36 @@ class TestRasterWizard(unittest.TestCase):
         try:
             kernel_root = kernel.get_context("/")
             # kernel_root("channel print console\n")
-            for mode in ("RGBA", "RGB", "L", "1", "P", "F"):
-                svg_image = SVGImage()
-                svg_image.image = Image.new("RGBA", (256, 256), "white")
-                draw = ImageDraw.Draw(svg_image.image)
-                draw.ellipse((50, 50, 150, 150), "black")
-                draw.ellipse((75, 75, 125, 125), "blue")
-                draw.ellipse((95, 95, 105, 105), "green")
-                svg_image.image = svg_image.image.convert(mode)
-                node = kernel_root.elements.add_elem(svg_image)
-                node.emphasized = True
-            kernel_root("image wizard Gravy\n")
-            for element in kernel_root.elements.elems():
-                if isinstance(element, SVGImage):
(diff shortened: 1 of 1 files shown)


Sophist-UK:
> I'm biased but I think we should consider \[this one over #600\].

I am not biased and I have no issues considering #601 vs #600.

I will need to try this code against my test cases, but I wonder whether:

a. We should reduce coloured images to greyscale every time - I have no idea either way but I shied away from going this far with my code but if you think it is OK, then it is fine with me;
b. If we are converting to "L", whether the code should mimic the greyscale code in RasterScripts;
c. Whether we should aim for an LA mode which preserves alpha rather than convert transparent to white (which I would personally like to see because you don't end up with white background overlaying other elements);
d. Whether the mask code is correct `255 - e` rather than `255 if e == 0 else 0`
e. Whether #601 addresses some of the stuff that I think still needs fixing as described above.

Perhaps we should combine the best of both.

tatarize:
Discussion on #599

tatarize:
Responded in the issue #599 (trying to keep things organized)

Sophist-UK:
Looking good except some minor stuff:
1. After right-click actualise, scene is not updated with cropped image size.
2. In raster wizard Gravy, with `grayscale` **disabled** and `tone` enabled there are some light gray artifacts showing in the Heather Stone image.

Otherwise looking very good!!!!

Once again you have proven your coding awesomeness!!

PR #602, opened 2021-10-10
refresh scene after actualize...

Sophist-UK:
... so as to e.g. show new bbox when image has been cropped.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -4987,9 +4987,6 @@ class Elemental(Modifier):
         @self.tree_operation(_("Actualize pixels"), node_type="elem", help="")
         def image_actualize_pixels(node, **kwgs):
             self.context("image resample\n")
-            node.object.node.altered()
-            self.context.signal("element_property_reload", node.object)
-            self.context.signal("refresh_scene")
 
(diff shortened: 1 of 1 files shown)


Sophist-UK:
Code missed from #601

Sophist-UK:
RasterWizard has similar problem where bbox is not refreshed after you click the "Process Image and Export button"- but no time to look at this tonight or tomorrow.

tatarize:
Might need to put the .modified() somewhere deeper so that it triggers on any actualization. It shouldn't matter in the edge cases where it's actualized while in the planner. Or change the functionality since I believe both end up in wxPython `ShadowTree` which does the gui-listening and updating of the actual wxTree The differences between altered and modified is that altered does a label_update(node, force=True) and modified calls `self.context.signal("modified", node)` whereas altered calls nothing. 

Since the meaning of `altered` is that the core underlying data has changed in fundamental ways. It *should* clearly update the bounding box and does not.

tatarize:
```
            if hasattr(element, "node"):
                element.node.altered()
```
Is found in resample in `imagetools.py` and so really the problem here is that `altered` isn't triggering the update that `modified` does.

Sophist-UK:
I reverted my original changes and it seems to work great for the cases that I reported as failing. So this should be good to go.

PR #603, opened 2021-10-10
Require Actualization

tatarize:
## Adds Constraints
* Requires all images be actualized to become cutcode
* Actualization requires overt step_level parameter.
* Actualization must occur before path defining the created for inner first calculations.
* Actualized size is used for path defined for creating inner first calculations.

## Classes Modified
* RasterCut objects use PIL.Image objects rather than SVGImage objects
* RasterCut requires explicit `tx` and `ty` values.
* `as_cutobjects` passes crosshatch value rather than uses settings.

## Functions Moved
* Actualization code moved into `actualize.py` rather than imagetools.
* Actualization code removed from `imagetools`.

## Code Simplified
* Removes duplicate code for crosshatch version of settings
* Remove alternate image filters for rasterplotter.

## Testing Added
* Include Robust Testing of actualization changes

### DIFF
--- a/test/test_core_cutcode.py
+++ b/test/test_core_cutcode.py
@@ -109,6 +109,7 @@ class TestCutcode(unittest.TestCase):
         self.assertTrue(isinstance(image, Image.Image))
         self.assertIn(image.mode, ("L", "1"))
         self.assertEqual(image.size, (3, 3))
+        self.assertEqual(rastercut.path, "M 100,100 L 100,106 L 106,106 L 106,100 Z")
 
     def test_cutcode_raster_crosshatch(self):
         """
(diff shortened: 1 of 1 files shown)


Sophist-UK:
Good concept - have not reviewed the code.

tatarize:
Yeah, still needs some stuff moved around. RasterCut, for example, takes a svgimage rather than pil.image which would be clearly superior there. So if we know what we're getting more stuff can simplify. I'll open it for review when done. It has the disadvantages of no being mission critical and being changes to semi-critical code.

tatarize:
This should be good now.
Bit of a big set of changes though, though heavily checked with tests.

PR #605, opened 2021-10-15
Create assets for pull_requests / pushes

Sophist-UK:
It would be great to build executables for pull-requests so that we can test PRs as executables rather than just running from source.

I am building here on the excellent work done by tiger in the existing ci.yml.

This is an interim script which has releases commented out but the intent will be to replace the existing ci.yml once it is fully working.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -239,4 +239,3 @@ jobs:
         path: |
           dist/MeerK40t-${{ github.event.after }}.exe
           dist/wx-MeerK40t-${{ github.event.after }}.exe
-


tatarize:
Pretty sure this would work. Not sure this is needed. Though I guess it might be somewhat useful to make a PR and have a version. Not sure where the files would end up.

Sophist-UK:
See the most recent build jobs for this PR - there is an artifact which is a zip file of executables for all environments.

tiger12506:
What I don't appreciate about this is that anyone can make a pull request. Including ones that execute arbitrary code through my self-hosted centos runner, and it will happily build upon trigger of pull request without a second thought.

I'd prefer that you set these to manually trigger, and Tat can run them as he needs them. It also wastes a lot less build time, creates a lot less random artifacts, etc.

tatarize:
I don't actually think I'd ever trigger this. I have thus far never needed a compile to review a change. I've no qualms with publishing a release based on a particular branch if that was a needed thing. I am fairly sure the code revisions can be done in source. And while inpain does often compile his builds to test things that's a bit above and beyond on his part. And isn't really intended to check the code but the compile methods and if they lost something somewhere.

Sophists pulls are from branches within the project that means he has write access so assuming it's just any code from within this base that has the authorized runner it would at least just imply that anybody who has access to the code base has control over the runner.

I'm also a little concerned on the time limit things, while it was clear we weren't going to hit them very easily with a release every couple weeks we *might* be able to hit them when just about every commit triggers a full build.

Sophist-UK:
> Checking the artifacts, and downloading half a gig, I don't think I'd ever download them if I _did_ trigger this. And I'm not sure how limited the resources are but generating like half a gig data for each commit and how ever many minutes of compile time doesn't seem very frugal. 

1. The Zip file contains all builds, but it is easy enough to split them into separate zip files
2. For testing, the zip file currently contains both a build made under ubuntu and one made under Centos.
3. The Centos build is currently NOT made using UPX compression as this is missing from the Centos build - and that makes the Centos build 50% bigger than the Ubuntu one.

> While I usually just leave edge conditions to the people governing them, and I certainly will leave the final vote here to Tiger, I don't think this is either needed or useful. 

You didn't think that automated builds were worthwhile initially either. The great thing about these are that they are automated, they cost us nothing and they are there if you need them. (The Picard automated builds are not used very often either, but the odd occasions when they are needed makes them worthwhile anyway.)

>I think we can trigger branch based releases that already "triggers" this. 

Not sure what you mean by this - but the tigger script currently only runs on a release. And you don't want to be releasing (for general use) new versions immediately after a new commit which have not had some real-life testing - as we have found recently this can result in bugs which are significant enough to require an immediate urgent re-release, and IMO we should try to avoid these happening.

> While I'm sure it's not too costly for github or whatever, I wouldn't want to bring any eye of Sauron from Microsoft on this nifty feature as a cost cutting measure, or use more resources than are needed. 

This is not going to do that. It is actually **normal** practice.

Sophist-UK:
> I'm also a little concerned on the time limit things, while it was clear we weren't going to hit them very easily with a release every couple weeks we _might_ be able to hit them when just about every commit triggers a full build.

Our usage is never going to even come slightly close to reaching the limits.

1. A workflow can run for up to "only" 72 hours - ours run in c. 6 minutes.
2. A single job (multiple steps) can run for up to "only" 6 hours and ours (could) run in 4 mins.
3. You can "only" queue 500 workflows every 10 seconds - so we would need to get to (say) 100 pushes to PRs every 10 seconds or 36,000 pushes an hour to reach this.
4. You can "only" run 20 concurrent jobs inc. 5 Macos jobs - the rest are queued. Assuming that jobs take (say) 10mins each, and each commit creates 10 jobs, that means would have to continuously push > 12 commits an hour to have an ever increasing queue.
5. The workflows I create have caching to reduce footprint. They also cancel any currently running jobs for the same PR if you make two commits in quick succession. In other words, I am already taking care to control the usage despite the huge headroom we have.

Sophist-UK:
@tiger: Centos build is still 170MB - no error messages about UPX but executable is no smaller.

tiger12506:
We cannot run these build operations on a legacy branch. The structure of the project was sufficiently different then, and I'm pretty sure it will not work. If you wish to make it work, then I recommend having a separate legacy yaml file (in a separate pull request) with the build steps for legacy there.

Cosmetically, I'd prefer short hashes in the filenames, but no deal-breaker.

Is there a possibility we should be putting the separate platforms in their own build yaml for the sake of organization?

Otherwise, I think it is acceptable to continue.

In future, please consider changing how the pull requests are made. This created a parallel build script which changed many things all at once, rather than incremental changes to the existing build script. If you had developed in your own fork of meerk40t, you could have modified ci.yml, with commits made along the way, and following your changes would have been less time consuming for me. Specifically, I would have preferred the changes to the build steps as separate commits from the pull request trigger changes, etc.

tatarize:
Yeah, legacy branch probably should just be ignored, it's basically finished as a branch thing. I don't know any more needed changes and since 0.7.x works it's better to fix things there if there are any issues.

Sophist-UK:
The problem with creating releases will come when we have release checking available.

There will be no way of distinguishing between a true beta (which we believe to be stable) and an executable for test purposes which may be quite useless.

There are no limits to resource usage for public repos - the 2,000 minutes per month is for private repos.

Also AFAIK, you cannot specify a PR branch to trigger this manually.

tiger12506:
And we have given you at least two ways to still have what you want as well.

1) Host your own fork where you generate on every push/pull request as you see fit. (They are one-click update on github)

2) Make the trigger manual, and if you want a build to test on your production machine... click the button.

tatarize:
> I can't speak for Tat, but I assume he will allow it.

I often just defer to your expertise on questions like this. I could see some utility to weekly builds of main branch or something, but it's a bit non-obvious where those builds would go, and we've been releasing things more frequently with the autobuilds as is. Things fine as they are, or if they aren't we could give it longer to see what our other needs might be.

PR #608, opened 2021-10-15
Add in Modification Toolbar. Flip Horizontal Flip Vertical

tatarize:
Adds in a Modification Toolbar. Can be turned on provides flip horizontal and flip vertical buttons.


tatarize:
Just how toolbars work in AUI. -- Added right click flip and mirror to the normal menus as well.

Sophist-UK:
> The toolbar menus work like that. They only add to toolbar parts.

So toolbars are different from panes? This is probably a naive question, but can we not do it all as panes so that they act the same? Otherwise this is not intuitive to the user.

tatarize:
The toolbars differ from panes but they actually do interesting things that panes don't. You can put them at the side going down etc. They are generally quite nice. Yeah, went ahead and gave them a flip submenu.

PR #613, opened 2021-10-17
Improve Optimization

tatarize:
* Truncates search closest distance is already zero.
* Initially sets first attempt point to next cutobject in the original path (which is usually coincident).

Testing the code seemed to have reduced the speed of a test file from 2.6 to 2.4 seconds. May need additional profiling to figure out what is taking the bulk of the time and reduce that.

### DIFF
--- a/meerk40t/core/planner.py
+++ b/meerk40t/core/planner.py
@@ -1142,28 +1142,46 @@ def short_travel_cutcode(context: CutCode):
     while True:
         closest = None
         backwards = False
-        closest_length = distance = float("inf")
-        for cut in context.candidate():
-            s = cut.start()
-            s = complex(s[0], s[1])
-            d = abs(s - curr)
-            l = cut.length()
-            if d < distance or (d == distance and (backwards or l < closest_length)):
-                distance = d
-                backwards = False
-                closest = cut
-                closest_length = l
-            if cut.reversible():
-                e = cut.end()
-                e = complex(e[0], e[1])
-                d = abs(e - curr)
-                if d < distance or (d == distance and backwards and l < closest_length):
-                    distance = d
+        closest_length = float("inf")
+        if ordered:
+            last_segment = ordered[-1]
+            if last_segment.normal:
+                # Attempt to initialize value to next segment in subpath
+                cut = last_segment.next
+                if cut and cut.permitted:
+                    closest = cut
+                    closest_length = abs(closest.start() - curr)
+                    backwards = False
+            else:
+                # Attempt to initialize value to previous segment in subpath
+                cut = last_segment.previous
+                if cut and cut.permitted:
+                    closest = cut
+                    closest_length = abs(closest.end() - curr)
                     backwards = True
+        distance = closest_length
+        if distance > 1e-5:
+            for cut in context.candidate():
+                s = cut.start()
+                s = complex(s[0], s[1])
+                d = abs(s - curr)
+                l = cut.length()
+                if d < distance or (d == distance and (backwards or l < closest_length)):
+                    distance = d
+                    backwards = False
                     closest = cut
                     closest_length = l
-            if distance <= 1e-5:
-                break  # Distance is zero, we cannot improve.
+                if cut.reversible():
+                    e = cut.end()
+                    e = complex(e[0], e[1])
+                    d = abs(e - curr)
+                    if d < distance or (d == distance and backwards and l < closest_length):
+                        distance = d
+                        backwards = True
+                        closest = cut
+                        closest_length = l
+                if distance <= 1e-5:
+                    break  # Distance is zero, we cannot improve.
         if closest is None:
             break
         closest.permitted = False


Sophist-UK:
P.S. We would need to test the timing, but it might be quicker to find the shortest travel by creating a full list of candidates with entries for both directions (in both directions) and using `list.sort` to calculate the distances, like this:
```
cuts = []
for cut in context.candidate():
    cuts.append((abs(cut.start() - curr), False, cut))
    if cut.reversible():
        cuts.append((abs(cut.end() - curr), True, cut))
cuts.sort()
cut = candidates[0][2]
```
The sort is a single bytecode instruction, so it is effectively done by C-code rather than python interpreted code.

Sophist-UK:
P.P.S 

1. If python `sort` is faster (running single threaded) then it should be simple enough to extend this concept to do the sort in numpy (which may be able to take advantage of (e.g. 8) multiple cores) or even in a GPU equivalent of numpy (if the user has a suitable GPU - taking advantage of e.g. several hundred cores). 

2. The speed improvement from your new algorithm is going to be greatest with SVGs which are not already optimised - i.e. where subpaths have a large number of pathsegments. 

I have several examples of these - some of mine (which I had to optimise by hand in inkscape to make them optimise in MK in a reasonable time) and some from a user who was complaining of unacceptably lengthy optimise times.

We are out for the day tomorrow (we have a friend staying and are going on a day long boat trip on the local nature reserve ocean lagoon) but I should be able to measure performance soonish. (If you want my test files to time them yourself, I am happy to send them to you. Indeed, I may just do that before I turn in anyway.)

tatarize:
> Otherwise this is pretty much the same algorithm I had in mind. smiley
> I will test it tomorrow with a file I have that has lots and lots of very short segments and takes over a minute to optimise.

I know, because as I said, the suggestion you had for your 1 and 2 is inelegant. If we abort part way through and offer up the next correct value first we get a general algorithm that combines them.

> P.S. We would need to test the timing, but it might be quicker to find the shortest travel by creating a full list of candidates with entries for both directions (in both directions) and using list.sort to calculate the distances, like this:

Testing the timing was a kick in the teeth. My test file which is actually pretty reasonable weighed in my favor went from 2.6 to 2.4 seconds cutting the timing needed for the brute force greedy algorithm to merely a fraction of what it was before. The typical case is find the nearest point and when that point is on a new path it will basically just walk that contiguous path without any tests needing to be run. But, that means *most* of the time isn't in the optimization, and I honestly don't know what this thing is doing. Pycharm has some profiler plugins but most of the profile things are for Professional and I run Community though there might be some others. But, the fact remains I do not know what is sapping my time. My tweak should have speed things up by like 80%-90% but it sped it up by 8% so like dark matter in the universe I don't know what's up with 90% of this. It seems like even if I went and wrote a fancy acceleration structure to solve this faster or even switched to an optimized two-op thing or something I would just be hyper-optimizing the remaining like 2% here.

> PPS.

Even my cursory timing tests show that yes, I bypassed most of the searches it's doing far fewer of these. But, it's also exposed that weirdly little of this is anything to do with this sped up section since it saved literally 0.2 seconds or so. I thought maybe the quick thing where I wrote the next-previous matters but that came down at around 0 seconds. So I think further optimizations here might be somewhat pointless unless they are to whatever section of the code is *actually* responsible.

Sophist-UK:
I decided I had to know:

Bat body - main 77s, your new code 9s (yes nine).
Bat wings - main 51s, your new code 3s
47 edit - your new code 127s (I didn't bother trying on current main but I assume it will be c. 18mins)
Mario - got to 200s on your new code and killed it.

So your new code is c. an order of magnitude faster on badly optimised files.

But I also thought it might be useful for a quick before and after test on the different types of paths in the bat body file. So I tested an optimise on only the engraves (many short subpaths) and the cuts (two closed subpaths) for both the file I sent you and my manually optimised version of the cuts.

Engraves (not manually optimised)
Original file	(218 subpaths, 4689 path segments): main 61s, your new code 5s
Optimised file	(218 subpaths, 4689 path segments): main 74s, your new code 6s

As expected both files were the same. But new code was an order of magnitude faster.

Cuts (manually optimised)
Original file	(2 subpaths, 1160 segments): main 5s, your new code maybe 0.5s
Optimised file	(2 subpaths, 102 segments): main maybe 0.5s, your new code maybe 0.25s

So still noticeably faster for both original and manually optimised (and much shorter) paths.

I also did a quick test of using list sort and it was slower for the Sander files and not noticeably different for the Bat Body files. I tried a couple of optimisations and they didn’t help, so let’s forget that idea and take the win we have.

Sophist-UK:
> I know, because as I said, the suggestion you had for your 1 and 2 is inelegant. If we abort part way through and offer up the next correct value first we get a general algorithm that combines them.

It's easy for me to say after the event, but your implementation was pretty much what I had in mind and was attempting to describe. We follow a path if we are on it and segments are joined, otherwise we find the shortest travel.

PR #617, opened 2021-10-19
Implements 2-opt in numpy

tatarize:
Calls and quickly returns `short_travel_cutcode_2opt` from within short_travel_cutcode rather than using the regular code present.

This firstly implements this code to do a 2-opt algorithm and secondly, uses numpy to do it.

* Due to the skipping of the candidate code it does not perform inner first optimizations.
* Due to the numpy requirement it doesn't really work per usual without numpy.

This code should live and die as a draft.  The job here is mostly to show the speed and utility of two-opt implemented in this fashion.

--

Uses code I wrote for vpype: https://github.com/abey79/vpype/commit/7b1fad6bd0fcfc267473fdb8ba2166821c80d9cd this isn't an issue since I wrote that code too. Also, vpype is also MIT licensed.

### DIFF
--- a/meerk40t/core/planner.py
+++ b/meerk40t/core/planner.py
@@ -1193,33 +1193,41 @@ def short_travel_cutcode(context: CutCode, channel=None):
     if channel:
         channel("Executing Greedy Short-Travel optimization")
         channel("Length at start: %f" % context.length_travel(True))
+
     for c in context.flat():
         c.permitted = True
+
     ordered = CutCode()
     while True:
         closest = None
         backwards = False
-        closest_length = float("inf")
+        distance = float("inf")
+
         try:
             last_segment = ordered[-1]
+        except IndexError:
+            pass
+        else:
             if last_segment.normal:
                 # Attempt to initialize value to next segment in subpath
                 cut = last_segment.next
                 if cut and cut.permitted:
                     closest = cut
-                    closest_length = abs(complex(closest.start()) - curr)
+                    distance = abs(complex(closest.start()) - curr)
                     backwards = False
             else:
                 # Attempt to initialize value to previous segment in subpath
                 cut = last_segment.previous
                 if cut and cut.permitted:
                     closest = cut
-                    closest_length = abs(complex(closest.end()) - curr)
+                    distance = abs(complex(closest.end()) - curr)
                     backwards = True
-        except IndexError:
-            pass
-        if closest_length > 1e-5:
-            distance = closest_length
+
+        if distance > 1e-5:
+            if closest:
+                closest_length = closest.length()
+            else:
+                closest_length = float("inf")
             for cut in context.candidate():
                 s = cut.start()
                 if (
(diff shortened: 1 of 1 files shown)


Sophist-UK:
I tried creating a candidate list and sort it to see if it would be faster than the current short_travel_cutcode and it wasn't any faster than the current main. I tried it both creating the candidate list at every need to travel and creating it at the start of the method and removing candidates as they cut. But neither was faster. In both cases it needed to calculate the distance from current point every time it had to find a travel point and this wasn't twice as fast as part of a python sort as the current loop iteration (average half of the candidates) and in the latter case it needed to find and delete same-subpath candidates which was extra overhead. 
 
If I read your code correctly you are implementing 2-opt without greedy (presumably to try it out and to understand the performance implications). It might be worth trying it with and without greedy first to see which is faster. If you prefer I would be happy to try numpy for greedy.

Hopefully we will find that greedy using numpy is faster because it uses all cores when calculating the distances. I am not sure whether numpy can parallise the actual sort, **but** since you only want to find the minimum distance for greedy there is a partial sort (to find the lowest x i.e. 1 value) or perhaps there is a numpy function to find the minimum.

tatarize:
2-opt here should be faster and better optimize the results. Yeah, I'm not doing anything fancy except forcing it in there as a weird step. I might be able to run it parallel to normal optimize or something to have it an option. But, it does a pretty nice thing and prints the times etc.

Sophist-UK:
If 2-opt is faster in most non-trivial situations, then I guess that should be preferred over greedy.

If I remember correctly, I think you said that numpy is pre-requisite with Pillow but optional when you run MK without a gui. Is that correct? If so, then I guess we can test for numpy import at start-up and set a NUMPY_AVAILABLE global variable which we can then use to choose which algorithm to run.

I have never used numpy myself - and don't really know either its capabilities nor which bits are capable of using multiple cores, so I am not really able to understand or review your 2-opt code in detail without doing some research. But if you would like me to do that, I would be happy to put the time in and see if I can suggest potential speed optimisations.

Sophist-UK:
Suppose we did a greedy optimisation, then treated all burns and shortest travels (shortest 80% or shorter than xxxx mil) as fixed in order to try to shorten the longer travels.

Would it help to pre-calculate (using numpy) the distance between all pairs of endpoints of these remaining travels in order to be able to give numpy the data it needs to find the optimum solution?

tiger12506:
Essentially numpy helps you do matrix math. Not all problems can be solved only by affine transformations (ax+c), but those that can be (which can be expressed by vector math) can be made very very fast using SIMD instructions and if it so supports it, direct GPU calculation. (3D graphics is done entirely that way in matrix math). If your problem isn't representable in matrix form, then numpy is useless to you.
The devil is always in the details. If exactly this type of problem were easy to solve, everything would be a multi-core program -- and it isn't.
Good luck, Tat, you're well beyond me. I never liked matrix math much, preferring explicit systems of equations and trig functions over matrices of coefficients and rotation matrices.

tatarize:
I switched this up to checking if numpy is available. In theory this could be checked for correctness if opt_2opt was set to True and then we could default `opt_2opt` to True.

This would potentially allow rather marked improvements for a lot of cases. Since without inner first checked the 2opt algorithm actually works fine.

tatarize:
Hyper-optimized some more stuff there.
Renamed greedy, nearest neighbor. There's other greedy algorithms with stuff.

Since 2-opt suffers other defects with non-reversible paths it is set to false.

tatarize:
I think this might be merge ready. It depends on a couple factors the default 2-opt is off and it fails with a couple methods but since it's defaulted to off that shouldn't really matter. If the defaults for nearest neighbor are good. We're good.

PR #632, opened 2021-10-26
Kernel modifiers

tatarize:
The goal of kernel modifiers is to change the kernel modifier objects to apply not to a single context but to every context. While that context.root settings are often useful there's a need to have general access to certain functionality with certain understandings. These functions should be established before booting the kernel and they should give access to .<modifier> for all contexts of the kernel.

These should be able to be swapped, where `.device` gives the actual device currently loaded, and if there is a swap of this device it should be possible to have anything that watches/listens to the modifiers be notified of the change in case they need to update in some functional way. For example the bed-size should be part of .device and needs to be able to watched for a change here and swap the size of the bed when the .device changes.

This also needs to swap kernel commands, so the kernel commands are actually stored within the modifier and when the active modifier switches the permitted commands also switch. And it needs to swap the rest of the registered values in the kernel, so if we define some panes that are valid for our driver these should be attached to just that `.device`. Then when the device is switched we can go ahead and load a different perspective (aui gui settings and arrangement) altogether and switch the entire setup for meerk40t.

Current under modifiers:
* Bind/Alias: determines the keybinds and aliases for the command line. Very few things would need to know what these are but it would be nice if everything simply could check .bind and access the keybinds regardless of the window or the settings for that window.
* Elements: this is currently well implemented and fundamentally important and strengthening this would be extremely helpful. .element controls the tree currently as well as defining most of the tree commands. If we permit different modifiers from having their own registered space that swaps in the kernel when the modifiers swap it would be easy to give different tree responses depending on the device currently selected.
* Planner: The planner wouldn't need to be swapped but it would be helpful if everything had access to this since the cutcode planner is important and accessing and modifying it matters. This is mostly done with commands currently but there's clearly a need to look into and access the planning data such as spoolers and laserpane.
* Drivers: this is currently rather weakly done and poorly implemented kernel modifiers would be implemented to make these much more coherent.
* Inputs: This is currently weakly implemented and is part of drivers (more or less)
* Output: This is currently weakly implemented and is part of drivers(more or less)
* Spoolers: This is currently well implemented but might be better as part of drivers or parallel to drivers.

Additionally:
* Camera, this likely does not need to be a modifier, it could rather be a modifier aspect for what camera hub points to most of the time.
* CameraHub: this should be properly camera. The utilities here is that camera and camera-hub serve to switch the camera in the camera hub to an active current camera.

There are several places where there are multiple plans or multiple spooler or multiple cameras and we have one things that weakly points at the other things. These more or less call for the altered kernel modifier functionality.

`.device` could, for example call for a device even without a driver set. This would provide device specific functionality like home positions or bed-size or perspective but not require anything that reads the `.device.spooler` information.

### Devices

Devices would allow for a driver switch that would change the tree commands, panes, windows, commands, and anything else dynamically created by the registered values, for example rasterwizard scripts (likely pointless) and a lot of options that can now be dynamically created. Switching a driver would switch the bed size and arrangement of the gui and panes and windows you have access to.

This serves to not only fix some of the issues with how to deal with multiple drivers, but how to deal with multiple drivers with regards to the UI. If somebody wanted to code their own driver, this would allow that to be done much more simply than the current system would ever permit. You would need to define a driver then use devices to initialize them based on the driver-type. And defer some of the functionality there. You would have multiple devices that could swap and these could often have the same driver.

### Rotary

Rotaries could be used in multiple devices or you could have the same device with and without a rotary. The idea here is that you could have a default rotary which consists of just .rotary code that gives you no-rotary information.

Default of the modifier is the unswitched version that isn't swapped out.


tatarize:
This will be revisited with another attempt that is less destructive. There's too much code base to be that violent and expect it to continue working. Rather than changing modifiers, I will add in services which do what the new modifiers were meant to do. So things that genuinely need to switch functionality can do that, offering a service such as .elements at every context or .device or .gui where these services within the kernel are worth accessing and modifying the `registered` dict accordingly.

PR #645, opened 2021-10-29
Further optimize tweaks

Sophist-UK:
Fix a buggette (potentially burning a segment an extra time). 

Stay on current path rather than travel away, if path is not closed i.e. a small gap <= 1/20".


tatarize (approved):
Solid enough changes

PR #646, opened 2021-10-29
Improve Execute Optimisation tooltips

Sophist-UK:


tatarize (changes requested):
Can this also change  https://github.com/meerk40t/meerk40t/blob/643a7fe2955b120030fc88c30670dd8d5162522d/meerk40t/core/planner.py#L138 lines. This section of the tooltips is going to go away in favor of choices. Basically that panel in ExecuteJob is going to be replaced with a choices value set from `choices/planner`.

In laser-panel second page you will see optimize tips there. Those are automatically generated, and these will also be automatically generated here even if I don't include the distance thing in that section. Copy and pasting would suffice for now, though in theory if Execute's Optimization options are in a panel that entire panel could be replaced with a PropertiesPanel given the 'planner' choices data.

But, approving this now would give me copying and pasting work later.

Sophist-UK:
Yes - I can retrofit it (if that is the right phrase for fitting it to something newer rather than older). Will need to wait until tomorrow or Monday or something.

tatarize (approved):
I changed it to fix those sections.

tatarize:
I converted the panels over. And I can fix things better in a PropertiesPanel. ExecuteJobs also has the Automatic menu which was replaced with choices instead of being hard coded in the menu there. Since I know the setting and all the data about it, I can alter the form pretty easily. It also registers the properties in the kernel.

PR #647, opened 2021-10-29
Plugin Extension Locale

tatarize:
Adds in locale commands to build check and validate the locale information specifically for meerk40t.

`locale change hu update` for example uses `locale`, `change hu`, and `update` which uses the current template file in `locale/messages.po` to update the structure and take the work from the given translation file. It won't work with English but we changed to Hungarian so it reads the Hungarian and writes meerk40t.update to the root directory (It's not good enough yet to risk writing even when we can just pull from git).

There is also a `generate` command this currently reads all the .py files in the meerk40t directory and finds any place they contain `_(".*")` or `_('.*')`. I do not believe this supports multiple lines currently, and prints the found string and the translation within the current set language. This isn't very useful but it's a building block for more useful other things to come.


tatarize:
Apparently the method to parse these things is to use the builtin for the python parsing. https://docs.python.org/3/library/tokenize.html which can correctly parse all the python you throw at it. This could then be compared against a parser for .po files.

I think the level criterias should be as follows:
* Any non-tool-tip within a wx.Python panel code that is not part of a kernel.command() should be top priority. 
* Any tool-tip within a wx.Python panel.
* Anything translated within a kernel.command()
* Anything not found in the first four criteria.

Then generate a custom messages.po file from that information sorted by priority. Valid translations should provide all priority 1 and priority 2 messages, anything below that is just gravy.

---

1. Calculate the latest correct messages.po file.
2. Check the validity of the current translation for a particular language. What percent of high priority messages are translated.
3. Provide a gui to translate messages. wx.tree with the listed priority and source file
4. Provide a method of sending translation patches to developers. 
5. Provide a method of importing translation patches.
6. Provide a method for compiling translation patches to .mo files.
7. Alter the current methodology to allow the application to swap meerk40t.gui out without shutdown

6 and 7 would just be nice to have. 4 and 5 would be really helpful. 3 would possibly be a tall order but it doesn't really seem that hard. There maybe oddities with various regions but I'm not doing very much with them, and people could use their own po editor stuff.

4 and 5 might be better done simply as sending the entire replacement .po and just calculating a diff file for that locally. What translations were added or changed. 6 is actually done here: https://github.com/python-babel/babel/blob/cc36c84a83dd447bf48a6af3eb03c97bf299e8cb/babel/messages/mofile.py#L106 (licensing seems to require copyright inclusion which isn't too hard.)

---

* Calculate the messages.po file.
* Find all the entries in the messages.po file that are not in the localized meerk40t.po file. List these based on priority.
* Find all the entries in the messages.po file that are in the localized meerk40t.po file and permit transferring the current comments and entries from the messages.po file into the meerk40t.po file using messages.po for the structure.
* Find all entries which are in the localized meerk40t.po and not found in the messages.po, mark these as orphans and append them to the updated meerk40t.po file at the end.
* In the translation gui provide these orphaned translations as suggestions if the levenshtein distance below about 10 or so.
* In the translation gui do not permit strings which fail to match the required op order. Tokens of `\n`, `%s`, `%f`, `%d`, {.*} must be preserved within the strings.

tatarize:
This is being spun off into https://github.com/meerk40t/poboy and treated as a standalone application and plugin. It's not exactly clear yet, but the functionality here should be largely borrowed from babel ( https://github.com/python-babel/babel/tree/master/babel ). Given some gui interfacing and corrections.

PR #649, opened 2021-10-30
Make interrupt msg translatable.

Sophist-UK:


tatarize (changes requested):
I think it might not be allowed. Since this is the text based one which implies no gui.

Sophist-UK:
Ah - you may be right. `_` is a synonym for a wx function which uses Python's i18n stuff under the covers.

But I have never understood why wx (and Qt) provide translator functions when Python already has its own. I am sure there must be a reason, but I just don't see it.

PR #652, opened 2021-10-30
Settings -> Preferences for consistence with OSX standards.

tatarize:
* Version Bump to 0.7.4-beta 2
* rename some plan commands to use data.name rather than default_plan
* Adds command plan.sublist to sublist the current planned data.
* Changes Device Preferences to be consistently Configuration.
* Changes Settings to be consistently Preferences.
* Add in signal listener for simulation to watch the current plan.
* Add in commands to invoke plan.sublist within simulation to remove parts of the cutcode.
* Adds in choices for units in scene.
* Removes other declarations of the unit values within scenewidgets. 
* Converts MEERK40T_VERSION to APPLICATION_VERSION.
* Adds APPLICATION_NAME for the name "MeerK40t". 
* Moves the sys.excepthook to the start of wx.App() since it's required there. Anything that crashed after the wxmeerk40t was imported and the app was loaded would not be able to crash gracefully.
* `send_file_to_developers` was changed to read data and `send_data_to_developers`. The crash no longer requires that the data be saved, it can merely be sent without saving the log.
* Added in a `kernel.open_safe()` function which should permit opening of the working directory or changing the working directory a user one depending on the type of OS.
* Used `open_safe()` to guard against -v debug log, channel save logs, and crash logs.
* Added get_safe_path() which returns the expected safe path. For windows this is '%LOCALAPPDATA%'\MeerK40t for Mac this is ~/Library/Application\ Support/MeerK40t and Linux this is ~/.config/MeerK40t
* Fixed series of crash bugs.


joerlane (approved):
Reviewed source and tested as needed.

PR #659, opened 2021-10-31
Fix #657 StepRepeat incorrectly repeats pre/post special ops

### LINKED ISSUE #657: [Bug?] Execute/StepRepeat includes Homes, Beeps etc. between
IMO (and I may be wrong) automatic before/after special operations (like home, beep etc.) should apply to the job as a whole and not to each step in a `Step Repeat`.

Sophist-UK:
This PR fixes #657 (where pre/post special ops were incorrectly repeated).

It also improves the UI - terminology ("delta" vs "gap"), removes default repeat dimensions (5 is more likely wrong), displays default delta in user friendly measurement units, correctly determines the delta sizes using operation elements not what happens to be selected (or not) using possibly invalid bbox.


Sophist-UK:
The implementation of the main fix is a bit after the fact. It might be better to do StepRepeat first and then add the pre/post operations after that has been done. But that would give you an Execute where the Pre&Post were not included on first display of the ops list.

tatarize (approved):
Yeah, this works fine and is pretty harmless. If I get around to something fancier it won't really be needed.

tatarize:
And you're right. It's delta and not gap. Since the latter would need to know the distance between the objects rather than the distance to the next object.

Sophist-UK:
Personally I think gap would be better i.e. you specify the distance between offsets not the origin, but that was a UI change I was not going to make without it having been discussed first.

tatarize:
I can't be sure we know the size. If you do this when it's already in cutcode form you don't actually know the size of the various parts.

PR #662, opened 2021-11-01
Preferences/About lose min/max buttons, save-as cookies, pane commands.

tatarize:
* About window sets an explicit style to suppress maximize and minimize buttons.
* Preferences window sets an explicit style to suppress maximize and minimize buttons.
* MWindow permits the setting of an explicit style.
* wxmmain window saves a `needs_saving` value. This is reset by loading a file, saving, going to file new. And is set false by anything in the project being saved.
* Pane commands.
     * In console, `pane create preferences` for the normal test of preferences as a panel.
     * There are others like pane show <pane>
     * `pane hide <pane>` -- Hides pane.
     * `pane float <pane>` -- Shows pane.
     * `pane reset` -- Resets all panes.
     * `pane lock` -- Locks all panes
     * `pane unlock` -- Unlock all panes.
     * `pane create <about/preferences` -- This was a test to create a pane from the various panes that are available. This is not expected to be used for anything.
* OSX Help includes launching static internal `help/meerk40t.help` file from either the pyinstaller directory or the py2app or source value.
* OSX Help gets a special "Online Help" that does normal help functionality.
* Kernel console_command() gets a long explanation of what it's supposed to do (Documentation).
* The console_command throws a MalformedCommandRegistration() error if you try to add a `@console_command` to `@console_command`.


joerlane (approved):
Checked with source and built binaries.

Sophist-UK:
Unfortunately (and it only happens sometimes) use of e.g. alt-f / alt-h conflicts with the shortcuts on the main menu (File / Help) and of course, in translations these might use other letters.

IMO we should probably avoid using alt-letter. Does anyone have preferred mappings for the current alt-letter usage?

tatarize:
There's some mac are restrictions or so with some commands got booted out for using various values at some point. I don't recall the specific tho.

PR #663, opened 2021-11-01
Add `image slash`, correct return values for `image slice`

tatarize:
Add in an `image slash` command to slash an image horizontally at a given y value.


tatarize:
Good to go.

tatarize:
* Load an image.
* Type "image slash 40%" in console.

Image will become 2 images sliced 40% of the way through the image. The top image can then be deleted and the bottom rastered to completion.

joerlane (approved):
Features are useful.

PR #669, opened 2021-11-05
Version 0.7.4-beta3

tatarize:
The idea here is to have the version bump waiting in the wings in case we want to push a new version.


joerlane (approved):
Version bump for beta release.

PR #670, opened 2021-11-07
Use SetValue rather than SetLabel for a TextBox

tatarize:


tiger12506 (approved):
This tested ok on my end (and fixed the issue)

PR #671, opened 2021-11-07
Add Set URI, fix crash issues.

tatarize:
* Add Set URI loader from right-click.
* Fix right click crash issues. %d format.


joerlane (approved):
Able to set IP cameras again.

PR #673, opened 2021-11-08
Update Translations

tatarize:
With regard to POBoy and other change to translation we should dump a bunch of potentially useful fuzzy  translations into meerk40t and make a push to have some of these corrected by users. While obviously online translations aren't perfect, there's either a gain of making a not terrible one saving some typing vs. a terrible one that requires the same amount of typing.


Sophist-UK:
Personally I am hoping that having a poor (automated) translation will encourage more users to use Meerk40t and then update the translations rather than take a look, realise that they cannot find a language that they understand and then uninstall. But I think this only really makes sense with the ability to do translations in the app.

But this is only a guess about how people will react.

Sophist-UK:
Do we have any priority list for translations?

Sophist-UK:
Given the proposed switch to str.format() I suggest that we hold off doing anything further (and even close this PR) until that change has been done.

tatarize:
I'm going to salvage what can be salvaged here and add these things in when they are conflicting the deepl translations are fine and should be included.

tatarize:
Closed in favor of #820

PR #674, opened 2021-11-08
Change string substitution from % to format

Sophist-UK:
In order for translations to work consistently substitution needs to be done using named parameters rather than position.

For example `_("%s is in %s")` might be translated in a different language to something meaning `"%s contains %s"` and a substitution like `_("%s is in %s") % (inside, outside)` which in english would give "inside is in outside" might result instead in "inside contains outside" rather than "outside contains inside" as it should. 

Instead we plan to do `_("{inside:s} is in {outside:s}").format(inside="inside", outside="outside")` which uses named parameters so that a translation "{outside:s} contains {inside:s}" will still result in the correct substition order.

The minimum required is to convert translation strings containing 2 or more variable substitutions. Ideally for consistency all translations should use `str.format()`, and ideally all `%` substitutions would be changed also.

**PLUS...**
This is also an opportunity to ensure that console command help strings do **not** have the command names translated (which is an error in current code). e.g. `@self.context.console_command("bind", help=_("bind <key> <console command>"))` should be `@self.context.console_command("bind", help="bind "+_("<key> <console command>"))`.

### DIFF
--- a/meerk40t/core/bindalias.py
+++ b/meerk40t/core/bindalias.py
@@ -27,7 +27,10 @@ class BindAlias(Modifier):
         self.context.default_keymap = self.default_keymap
         self.context.default_alias = self.default_alias
 
-        @self.context.console_command("bind", help="bind "+_("<key> <console command>"))
+        @self.context.console_command(
+            "bind",
+            help=_("Lists key bindings or adds/removes a binding of a command to a key.")
+            )
         def bind(command, channel, _, args=tuple(), **kwgs):
             """
             Binds a key to a given keyboard keystroke.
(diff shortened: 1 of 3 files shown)


Sophist-UK:
I am committing only the first file at this stage and request review of this to make sure that the changes are on the right lines.

tatarize:
This set of changes is still good but will need to be done in 0.8.0 given that a number of gui files are changed in there.

Sophist-UK:
Yes - I agree. Hopefully we will move to 0.8.0 beta 1 soon.

I think we will need a 0.7.5 first though as a final stable with all other changes before we switch to 0.8.

PR #675, opened 2021-11-12
0.8.0 - Kernel Services

tatarize:
Services are operational. We are down to rough edges 

- [ ] Check if two different lihuiyu drivers on the same raspberry pi can run two different non-conflicting servers on two different ports.
- [ ] Test all device lihuiyu driver commands again. `pulse`, `speed`, `power`, `acceleration`, `status`, `continue`, `pause`, `estop`/`abort`, `rapid_override`, `egv_import`, `egv_export`, `egv`, `challenge`, `start`, `hold`, `resume`, `usb_connect`, `usb_disconnect`, `usb_reset`, `usb_release`, `usb_abort`, `usb_continue`, `lhyserver`, `lhyemulator`


tatarize:
This should probably get reviewed. Should merge #669 and push out a new beta before merging this. Perhaps even complete 0.7.4 proper.

tatarize:
Currently the code seems to work. A few items are not clearly translated from legacy to new system.

* The lhyserver attaches to the output. New system has no output. That work is done entirely internally.
* Lhyemulator checks the fixspeeds variable which is not always a lhystudios context.
* LhystudiosEmulator hooks into the active %s/usb_send channels. These are not going to certainly exist.
* ExecuteJob, Simulation and LaserPanel call devices for the connected spoolers. This isn't compatible with non-legacy models.
* UsbConnect attaches to %s/usb this is not certain to exist.
* Window open for custom window types. `-o` `-d` `-s` does not necessarily have output, driver or source values to trigger custom windows.
* Devices panel matches and displays active vs. others in legacy driver mode.
* SpoolerPanel displays content based on the current selected device.

tatarize:
* The USB error failure launches Controller window but this could happen in a different device than the active devices. This was previously run through `window open -os {device} Controller` but that was deprecated.

tatarize:
* The active-legacy device custom window type also breaks the path values for the Controller and Configuration windows which would previously have launched at expected context paths to modify just those values.

tiger12506 (dismissed):
Review blocking is pointless. I care about whether or not it can merge cleanly, Github!!!!

Sophist-UK:
Just ran 0.8.0 on dev machine for first time - a few comments:

* Flip on ribbon bar - if we are going to add these sorts of buttons (and we probably should) then we are either going to need a LOT more real estate, or we need a dropdown bar for flip and a drop down bar for rotate etc.

* Bed size - enter in pixels?!!! WTF! This is not user friendly. Also appears ion both 

* Devices button on ribbon was before Config, not it is at the end.

* A lot of my config has not carried across to 0.8.0 - if we can carry config across that would be good.

* Device config - Fields shown should change depending on whether USB/Network/Mock is chosen. Also, do we need to have separate Configuration (should that be "Connection") and Setup (should that be ?) tabs or can we have all values displayed? (IMO a lot of people are going to miss that there is a 2nd tab.) And most of the Board Setup group probably belongs on the Setup tab.

* Pause and Stop buttons on the button bar have icons swapped around i.e. Pause has Stop icon and Stop has Pause icon. Flip Vertical and Horizontal icons are also the wrong way around.

Sophist-UK:
There is also a noticeable delay of several seconds on startup between the main window appearing and the software finishing displaying and being available for use.

tatarize:
> Flip on ribbon bar - if we are going to add these sorts of buttons (and we probably should) then we are either going to need a LOT more real estate, or we need a dropdown bar for flip and a drop down bar for rotate etc.

There's a lot of options for those and yeah, things like tools and subtools would probably need a lot more stuff and they'd be pretty easy to add.

> Bed size - enter in pixels?!!! WTF! This is not user friendly. Also appears on both Device Config and Preferences.

Bed size is in steps. Though yeah, certainly taken under advisement.

> Devices button on ribbon was before Config, now it is at the end - there are other similar moves but this one is the only one of consequence.

They are created dynamically though these elements are able to be configured for position. Or at least they should be able to be configured position-wise. 

> A lot of my config has not carried across to 0.8.0 - if we can carry config across that would be good.

Not sure how much will maintain between the versions given that it might be at an entirely different address and a lot of properties are device specific now.

Device config - Fields shown should change depending on whether USB/Network/Mock is chosen.

I believe this was the general idea but I didn't get around to turning the panels on/off depending on the selection.

>Also, do we need to have separate Configuration (should that be "Connection") and Setup (should that be ?) tabs or can we have all values displayed? (IMO a lot of people are going to miss that there is a 2nd tab.)

The second tab options should hopefully be unneeded or much more highly optional.

> And most of the Board Setup group probably belongs on the Setup tab.

It that's the case then setup tab should probably be moved to first tab or something.

> Pause and Stop buttons on the button bar have icons swapped around i.e. Pause has Stop icon and Stop has Pause icon. Flip Vertical and Horizontal icons are also the wrong way around.

That'll I'll fix post haste.

Sophist-UK:
> > A lot of my config has not carried across to 0.8.0 - if we can carry config across that would be good.
> 
> Not sure how much will maintain between the versions given that it might be at an entirely different address and a lot of properties are device specific now.

The Updater really needs to warn the user they may need to reconfigure when they update is from e.g. 0.7.x to 0.8.x. So we need a "rule" here - that if configs are not going to carry across, then we need to have a new second version (as we do in this case).

tatarize:
All changes between major versions can absolutely do that.0.6.x settings to 0.7.x won't carry either. While it can be coded up and made to work it's not exactly trivial since you've got to write code to perform the transitions and store the version numbers etc.

Issue #655, opened 2021-10-31
Optimise greedy improvement idea: Merge operation passes


Sophist-UK:
1. Rather than have permitted as a boolean, have burns_remaining as an integer
2. When Merge Passes is **not** checked, work as currently, i.e. iterate into separate cutcode objects, with burns_remaining = 1.
3. When Merge Passes **is** checked, only put cutcode-objects in once but with burns_remaining = passes.

The idea here is that merge passes does NOT increase the number of objects to sort / search through.

I will try to find the time to give this a go tonight.

Sophist-UK:
P.S. I don't think this will impact anything if greedy is on (because it will produce multiple copies) but not sure what happens if user sets greedy off in console with a `set`.

tatarize:
This actually quite clever and quite actionable. This will also not duplicate out the data in memory or cost any of that time which is beginning to be the larger timesuck. Candidate() just means that everything inside the the routine has reached zero, the number of next and previous links will be reduced as will the value of N for anything of any time complexity.

This will also speed things up in another way that might not be as desirable but would certainly be the case. If you're going to 10 circles as your job, and these all perfectly overlap because they are 1 circle with 10 passes. We'd store this as a linked set of 10 circles and the cuts they break down into namely some cubic rasters, then the object themselves will be 1 object and consequently the loops the next and prev connection have will be used for additional passes bypassing the distance check altogether. Making the fastest method, basically by definition doing all 10 passes in constant time and no searching.

This is *not* what you wanted with your leather since it gave things the chance to burn however, this tweak opens up *another* change. We can implement merge passes into something closer to what you originally wanted without any any issues. What we can do is level the cutcode. We do not permit elements to be candidates *unless* their burns_remaining is the largest number of burn remaining.

So if there are anything with 5 burns left, and nothing with 6 only things with 5 burns left are candidates for inclusion into the plan.

Sophist-UK:
Implemented in #677.

Issue #528, opened 2021-08-29
v0.7.0 RC10: Please report any issues found here


Sophist-UK:
# PLEASE DO ALL FURTHER TESTING ON v0.7.0b35 RC-10
The latest beta release is v0.7.0beta35 Release Candidate 10. This issue is for reporting issues with this specific version of MeerK40t.

Please do all further testing on this release. Please do not report any further bugs with previous beta releases. Do give any other types of feedback, questions, concerns or comments.

https://github.com/meerk40t/meerk40t/releases/tag/0.7.0-beta35

# CAUTION
**CAUTION:** Please note that (depending on your design) if you use Merge Operations after you have used Duplicate Operation, or you use Merge Passes when you have set Operation Passes > 1, these can result in the laser beam burning the same path several times in quick succession without allowing time for the material to cool down in between. This can have unintended consequences. **Please do not use this feature of the beta whilst your K40 is unattended.**

Sophist-UK:
# Bugs
1. Execute two operations from right-click with Merge Operations NOT checked, and operations are merged nonetheless.
2. Path with v0 segment (and presumably h0) is held by svgelements as l0,0 and causes closed paths with even number of Merge Passes to be burned piecemeal i.e. back-and-forth rather than round-and-round.

Sophist-UK:
# Bugs
1. Do not allow user to cancel the current op in Spooler - it seems to delete the following op or two.

tatarize:
![IMG_20210902_131357](https://user-images.githubusercontent.com/3302478/131910179-39c0d2a7-efd4-4f6a-85e0-b3063f56cebc.jpg)

No critical lines found in sven's report but he clearly has them.

![image](https://user-images.githubusercontent.com/3302478/131910603-b9424e6a-1078-4230-92cb-fd01ef5770aa.png)

Further investigating...

tatarize:
![3pass](https://user-images.githubusercontent.com/3302478/131913227-8a80edf5-7d85-4fdf-87bb-31575cfa3040.png)

tatarize:
* After restore, double click on cutcode in operations does not simulate it correctly. Cutcode is not flagged as optimized and will try to optimize anyway.
* The execute should detect automatically done stages and not require a click for them.

tatarize:
```
MeerK40t crash log. Version: 0.7.0 RC-8 on win32
Traceback (most recent call last):
  File "mk40t.py", line 13, in <module>
...
[...]

PR #677, opened 2021-11-15
Merge fixes and optimisation improvements

### LINKED ISSUE #656: [Bug] Execute with Merge Operations and not Merge Passes does not merge Operations

Sophist-UK:
1. Fixes various merge issues, specifically:
    a. Merge operations not working when passes >=2 and merge passes not set (because CutCode sequence was inappropriate)
    b. Operations being merged when Merge passes set and Merge ops not set because operations of same type were being incorrectly merged. Fixes #656 and reported in #528 and #465.
    c. Merge passes reversing direction unncessarily.

2. Improves Greedy optimisation with Merged Passes by reducing the size of the CutCode object being optimised by holding only a single copy of each CutGroup and continuing to burn this object rather than having to find the next copy of the same CutGroup. Implements #655.

### DIFF
--- a/meerk40t/core/cutcode.py
+++ b/meerk40t/core/cutcode.py
@@ -134,7 +134,7 @@ class CutObject:
     reversed.
     """
 
-    def __init__(self, start=None, end=None, settings=None, parent=None):
+    def __init__(self, start=None, end=None, settings=None, parent=None, passes=1):
         if settings is None:
             settings = LaserSettings()
         self.settings = settings
@@ -145,6 +145,7 @@ class CutObject:
         self.next = None
         self.previous = None
         self.permitted = True
+        self.burns_remaining = passes
 
         self.mode = None
         self.inside = None
@@ -219,7 +220,7 @@ class CutObject:
             return False
         for c in self.contains:
             for pp in c.flat():
-                if pp.permitted:
+                if pp.permitted and pp.burns_remaining >= 1:
                     return True
         return False
 
@@ -227,7 +228,7 @@ class CutObject:
         yield self
 
     def candidate(self):
-        if self.permitted:
+        if self.permitted and self.burns_remaining >= 1:
             yield self
 
 
@@ -238,7 +239,8 @@ class CutGroup(list, CutObject, ABC):
     """
 
     def __init__(
-        self, parent, children=(), settings=None, constrained=False, closed=False
+        self, parent, children=(),
+        settings=None, constrained=False, closed=False
     ):
         list.__init__(self, children)
         CutObject.__init__(self, parent=parent, settings=settings)
@@ -283,7 +285,7 @@ class CutGroup(list, CutObject, ABC):
 
     def candidate(self):
         """
-        Candidates are permitted cutobjects permitted to be cut, this is any cut object that
+        Candidates are permitted cutobjects with at least one burn remaining, this is any cut object that
         is not itself containing another constrained cutcode object. Which is to say that the
         inner-most non-containing cutcode are the only candidates for cutting.
         """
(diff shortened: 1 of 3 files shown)


Sophist-UK:
Not working quite right with step_repeat and a special operation beep between two other ops.

tatarize:
I would go ahead and remove permitted and only do burns remaining, since it's a largely improved system.

Sophist-UK:
I left it in because it is used by `inner_selection_cutcode` and I didn't want to touch that (because I could not work out what it was doing). 

It appears to make every line segment a candidate, then gets the list of candidates (i.e. everything), adds all of them to the output list, then makes all of them not a candidate, then checks to see if there are any candidates left (no) and finishes. But that is clearly not what is happening, so I don't understand the code - and so I left it in rather than break something.

Sophist-UK:
I noted that 2-opt actually introduced a lengthy travel move in one of my tests.

I have spent several hours testing this code with all combinations of optimise options and checking that the simulation was exactly as expected. At least as far as my testing goes, this now seems to be solid.

meerk40t/core/cutcode.py
```diff
@@ -219,15 +220,15 @@ def contains_uncut_objects(self):
             return False
         for c in self.contains:
             for pp in c.flat():
-                if pp.permitted:
+                if pp.permitted and pp.burns_remaining >= 1:
```
tatarize, line 223: pp.permitted should probably be trashed.

tatarize (approved):
Burns remaining should likely completely replace permitted since all permitted did was flag whether a thing was in the list or not.

meerk40t/core/cutcode.py
Sophist-UK: See my comment above about `inner_selection_cutcode` - if I understood what it was doing, then I can probably remove `permitted` altogether, but I have learned the hard way not to touch stuff unless I know for sure what it is doing.

Sophist-UK:
I am going to merge this anyhow. We can remove `permitted` in a future PR.

tatarize:
Literally the only thing permitted does is a boolean between 0 and 1 burns remaining. Setting that to an integer does the same thing. You could set permitted to 1 and then set it to 0 when the burns_remaining reaches 1. It literally does the same thing as yours but with a bit more naive usage.

Sophist-UK:
Yes - but what does `inner_selection_cutcode` do?

Sophist-UK:
I can't set burns_remaining to 1 at the beginning of greedy - like we currently do with permitted - because burns_remaining can be >1 because with Merge Passes checked it is equal to the value of operation passes. And `inner_selection_cutcode` also sets and clears permitted.

Which is why I need to understand what `inner_selection_cutcode` is doing.

Sophist-UK:
Additional run for new code 1 pass with and without merges - these are consistent with the original code.

1. One pass, no merges

[22:50:17] Executing Greedy Short-Travel optimization
[22:50:17] Length at start: 2648966.038968
[22:51:20] Length at end: 81466.042493, optimized in 62.833405 seconds
[22:51:20] Executing Greedy Short-Travel optimization
[22:51:20] Length at start: 6450.561009
[22:51:20] Length at end: 0.020111, optimized in 0.012998 seconds

2. One pass, with merges

[22:22:32] Executing Greedy Short-Travel optimization
[22:22:33] Length at start: 2650756.095405
[22:23:35] Length at end: 81466.099295, optimized in 62.146681 seconds

Sophist-UK:
I have realised that there will be an issue with step_repeat with this code - because the cutobjects are shared across repeats and the burns_remaining value will not be reset. I have an easy fix in mind and will create a new PR to resolve this this evening. This new PR will also address the very minor review comments I just submitted.

PR #679, opened 2021-11-16
0.7.4 Finalize

tatarize:
Finalize changes for 0.7.4

* Check through all crash logs.
* Fix optimize checkbox on main window
* Fix restore color picker for OperationProperty


tiger12506 (approved):
It's my understanding that everyone is satisfied with these changes. Looks good to me.

PR #685, opened 2021-11-19
Limit Resource Usage

tatarize:
We should be more sparing with resources.

Might be out of minutes for the month.
OSX unit tests do not complete.
Unittests do not have Timeout.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -27,6 +27,7 @@ jobs:
 
     name: ${{ matrix.os }}+py${{ matrix.python-version }}
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 5
     strategy:
       fail-fast: false
       matrix:


tatarize:
The red x means it's working. Cancelled after 5 minutes.

tatarize:
...

tatarize:
Report said we've used 0 of our minutes for private repositories. Still limiting is preferred.

tatarize:
...

Sophist-UK:
I don't disagree with setting a limit in case we get some sort of infinite loop. We should not waste resources unnecessarily.

However since our repositories are all public, there is no monthly limit whatsoever, hence the private usage so far this month being zero.

tatarize:
So approve and merge. Rather than leave it here and have additional things run for 6 more hours.

joerlane (approved):
You have been doing bad things... Very bad things.

PR #689, opened 2021-11-20
Minor batch and console improvements

Sophist-UK:
1. Allow batch to include "channel print console" to retain console printing
2. Indent command output with 4 spaces to distinguish commands from their output

### DIFF
--- a/meerk40t/kernel.py
+++ b/meerk40t/kernel.py
@@ -2062,17 +2062,18 @@ class Kernel:
                     channel(command_name.split("/")[-1])
 
         @self.console_command("loop", help=_("loop <command>"))
-        def loop(remainder, **kwargs):
+        def loop(remainder=None, **kwargs):
             if remainder:
                 self._tick_command(remainder)
 
         @self.console_command("end", help=_("end <commmand>"))
-        def end(args=tuple(), **kwargs):
-            if len(args) == 0:
+        def end(remainder=None, **kwargs):
+            if remainder:
+                self._untick_command(remainder)
+            else:
                 self.commands.clear()
                 self.schedule(self.console_job)
-            else:
-                self._untick_command(" ".join(args))
+
 
         @self.console_option(
             "off", "o", action="store_true", help=_("Turn this timer off")
--- a/meerk40t/kernel.py
+++ b/meerk40t/kernel.py
@@ -21,7 +21,6 @@ STATE_SUSPEND = 6  # Controller is suspended.
 STATE_WAIT = 7  # Controller is waiting for something. This could be aborted.
 STATE_TERMINATE = 10
 
-
 _cmd_parse = [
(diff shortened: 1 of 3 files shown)


meerk40t/kernel.py
```diff
@@ -2647,9 +2646,11 @@ def __repr__(self):
             repr(self.line_end),
         )
 
-    def __call__(self, message: Union[str, bytes, bytearray], *args, **kwargs):
+    def __call__(self, message: Union[str, bytes, bytearray], *args, indent=True, **kwargs):
         if self.line_end is not None:
             message = message + self.line_end
+        if indent and not isinstance(message, (bytes, bytearray)):
+            message = "    " + message.replace("\n", "\n    ")
```
tatarize, line 2653: It seems like indent should maybe be part of the channel values. But, I guess you could want indenting and non-indenting.

meerk40t/main.py
```diff
@@ -340,10 +340,17 @@ def run():
     if args.batch:
         # If a batch file is specified it gets processed here.
         kernel_root.channel("console").watch(print)
+        unprint_console = True
         with args.batch as batch:
             for line in batch:
-                kernel_root(line.strip() + "\n")
-        kernel_root.channel("console").unwatch(print)
+                line = line.strip()
+                if line == "channel print console":
```
tatarize, line 347: This solution is too hard coded to work.

While it's not terrible, there might be other options are arguments in the channel or print commands (there currently are not) which means you can't do that. Also you could have done something in -e `channel print console` or in a batch file. This is a pretty narrow ranged bug but it might be more correct to write a local function that delegates to print and just add that and remove that within this. That way you can be sure that unwatching will remove that print. Since it's not the same print function that `channel print console` adds.

This is solidly identified but too kludgy of a solution to too narrow of a bug. The better solution might actually just to be do not watch print for executing the batch if you weren't asked to. Or, use a print delegate, such that `my_print_function not is print` and you can add multiple prints there. Or just echo the lines for the user.

tatarize (changes requested):
Too hard coded of a solution to too narrow of a bug.

meerk40t/kernel.py
Sophist-UK: The idea here is to assist the user in separating commands from output - I experimented with various solutions before deciding to go with indentation. I am very open to alternative suggestions.

Default text is indented and input commands are unindented from that.

tiger12506: In actual console, commands have the > prompt in front of them. That makes more sense to me, to replicate in the console window (i.e. any command that is run and appears in the console window without you having typed it gets a > in front of it.)

tatarize (approved):

tatarize:
Other one seemed weird, but harmless and the kludgy one was fixed up better than I might have done so. Checking for other print watchers is a nice touch.

PR #690, opened 2021-11-21
0.7.5 beta1

tatarize:
* Version bump
* Laserpane combo fix
* Right click open in system (needs checking for other OSes).
- [x] Run in System: MSW
- [x] Run in System: rPi
- [x] Run in System: Linux
- [x] Run in System: OSX

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -5125,13 +5125,20 @@ class Elemental(Modifier):
             node.remove_node()
             self.load(filepath)
 
-        @self.tree_operation(_("Open in System: '%s'") % "{name}", node_type="file", help="")
+        @self.tree_operation(
+            _("Open in System: '{name}'"),
+            node_type="file",
+            help=_(
+                "Open this file in the system application associated with this type of file"
+            ),
+        )
         def open_system_file(node, **kwgs):
             filepath = node.filepath
             normalized = os.path.realpath(filepath)
 
             from os import system as run_in_shell
-            run_in_shell('"%s"' % normalized)
+
+            run_in_shell('"{file}"'.format(file=normalized))
 
         @self.tree_submenu(_("Duplicate element(s)"))
         @self.tree_operation(_("Make 1 copy"), node_type="elem", help="")
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -5136,15 +5136,15 @@ class Elemental(Modifier):
             filepath = node.filepath
             normalized = os.path.realpath(filepath)
 
-            from os import system as run_in_shell
+            from os import system as open_in_shell
             from sys import platform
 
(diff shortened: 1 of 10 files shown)


tatarize:
This should fix most of the desync issues @apbarrett identified.

meerk40t/core/elements.py
```diff
@@ -5124,6 +5125,14 @@ def reload_file(node, **kwgs):
             node.remove_node()
             self.load(filepath)
 
+        @self.tree_operation(_("Open in System: '%s'") % "{name}", node_type="file", help="")
```
Sophist-UK, line 5128: May I suggest `@self.tree_operation(_("Open in System: '{name}'").format(name=name), node_type="file", help=_("Open this file in the system application associated with this type of file"))`

meerk40t/core/elements.py
```diff
@@ -5124,6 +5125,14 @@ def reload_file(node, **kwgs):
...
+        def open_system_file(node, **kwgs):
+            filepath = node.filepath
+            normalized = os.path.realpath(filepath)
+
+            from os import system as run_in_shell
+            run_in_shell('"%s"' % normalized)
```
Sophist-UK, line 5134: `run_in_shell('"{file}"'.format(file=normalized))`

meerk40t/core/elements.py
tatarize: The {name} would get replaced wrong there. It's actually using {name} in the tree to replace that with the file's name. In a mock sort of operation thing.

tatarize:
Issue report for the beta
```
The current 0.7.5-beta1 doesn't remember the device selection now. And I'm not sure the "fix" is totally working (link between devices and laser pane). I had to restart meerk40t a few times and I still can't explain the results

Just know that it starts up with spooler 4 selected, and nothing listed in the main window title bar.
```

meerk40t/core/elements.py
Sophist-UK: Ah - ok. Does that mean it could be simplified to: `@self.tree_operation(_("Open in System: '{name}'"), node_type="file", help=_("Open this file in the system application associated with this type of file"))`?

tatarize: Yeah, and in fact it should. We're going to allowing translators to move those `.format()` stuff around. So this shouldn't be any different.

tatarize:
Paiin's right, the reload doesn't automatically select the correct item in the laserpane. It's initialized wrongly.

tatarize:
And fixed that one. Best I can tell this works fully now. And could be merged and versioned...

tiger12506 (approved):
This all seems to work now.
Open in system tested on debian buster, ubuntu 18.04, mac 10.14.4, windows
device window/laserpane sync tested debian, mac.

PR #691, opened 2021-11-21
Dual Scene Test

tatarize:
Suggested by @apbarrett. Two window mode. This basically works by swapping the .elements service and refreshing the tree.

Most of the changes to force this to work are somewhat minor and while this shouldn't be implemented and isn't well coded, the proof of concept is there. I got it working in about an hour.

### DIFF
--- a/meerk40t/gui/wxmscene.py
+++ b/meerk40t/gui/wxmscene.py
@@ -32,6 +32,16 @@ def register_panel(window, context):
     pane.dock_proportion = 600
     pane.control = notebook
 
+    def on_notek_page_change(event=None):
+        if notebook.GetPageText(notebook.GetSelection()) == "scene1":
+            context.kernel.activate_service_path('elements', 'elements')
+        else:
+            context.kernel.activate_service_path('elements', "elephants")
+        context.signal("rebuild_tree", 1)
+        context.signal("request_refresh", 1)
+        context("refresh\n")
+    notebook.Bind(aui.EVT_AUINOTEBOOK_PAGE_CHANGED, on_notek_page_change, notebook)
+
     window.on_pane_add(pane)
     context.register("pane/console", pane)
 
--- a/meerk40t/gui/wxmscene.py
+++ b/meerk40t/gui/wxmscene.py
@@ -32,15 +32,13 @@ def register_panel(window, context):
     pane.dock_proportion = 600
     pane.control = notebook
 
-    def on_notek_page_change(event=None):
+    def on_note_page_change(event=None):
         if notebook.GetPageText(notebook.GetSelection()) == "scene1":
             context.kernel.activate_service_path('elements', 'elements')
         else:
             context.kernel.activate_service_path('elements', "elephants")
-        context.signal("rebuild_tree", 1)
-        context.signal("request_refresh", 1)
         context("refresh\n")
-    notebook.Bind(aui.EVT_AUINOTEBOOK_PAGE_CHANGED, on_notek_page_change, notebook)
+    notebook.Bind(aui.EVT_AUINOTEBOOK_PAGE_CHANGED, on_note_page_change, notebook)
 
     window.on_pane_add(pane)
     context.register("pane/console", pane)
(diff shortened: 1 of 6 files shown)


meerk40t/gui/wxmscene.py
```diff
@@ -22,27 +22,38 @@
...
 
 def register_panel(window, context):
-    # self.notebook = wx.aui.AuiNotebook(self, -1, size=(200, 150))
-    # self._mgr.AddPane(self.notebook, aui.AuiPaneInfo().CenterPane().Name("scene"))
-    # self.notebook.AddPage(self.scene, "scene")
+    notebook = wx.aui.AuiNotebook(window, -1, size=(200, 150))
+    panel1 = MeerK40tScenePanel(window, wx.ID_ANY, context=context, index=1)
+    notebook.AddPage(panel1, "scene1")
+    panel2 = MeerK40tScenePanel(window, wx.ID_ANY, context=context, index=2)
+    notebook.AddPage(panel2, "scene2")
```
tatarize, line 29: These changes are very kludgy and just for the test.

meerk40t/gui/wxmtree.py
```diff
@@ -205,18 +205,16 @@ class ShadowTree:
     reflected in the tree, the shadow tree is updated accordingly.
     """
 
-    def __init__(self, context, gui, root, wxtree):
+    def __init__(self, context, gui, wxtree):
         self.context = context
-        self.elemtree = root
+        self.context.kernel.shadow_tree = self
```
tatarize, line 210: Remove hard coded elements

tatarize (commented):
Many of these changes are pretty reasonable. And would keep this closer to working and make similar weird things in the future still work.

Sophist-UK:
I am unclear what the purpose of this is. Is there a link to Andy's suggestion?

tatarize:
He asked about a second scene. Or rather a second window within the same run of meerk40t. He apparently has dual lasers and is running 2 things that are not the same thing. And sort of wants to see what's going on, or what part it's working on without switching the file he's currently running.

To do that correctly it needs a second elements thing since that's all setup to be singular and it needs a second scene in a notebook. But, both of those worked pretty well and were fairly easy in 0.8.0x.

tatarize:
apbarratt — Today at 3:35 AM
Can we add a "new window" option?  Because right now I have to right click my way inside of the mac app contents and launch meerk40t from within to get a second instance going.

Tat — Today at 3:53 AM
What do you mean specifically by "new window" like running two instances is weird or something on the mac?

apbarratt — Today at 3:53 AM
Well on a mac if I just try to open Meerk40t when I've already got an instance running, it just ignores me, because I've already got one running.  I can do it from terminal or drilling into the app contents folder though.

Admitedly, running multiple instances of a single app isn't really something I can think of doing much with any other app on a mac.

--

With the services thing, I can easily swap out parts like `elements` so I gave it a try. Made a second scene, and second elements thing and swapped out the service when I switched pages. Took a bit, and found some unneeded rough edges but it would be possible to just add several scenes and have them run perfectly fine.

Not sure the long term utility of that. But, maybe you'd want to load a different project in a different tab or something. It might make more sense if there were more tools.

tatarize:
Mostly a proof of concept test, since it might figure out some more bugs that could be lurking in the services code. It only found some too tight coupling.

Sophist-UK:
This seems like a complicated alternative when you could just run two instances of MK (and either have separate settings files or have both lasers defined in a common settings file and simply select the device when you start MK).

There was a separate discussion about running 2x k40 on a remote server - and TBH I would run 2x MK there too, as I would personally prefer to have separate IP addresses for each laser, and because I would prefer to let the O/S handle ensuring that the drivers get enough CPU rather than rely on Python's temperamental GIL-limited multi-threading implementation.

tatarize:
I am not sure it's worth giving the choice to the user, but it's worth knowing if the software hits any trouble with it. Hence the test. And it's the thing I'm currently working on, so it's good to know.

tatarize:
Implemented and retested in main 0.8.0 branch.

Issue #73, opened 2019-10-01
Update checker


tatarize:
Would require server checks etc, but it was suggested. Might as well leave the issue here and see if anybody comes out strongly in support or anything else.

tatarize:
Could also be used to help deal with modules. Where is there is a dxf reader but it has a requirement that you don't need that module. However the pre-packaged stand alone builds would kinda need them all to work.

tatarize:
https://api.github.com/repos/meerk40t/meerk40t/releases/latest

Would be fairly doable. Just grab and parse the json there.

Sophist-UK:
That is the plan - or part of a slightly more sophisticated plan anyway.

Sophist-UK:
Update checking now has a UI. I had in mind a download and replace capability as well, but there are several complications with current executable location being writeable and on windows the current executable being locked.

ALso some users like me put the version number in the filename in order to keep multiple versions available, and (of course) you probably want to provide backup/backout capability in case new version is broken.

PR #693, opened 2021-11-23
Basic check for updates

Sophist-UK:
1. New Preference for checking for beta versions
2. Checks for beta if above setting checked or currently (temporarily) on a beta.

To-Do:
3. GUI ability to check for updates and report new version
4. GUI automated periodic (daily) check for new version, weekly reminders of previously reported new version
5. Download new version(s) if executable, and executable directory is writeable and new versions not already downloaded (with correct size).

### DIFF
--- a/meerk40t/extra/updater.py
+++ b/meerk40t/extra/updater.py
@@ -1,0 +1,147 @@
+import json
+import sys
+from urllib.request import urlopen, Request
+
+def plugin(kernel, lifecycle):
+    if lifecycle == "register":
+        _ = kernel.translation
+
+        # https://docs.github.com/en/rest/reference/repos#get-the-latest-release
+        GITHUB_LATEST = "https://api.github.com/repos/meerk40t/meerk40t/releases/latest"
+        # https://docs.github.com/en/rest/reference/repos#list-releases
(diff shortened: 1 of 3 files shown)


tatarize:
I might put the preference choice element in the updater script. The choice items will append if they have the same name (I think, might be an 0.8.0 change). So you could store everything in updater.py. But that all seems by definition harmless. 

#73

tatarize (approved):

Sophist-UK:
It's a work in progress. Still neds a GUI option (pretty much coded) and an automated check option (in progress) and download functionality (To do)./

Sophist-UK:
At present we cannot (AFAIK):

a. Put the GUI stuff associated with the updater (i.e. menus) into the plugin file.
b. Put the preferences into the plugin file (and have them appear in the Preferences window).

tatarize:
0.8.0 can have gui stuff added. Though currently the mechanism is a bit cumbersome if you want to to have gui and non-gui features. These are currently being shoved into main as a second plugin dependant on the other plugins since requiring wx won't work well out of the gate. But, if you want to include some menu option stuff that seems to sort of work in 0.8.0. Or at least all the menus are currently dynamic for windows and panes.

Sophist-UK:
Ok - let's merge this. But I am hopeful that the final version I am still working on will be elegant.

PR #694, opened 2021-11-24
Fix for mac and linux open in system

tiger12506:
The issue was using the double-set of quotes. Rather than running
$ xdg-open file.svg

it was trying to run
$ "xdg-open file.svg"

which doesn't work. Instead, do this:
$ xdg-open 'file.svg'

Tested working on 2 different linux distros and mac.
Didn't test on pi, but xdg-open exists there, I see no reason why it wouldn't work.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -5140,11 +5140,11 @@ class Elemental(Modifier):
             from sys import platform
 
             if platform == "darwin":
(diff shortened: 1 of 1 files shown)


Sophist-UK:
I haven't tested this code on Windows, but windows needs double quotes around a file which has spaces in it not single quotes.

Sophist-UK:
It was fixed in a later commit.

Issue #34, opened 2019-09-23
Undo/Redo ability.


tatarize:
Control Z, shift Control Z. 
Undo and Redo elements.

 Etc.

Sophist-UK:
👍

tatarize:
This was added in 0.8.2

PR #695, opened 2021-11-24
Undo / Redo processing.

tatarize:
In theory this should somewhat work. It bites a couple bullets.

1) It restores elements only, without depth.
2) It requires the 0.8.0 fixes for scheduler.
3) It recreates the nodes so they are reclassified.
4) Only works with the elements. Doesn't work when element is added or removed (only modified or altered).


tiger12506:
My personal opinion on this is it should really try to be an all or nothing thing. I think that having an undo that can cause changes by reclassifying would cause confusion for a user.

Only way to really do that is to create an object that represents the entire state of the application as a whole, and just save those on a stack as changes are made. I don't think memory is going to be as big of a concern as you think it is -- but if it is, you can have an X-level deep undo stack, rather than infinite undo...

Sophist-UK:
I agree with @tiger12506 - this might create more confusion that it resolves. I haven't tested this but I expect that there will be a tranche of use cases where there is no undo - in which case an undo might undo something several operations ago that the user has moved on from.

tatarize:
Yeah, it needs to be able to make a fully copy of the elements tree. Then undo will restore the tree correctly and the structure. 

0.8.0 has a nice set of methods to draw things a bit, so I figured an undo would be nice. And I tried to see what it would take. The answer seems to be not too much beyond making additional copies of the tree.

The current limit is 25. After that the undos are trashed. It needs a bit better copying to make full copies of the tree. Likely maybe also incorporating the elements of `full save` I've been meaning to do so that the program state can be saved a bit better. So all the items in the tree would need to serialize and deserialize correctly and save to disk.

Sophist-UK:
Simply saving a shallow copy of the elements tree won't undo Operations changes or actualisation of elements.

A deep copy of the elements tree would save element actualisation, but would use a massive amount of memory with images.

I think undo/redo functionality would require having undo and redo stacks and saving links to undo / redo functions and the old and new states of whatever changed.

tatarize:
A shallow copy covers most of it since the objects within the tree items get swapped out. You could reify a path or actualize and image and that would change the core contents but for the most part all the movement of the objects is done with the matrix which absolutely can do that.

This current PR is likely not the way to go, but currently does work. It even works with images. Since RasterWizard on an image replaces that image. The shallow copy has a reference to the old image. Which keeps it around. There's no point doing a deep copy of an image each time you move it. The movement is a manipulation of the matrix.

It just needs to keep some additional things like the tree structure and it should be good for full undo. There might be an edge case where reify doesn't call `altered` when it should. Since `altered` means the internal structure of the object has changed. And I think `reify` is called internally on the path segments which means it would change the path without replacing the path. So the old reference in an old tree would now be a reference to the new object. But, those are pretty minor edge cases, and we could just make sure to ditch the old object on any `altered` operation.

Sophist-UK:
Hmmm... I think you would have to be absolutely certain that no change modifies any object one or more levels below the shallow copies, and that if you copy the object before making the change that it is not two or more levels below the shallow copies. I foresee lots of bugs being reported where an undo does something different from expected. And a change of the transform matrix would NOT be covered by a shallow copy of the tree.

It's a lot of work to create undo/redo functions and put them on the stack, but there would be some benefits:

a. The Undo/Redo stack could come with descriptions i.e. rather than just "Undo" and "Redo" the entries could be "Undo/Redo Move of Element id from Cut2 to Engrave" or "Undo/Redo Move element id by (+2mm, -4mm)" etc. and these could even be translateable.

b. Because you are only storing the minimum information needed to undo, the data stored on the undo/redo stack will be relatively small, however complex the design. It would not need to be limited to the most recent 25 changes). The stack could effectively be infinite and (by adding memory size of the undo/redo stack objects) you can probably monitor memory and remove the oldest undo changes from the stack if memory starts getting short.

c. I suspect that users will start reporting Undos that don't work - because they modify an uncopied object - and that this will become a headache. (When a programmer says something will "mostly work" I start to worry, LoL.)

d. There is a BIG issue of user perceptions here. A) users need to have a simple concept in mind - that undo will always undo the single most recent change, nothing more nor less. If we fail to save a change and then you undo the second most recent change, confusion arises. B) Users need to have a consistent experience that full releases are fully reliable, and beta releases are almost fully reliable. If we do create an Undo function that has a significant number of bugs, then this will undermine user confidence.

tatarize:
Part of the point here is that shallow vs. deep copy is a spectrum. I absolutely need the matrix and the SVGElement. I can get away with a reference to the image data since that's not going to change unless the image does, in which case the reference points to the old object correctly.

a. This might be a slightly hard since the `modified` and the `altered` signals which are used pretty universally do not actually come with a description of *what* they did. But, this certainly *could* be done. It would just take a bit of work. These operations are always done inside meerk40t and the could absolutely be passed the data. Though the undo stack mark record is done 1 second after you stop changing the file so it doesn't do things while it's moving or rotating or resizing but only after you stop doing that. Getting that information into a readable format would be a some work. Like merging a lot of signals together "Object id="Path1" was moved, rotated, and resized." from a series of quick matrix manipulations.

b. The minimum information might help if the process is hard. We could store records of which things changed and *just* restore that. But, it might be pointlessly hard for no real benefit since the tree especially without the hard-linked deep copies is probably pretty small in memory. I am unsure if python allows you to store things like `SoftReference` which is one of those weird memory management things that rarely matter but in weird cases.

c. Clearly we just need to make the 98% working Undo a secret feature that nobody can access. Obviously. It just needs to get another class of changes that isn't `modified` and `altered` but something that indicates the structure of elements has been changed in some way and record a undo-state at those events too.

d. You can try this PR. It does *most* of what should be required and just messes up some stuff since it doesn't do things like record operations changes or something, and loses the tree structure. We do sort of need it for drawing tools etc. Especially if the user is going to be potentially crafting inside meerk40t which wouldn't be too hard to implement.

I've got no real issues coding this up and just including it in betas for 0.8.1+ and turning it off otherwise. But, this PR was like an hour of work. And for a few more I'm pretty sure most of the kinks would be out. #34 is a pretty old PR and it's certainly a desired feature. And if this didn't eat people's tree structure it would already be pretty okay. This isn't intended to be the fully working version, just a test to see how much work would be needed to make one of these work. The answer turns out to be around 3 hours.

* It needs to copy the elements.tree structure and be able to restore the elements.tree structure
* utilize this undo stack code here
* Correct the signals for when elements are added or removed from the tree, and likely if you alter a property in operations.

tatarize:
This will probably be integrated in 0.8.x at some point with full tree copy.

PR #696, opened 2021-11-24
Killed off the creepy "ding" on Macs.

apbarratt:
No longer shall our Macs sound like the ghost of Stephen Hawking exclaiming the word "DING!"


tatarize (commented):
Seems fine. platform.system() == "Windows" is correct.

tatarize (approved):
Windows hits no issues with that.

PR #698, opened 2021-11-26
Mac system alerts

joerlane:


tatarize (approved):

tatarize:
@joerlane I stopped merging other folks PRs for the most part if they have write access and can merge it themselves.

tatarize:
@joerlane merging this is fine... but I don't merge other folks PRs anymore if they can merge things themselves.

meerk40t/core/drivers.py
```diff
@@ -292,7 +292,19 @@ def command(self, command, *values):
                     except Exception:
                         pass
                 elif OS_NAME == "Darwin": # Mac
-                    os.system('afplay /System/Library/Sounds/Ping.aiff')
+                    """os.system('afplay /System/Library/Sounds/Ping.aiff')"""
+                    """os.system('plutil -p ~/Library/Preferences/.GlobalPreferences.plist | grep com.apple.sound.beep.sound | cut -d \'"\' -f 4 | xargs afplay')"""
+                    os.system('osascript -e \'display notification "Job Finished" with title "MeerK40t"\'')
```
Sophist-UK, line 297: I am positive about this PR - users need something to happen when there is a beep - but request the following changes:

1. This says "Job finished" when the job may not be finished if the Beep is mid-job. I think the notification should be "Meerk40t beep".
2. For people with multiple lasers, it would be useful to know which laser has beeped, and possibly the name of the project that has just finished.
3. The string(s) need to be translatable.

In addition, I have the following more general comments which are for another day / another PR (or this one if optionally you want to add to it):

4. We probably should have a proper "Job Finished" notification function when the spooler goes idle - and it would be great to have this generalised to create Windows notifications too.
5. As an aside, my Octoprint instance for my 3D printer creates pushbullet notifications too. Of course it is far safer to leave a 3D printer unattended than a K40, so I am very unsure whether we should encourage unattended operations by providing remote notifications.

Sophist-UK (changes requested):

meerk40t/core/drivers.py
Sophist-UK: P.S. For a generalised Job Finished message, we may need to decide between a generalised popup using wx.adv.NotificationMessage and o/s specific functionality (like this).

tatarize:
Yeah, at this rate we could also do the correct thing of making those things into commands.

Beep as a command would beep on all OSes. And the display notification thing is also a thing in windows so if we had a display notification command thing we could do different things on different OSes.

Sophist-UK:
Not sure whether beep special operation should run a beep command - we need to be careful about synchronicity. But in principle, that would be more inline with Meerk40t's command philosophy. Also, I think this PR is because there is no way yet found to do an actual sound beep on mac. (I can't quite image why this is, but since I have not used a Mac for more than 3 decades I have no actual knowledge.)

But I agree that we should implement generic a `notify command` which will generate some sort of balloon / toast notification, ideally using O/S functionality (so it will also work in non-GUI) but with use of `wx.adv.NotificationMessage` in the event that the O/S doesn't have this functionality built in. (If running in console mode and no O/S notification, then console message would need to suffice.)

Sophist-UK:
There is a Pypi package for Win10Toasts but this is just Win10 and we probably don't want to have another Pypi dependency - but the code is not complicated and is MIT license, so we can almost certainly copy and tweak the code.

I'll look into creating a tools plugin to provide this functionality as a command.

Sophist-UK:
The only way I can find to produce Windows 10 notifications from Python is [Win10Toasts](https://github.com/jithurjacob/Windows-10-Toast-Notifications) which in turn relies on [pywin32](https://github.com/mhammond/pywin32). In theory it should be possible to work out how the former uses the latter and how the latter does ctype calls to windows dlls, and thus eliminate these dependencies, but frankly this is a lot of work. What would the issues be for adding these two dependencies to requirements.txt?

Sophist-UK:
Ok - it looks like `wx.adv.NotificationMessage` actually interfaces with Windows and Mac notification subsystems, so the best choice is to use this if we have GUI availability, and notify to console if not. Working on this now.

tatarize:
I am leaving this targeting main, even after the legacy7 switch. The commands here should probably be both put into extras or dealt with properly. Turning `beep` and `os_notify` into proper commands.

Sophist-UK:
I did try to create a PR which put beep and notify into a separate plugin (in 0.7) but the difficulty is in making an o/s agnostic Notify - you cannot create a Windows o/s notify box without having a shortcut with certain parameters, and that means using an installer to install the executable and create a shortcut - and that means it wouldn't work running from source etc.

So the only way to be platform agnostic IMO is to crate our own notification windows that pop up and then disappear after so many seconds and which stack vertically like o/s notifications do - and this was too much work for the time I had available at the time.

tatarize:
Probably won't get merged.

PR #699, opened 2021-11-29
Provide Help/Check for updates and startup checks

Sophist-UK:
I am extremely happy with the functionality of this code. Here are a few highlights:

1. Preferences for Startup autocheck (Enabled by default), Check for betas (as before - disabled by default), Always download with browser (disables built in executable download functionality and opens browser window in all cases). If Check Betas is disabled, but user is temporarily running a beta (e.g. to address a specific issue they might have), then it looks for a later beta in the same minor version or a full version.

2. GUI menu Help/Check for updates - if running from pkg it gives example instructions to update, if running from src or exe it offers a download button (running from exe and exe directory is writeable, defaults to internal downloader which has a progress bar, or browser download).

3. Startup autocheck in console mode - runs every time MK starts in console only mode.

4. Startup autocheck in GUI mode - like GUI check, but errors and up-to-date messages are suppressed. Has a few extras - we developers can specify minimum length of time (typically days) between checks (default at present 1 day), and user can defer for a set number of days (default at present to 7 days) before being reminded again, but perhaps these should both be longer (7 and 14) in order to stagger the uptake over a longer period. Also there is a minimum period after publication of a release before it is shown (currently 2 days for beta and 7 days for full). But if user hears of the release, a manual check will always find them. 

5. Internal downloader for executables - if not already provided adds the tag to the filename so that it is unique because (AFAIK and on Windows almost certainly) you cannot overwrite the currently running executable. We might want an option for this. Also I have just thought that we could potentially do an internal download of a tar/zip and unpack it if user is running from source as I do not believe that python holds the source files open after loading them.

6. Test mode - setting only available from console (updater_test) allows you to set test mode whereby GUI always behaves like a startup check, current version is 0.0.0 (so latest versions are always found for update) and all the timescales defined above in 4. are reduced close to zero so you can run the autocheck multiple times.

7. Debugging details to a channel (channel open updater).

Technically - web activities (getting release details from Github and downloading) are done on a worker thread. All GUI windows are done on main thread.


Sophist-UK:
I am providing this for review because the code is fully working and tested. I still have four things I need to do before this is ready to merge:

1. Get uname details for Pi (32 and 64) and Mac (non M1 32/64 and M1) so I can define default filename specs.
2. Update the time periods for startup autochecks (as described above)
3. Add a published date to the messages advising of a new release available (as this would I think be useful for users).
4. Write unittests for as much of the internal functionality as makes sense.

But please have a play with this and give feedback.,

meerk40t/gui/mwindow.py
```diff
@@ -35,10 +35,10 @@ def __init__(self, width, height, context, path, parent, *args, style=-1, **kwds
...
             self.window_context.setting(int, "width", width)
             self.window_context.setting(int, "height", height)
-            if self.window_context.width < 100:
-                self.window_context.width = 100
-            if self.window_context.height < 100:
-                self.window_context.height = 100
+            if self.window_context.width < width:
+                self.window_context.width = width
+            if self.window_context.height < height:
+                self.window_context.height = height
```
tatarize, line 41: This would force the windows minimum size to being equal to width/height. That's not correct. If this were the intent a second width/height for min_width, min_height should be passed to this function call.

meerk40t/main.py
```diff
@@ -24,7 +25,14 @@
...
 if not getattr(sys, "frozen", False):
-    APPLICATION_VERSION += " src"
+    # If .git directory does not exist we are running from a package like pypi
+    # Otherwise we are running from source
+    if os.path.isdir(sys.path[0] + "/.git"):
+        APPLICATION_VERSION += " git"
+    elif os.path.isdir(sys.path[0] + "/.github"):
+        APPLICATION_VERSION += " src"
+    else:
+        APPLICATION_VERSION += " pkg"
```
tatarize, line 35: It doesn't seem like pkg is useful for mato folks or .git. And I'm not sure those folders existing mean this stuff.

Sophist-UK: These suffixes were introduced so that the bug reports had information about the type of execution. 

I have checked the Pypi distributions and the Github downloads and this does seem to be a valid way to distinguish between Git, Pypi and Github download executions.

meerk40t/gui/mwindow.py
Sophist-UK: The width and height provided in the __init__ call are the initial sizes of the window. But if you increase these in a PR, the window does not increase in size unless the user manually resets all windows position/size settings. In particular new Preference checkboxes are hidden.

The code already ensures a minimum size on opening if the user has shrunk it to something smaller. All this change does is to change the minimum size on open to something more meaningful without preventing the user from then reducing the size manually if they wish to.

The alternative would be to store the previous default size and compare it to when the window is opened again, and if the default size has changed, to resize it only in that case.

Sophist-UK:
Currently working in a revamp of some of the detail of this to deal with updates in place on a better basis.

tatarize:
Fixed the merging, though updater.py still has some obvious compile errors and isn't in a usable state.

Sophist-UK:
I haven't forgotten this - but it needs some more thought about how exactly to handle updates for those that run from source and those that run executable, with some checks for sufficient disk-space etc. So I am waiting until I have some time without any other distractions to work on this.

PR #703, opened 2021-12-11
Improve crash reporting...

Sophist-UK:
... and move `bootstrap("ready")` indicating ready for user input to after command line console commands have been executed.

### DIFF
--- a/meerk40t/main.py
+++ b/meerk40t/main.py
@@ -340,6 +340,8 @@ def run():
             except IndexError:
                 break
 
+    kernel.bootstrap("ready")
+
     def __print_delegate(*args, **kwargs):
         if print not in console.watchers:
             print(*args, **kwargs)
@@ -363,7 +365,7 @@ def run():
         if unprint_console:
             console.unwatch(__print_delegate)
 
-    kernel.bootstrap("ready")
+    kernel.bootstrap("finished")
 
     if args.auto:
         # Auto start does the planning and spooling of the data.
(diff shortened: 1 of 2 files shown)


meerk40t/main.py
```diff
@@ -349,6 +355,8 @@ def run():
                 kernel_root(line.strip() + "\n")
         kernel_root.channel("console").unwatch(print)
 
+    kernel.bootstrap("ready")
```
tatarize, line 358: This has functional impact. The ready lifecycle event in 0.7.x version does a flag for "on_controller_ready" for the lhystudios devices this causes the controller thread in the lhystudios driver and the moshiboard driver to kick in. This is also when the start driver for the drivers kicks in which does driver threading. This is where it starts the fetch/execute loop for the spooler. and this also is when the tree is initially built.

The consequence here is that batch and execute will happen before these events.  I believe this ordering is because "quit" is a valid command and was a common thing for -q for a while. This might be an acceptable ordering but the `quit` setting would typically work when the fetch execute loop finds an empty spooler. In this ordering that really could happen even with `execute` and `batch` commands.

At a minimum this change should *not* be part of improving crash reporting and should be justified in an independent PR. (also the change would be obsoleted by 0.8.0).

tatarize (changes requested):
This lifecycle change seems like it could spawn problems.

meerk40t/main.py
Sophist-UK: Ok - I can put this back and add a new `bootstrap("finished")` event for any plugins that want to run stuff after boot has finished.

tatarize (approved):

PR #708, opened 2021-12-15
NSE scanline switching / twitchless mode

tatarize:
This is mostly for testing unless there's a better reason to add it than it exists.

The core of this is understanding that G001 mode turns the switch of right moving raster or left moving raster into what is apparently the same as `N<step>SE` so if we're going right (B) we can switch, turn off the laser, and reverse with the command `T` which means turn right. The G001 flagged the scanline step at 1.

With NSE scanline switching we can simply replace the `T` in that case with `NLcSE` assuming we're stepping towards the top (L) and our step size for that step is 3 (c). We are going to be wasting 4 characters each scanline, however this type of scanline can be set dynamically. So we could do interlace rastering. Or raster with a variable step such as a 3,3,4 pattern which would give you a DPI of about 300 rather than 250 or 333. This would require pixels also have a 3,3,4 width-wise. We can also do more fine grain control of rastering. I at one time thought G001G002 would do variable stepping on each side, but it doesn't work. G000 might as well just be a flag that means perform unidirectional rastering. With NSE stepping we can more effectively perform a scanline step effectively.

Also, since this is a vector mode option, it should also work for things like lightburn ruida emulation which suffered serious problems. Since the options there included DPI of values like 225 or whatever, which aren't able to be achieved with a set step the current implementation is a hard-raster. It's vector mode raster implementation which often clicks at the edges and needs to run much more slowly. Here, if we could interpret the commands correctly and perform the relative movements correctly we could implement raster mode in a much more vector-mode friendly way. And we could even avoid ever calling vector mode anyway.

This means that the only real need to call `@NSE` or kick all the way back to real rapid mode is that we will sometimes need to set the major and minor axis. This is to say `NSE` and `other-direction-of-a-raster` perform a step along one specific and explicitly given axis. And there appears to be no way to set this axis elsehow. Which means we can change the speed and accel value and other values on the fly. This means the N<stepping-axis><initial-direction>S1E command that seems to be done in real raster mode is the only reason we'd be forced back to full raster mode.

Whichever value we set there is highly important since if we do an NSE command switch we end up performing a raster-endstop step. This means we do the raster-twitch in the last direction set. And the twitching works differently based on the step axis setting. If we could *also* change this more easily, we could likely get around some of the last remaining hurdles to getting rid of twitching.

### DIFF
--- a/meerk40t/core/cutcode.py
+++ b/meerk40t/core/cutcode.py
@@ -98,9 +98,22 @@ class LaserSettings:
         for q in dir(obj):
             if q.startswith("_") or q.startswith("implicit"):
                 continue
+            obj_type = type(obj)
+            if hasattr(obj_type, q) and isinstance(getattr(obj_type, q), property):
+                # Do not set property values
+                continue
+
             value = getattr(obj, q)
             if isinstance(value, (int, float, bool, str)):
                 setattr(self, q, value)
+
+    @property
+    def horizontal_raster(self):
+        return self.raster_step and (self.raster_direction == 0 or self.raster_direction == 1)
+
+    @property
+    def vertical_raster(self):
+        return self.raster_step and (self.raster_direction == 2 or self.raster_direction == 3)
 
     @property
     def implicit_accel(self):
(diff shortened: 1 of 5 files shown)


Sophist-UK (commented):
I do not run from source on my production pc, so without an executable artefact for this PR I cannot test this.

joerlane (approved):
This is a wonderful PR. Thanks.

PR #702, opened 2021-12-11
Standardise on platform.system for determining O/S

Sophist-UK:
For code consistency standardise on platform.system as a means of determining the operating system.

PR #709, opened 2021-12-17
Fix #702 and switch to os.path

Sophist-UK:
1. Fix bug from #702
2. Remove single reference to pathlib for consistency and replace with equivalent os.path functions.
3. Add support for unit testing all O/S'

### DIFF
--- a/test/test_kernel.py
+++ b/test/test_kernel.py
@@ -1,15 +1,10 @@
-
 import unittest
-
 from test import bootstrap
-
 
 class TestKernel(unittest.TestCase):
     def test_kernel_commands(self):
         """
         Tests all commands with no arguments to test for crashes...
-
-        :return:
         """
         kernel = bootstrap.bootstrap()
         try:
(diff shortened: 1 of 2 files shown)


Sophist-UK:
I had a crash on windows running from executable when I tried "channel save 0/usb_send" where it was trying to do something with the pyinstaller temp directory and a zip file - however I failed to save any details of this crash anywhere and cannot recreate it.

I cannot see how it could be a crash anywhere except when attempting to create a subdirectory in a zip file.

So I have made the other changes I felt were useful and if the crash recurs I will submit another PR.

test/test_kernel.py
```diff
@@ -26,3 +21,34 @@ def test_kernel_commands(self):
...
+                + sep + "test"
+            )
+        )
+        self.assertEquals(
+            str(get_safe_path("test", system="Linux")),
+            (
+                os.path.expanduser("~")
+                + sep + ".config" + sep + "test"
+            )
+        )
```
tatarize, line 53: I'm not sure this would work. Since it would only apply to one OS.

tatarize (approved):

test/test_kernel.py
Sophist-UK: This tests all o/s cases (except something unknown).

PR #712, opened 2021-12-22
Improve scene zoom menu

Sophist-UK:
1. Add Zoom to Bed = Zoom to Size with nothing selected.
2. Rename Zoom-to-Size to Zoom-to-Selected
3. Make Zoom-to-Bed be closer fit to scene area (104% rather than 110%)
4. Make Zoom-to-Selected have same fill factor (104%)


tiger12506 (approved):
Tested. Does what it says on the tin.

tatarize (approved):
Would totally do those things it says it does.

PR #718, opened 2021-12-27
Create references to kernel and root-context and translator as standard

Sophist-UK:


meerk40t/kernel.py
```diff
@@ -46,6 +46,9 @@ def __init__(self, context: "Context", name: str = None, channel: "Channel" = No
         self.context = context
         self.name = name
         self.state = STATE_INITIALIZE
+        self.kernel = self.context._kernel
+        self.root = self.kernel.root
```
tatarize, line 50: I'm pretty sure these exist as like property values or something. Like context.kernel is used a lot without the _kernel. It just does a property reference rather than an attribute like that.

tatarize:
```python

    @property
    def kernel(self) -> "Kernel":
        return self._kernel

```

It might be a 0.8.0 change I'll double check.

tatarize:
https://github.com/meerk40t/meerk40t/blob/0ed078ede3008f2026cb9b329aacd42cf680bd0a/meerk40t/kernel.py#L193

Nah, this stuff exists already.

Sophist-UK:
> ```python
>     @property
>     def kernel(self) -> "Kernel":
>         return self._kernel
> ```
> 
> It might be a 0.8.0 change I'll double check.

> https://github.com/meerk40t/meerk40t/blob/0ed078ede3008f2026cb9b329aacd42cf680bd0a/meerk40t/kernel.py#L193
> 
> Nah, this stuff exists already.

Neither of these are part of this PR - check the Files Changed tab.

meerk40t/kernel.py
Sophist-UK: These are just conveniences. So you can say `self.root instead of `self.context._kernel.root`.

tatarize:
My point is those already exist. They aren't attributes exactly they are property attributes. But context already has access to .root and .kernel and they work fine.

tatarize:
The link there was part of context already: https://github.com/meerk40t/meerk40t/blob/0ed078ede3008f2026cb9b329aacd42cf680bd0a/meerk40t/kernel.py#L125

you can already do `context.root` and `context.kernel` and I think they are used all over the place.

Sophist-UK:
> you can already do `context.root` and `context.kernel` and I think they are used all over the place.

Ok.

PR #720, opened 2021-12-28
Modify operation speed command to set speeds by percent.

tatarize:
Modifies the operation speed command to allow, `operation* speed -d +2%` or `operation* speed -1`.


tiger12506 (approved):
Tested, works fine, and solves the original issue.

PR #722, opened 2021-12-28
Improve classification edge cases

Sophist-UK:
1. Debug channel with summary of major actions / decisions
2. Edge case 1: Non-grey stroke, white fill at back i.e. fill not on top of anything else - now classifies as vector rather than raster
3. Edge case 2: No fill path same colour as overlapping raster elements of same colour

Note: This is a PR against the 0.8.0 services branch as it should probably have a decent beta period.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -6095,32 +6095,62 @@ class Elemental(Service):
     def classify(self, elements, operations=None, add_op_function=None):
         """
         Classify does the placement of elements within operations.
+        In the future, we expect to be able to save and reload the mapping of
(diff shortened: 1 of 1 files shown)


Sophist-UK:
[PR722 test cases.zip](https://github.com/meerk40t/meerk40t/files/7786548/PR722.test.cases.zip)

Sophist-UK:
NOTE: the test cases do highlight an existing issue - namely that the scene faithfully displays the Elements tree but when you have vector operations behind elements with white fill, these are not visible. The Simulation display does accurately show what is burned.

This PR does not attempt to address this issue - if indeed the issue can or should be addressed.

PR #725, opened 2022-01-02
Submit crash reports if running from git/main

Sophist-UK:


Sophist-UK:
This is good to go.

tatarize (approved):
At a certain point this stuff might end up being overbuilt. Looks fairly fine. Not sure about shoving actual branch information into the code though it won't activate unless actually part of a git install.

PR #727, opened 2022-01-02
Improve special ops handling

Sophist-UK:
1. Improves the Add Ops / Special Ops context menus: Can now add special ops after specific LaserOperations as well as append them to the end. Also add Home/Beep/Interrupt in one go as a commonly used operation to give access to the material part way through.
2. Now saves and reloads Special Operations when you Save/Use operations and when you close and restart MK.

However, during testing I found a couple of oddities:
a. Attribute "type" is not always saved when you persist operations - no idea why. I have coded around this - and it may be necessary to avoid point b. when user upgrades to a version with this PR.
b. If persistent settings are invalid, then MK can crash before starting - this really needs to be more robust esp. on load. I will try to add this before it gets merged if I can find a way to do it nicely.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -5606,7 +5606,10 @@ class Elemental(Modifier):
 
     def boot(self, *a, **kwargs):
         self.context.setting(bool, "operation_default_empty", True)
-        self.load_persistent_operations("previous")
+        try:
+            self.load_persistent_operations("previous")
+        except ValueError:
+            print("elements: Previous operation settings invalid: ValueError")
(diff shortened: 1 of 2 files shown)


Sophist-UK:
As far as I can tell, this should now be reasonably robust - however I am not convinced that I have found all possible issues created by adding "cmdop" (as well as "op") nodes to `elemental.ops()`.

Sophist-UK:
P.S. This PR does not save and load Special Operations in MK SVG saves. That is for a later day.

tatarize (approved):
All looks to be fairly reasonable.

PR #729, opened 2022-01-03
Initial svg xml extension definitions

Sophist-UK:
This PR provides new svg xml extension namespace definitions for a proposed rewrite of the SVG save and load functions which will support:

1. Saving and reloading special operations and
2. Storing and restoring the associations between elements and operations

I am submitting this PR early and separately so that these extensions can be reviewed and approved before the code is written.

I plan to write the full set of svg load and save enhancements in two stages as separate PRs to this:

1. A PR against 0.7.x adding support for groups and SVG shapes (which are currently converted to Paths on save). Because svgio.py already supports loading these tags, this should be fully backwards compatible.
2. A PR against 0.8.x (once the above has been merged also with the 0.8.0 branch) to add the support for the new tags. This needs to be made against 0.8.0 since files written with these tags will not be loadable by 0.7.x versions of MK.

### DIFF
--- a/meerk40t/camera/gui/camerapanel.py
+++ b/meerk40t/camera/gui/camerapanel.py
@@ -487,7 +487,7 @@ class CamInterfaceWidget(Widget):
             )
 
             camera_context = self.cam.context.get_context("camera")
-            keylist = camera_context.kernel.load_persistent_string_dict(
+            keylist = camera_context.kernel.read_persistent_string_dict(
                 camera_context.path, suffix=True
             )
             if keylist is not None:
@@ -785,7 +785,7 @@ class CameraURIPanel(wx.Panel):
 
     def pane_show(self):
         camera_context = self.context.get_context("camera")
-        keylist = camera_context.kernel.load_persistent_string_dict(
+        keylist = camera_context.kernel.read_persistent_string_dict(
             camera_context.path, suffix=True
         )
         if keylist is not None:
--- a/meerk40t/camera/gui/camerapanel.py
+++ b/meerk40t/camera/gui/camerapanel.py
@@ -805,8 +805,7 @@ class CameraURIPanel(wx.Panel):
             if not c.startswith("uri"):
                 continue
             setattr(camera_context, c, None)
-        for c in list(camera_context.kernel.keylist(camera_context.path)):
-            camera_context.kernel.delete_persistent(c)
+        camera_context.clear_persistent()
 
         for i, uri in enumerate(self.uri_list):
             key = "uri%d" % i
--- a/meerk40t/camera/gui/camerapanel.py
+++ b/meerk40t/camera/gui/camerapanel.py
@@ -166,10 +166,10 @@ class CameraPanel(wx.Panel, Job):
             CamInterfaceWidget(self.widget_scene, self)
         )
 
-    def pane_show(self, *args):
-        from sys import platform as _platform
+    def initialize(self, *args):
+        from platform import system as _sys
 
-        if _platform == "darwin" and not hasattr(self.camera, "_first"):
+        if _sys() == "Darwin" and not hasattr(self.camera, "_first"):
             self.context("camera%d start -t 1\n" % self.index)
             self.camera._first = False
         else:
@@ -645,9 +645,9 @@ class CameraInterface(MWindow):
         # ==========
         # MENU BAR
         # ==========
-        from sys import platform as _platform
+        from platform import system as _sys
 
-        if _platform != "darwin":
+        if _sys() != "Darwin":
             self.CameraInterface_menubar = wx.MenuBar()
             self.create_menu(self.CameraInterface_menubar.Append)
             self.SetMenuBar(self.CameraInterface_menubar)
--- a/meerk40t/camera/gui/camerapanel.py
+++ b/meerk40t/camera/gui/camerapanel.py
@@ -491,7 +491,7 @@ class CamInterfaceWidget(Widget):
                 camera_context.path, suffix=True
             )
             if keylist is not None:
-                keys = [q for q in keylist]
+                keys = list(keylist)
                 keys.sort()
                 uri_list = [keylist[k] for k in keys]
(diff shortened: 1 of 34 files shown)


Sophist-UK:
ToDo:
1. Check Operations attributes which need to be saved
2. Add DTD definitions.

tatarize:
Proper saving would also save the reference nodes (opnode) values within the operations. Largely this is required to avoid forced classification. So you can load and save what you had before.

Sophist-UK:
I am suggesting doing this the other way around - rather than save a list of the elements against each operation, I am proposing to store a list of operations against each element. There is a reason for this - that elements don't always have an id, whereas operations always have an index. (Its a bit more complicated when you load the file and want to use an existing operation with an existing index - you just need to maintain a mapping between the index of the `<op/cmdop>` in the file to an actual operation in the tree.

tatarize:
Given the galvo project thing I've been batting around the idea of pens. They are somewhat common in other laser software. And while the operations are 1:many rather than 1:1 like most pen implementations there are some things like hatch in galvo has like 3 different pens which basically makes the hatch operation a glorified set of three pens. There is therefore another reason to consider going with operation index matching for the references and that's because you don't necessarily need to save your operations. I mean clearly you should and that makes sense but if you have elements that say they belong in operation 3 and 7, then you really could just go ahead and put that into operations 3 and 7.

Now clearly there's some obvious non-parallel things with list of operations vs. list of pens like special operations or data-blobs that really make a lot of sense, but you could also do something like mark default operations as if they are a pen, you could assign operations to them with a yet not defined UI.  But, then you could treat your operation list like a pen list and maybe even give a weaker than TREE version that just hides the details from the user where you get a bunch of pens and the elements can be added to these. And if they contain anything they are called an operation and put in the operations code or could be put into operations while empty. But, you could also do things like have the drawing tools add an item into a particular operation when they are created. So select a default-operation for this thing you're drawing and we'll just add it to the operations.

It's a bit sloppy, but I went with the most useful methodology there, but there's clearly a lot to be gained by emulating the way other software works. In this case you'd be able to assign  an element to many different operations or to no operation at all. And if that could be made easier it would reduce the need for classification and let users more easily what their lasers should do. But, also would help to fit into other methodologies. I fully understand the idea of a jack of all trades, master of none sort of thing. But, if somebody loves how whisperer did things it really wouldn't be hard at all to let the give the user a little box with speeds and values that say engrave, cut, and raster. And when they hit the button it searches all the elements and finds anything that is exactly blue stroke and executes an engrave operation on it, by taking that and throwing it into the current device's spooler.

As for 1: Operations attributes are a little unknowable. The extra pages of settings in Operation Properties could well include settings that don't matter to K40 users but could matter a lot to other projects and development. So they could have things like frequency, pwm, wobble, split size, z-bed ticks per pass, angle, shimmy, number_of_unicorns. -- You really just can't know what operations are or could be required. We could if they need to ensure a unique namespace be prefixed as meerk40t:wobble or whatever as needed, but for the most part `LaserSettings()` as a class and on the whole is going to need to be basically a dictionary of you will not know what. This ambiguity will need to be attached correctly to the cutobjects and to Operations that create cutobjects, and most things that might create cutobjects. Backends are absolutely free to ignore any and all settings they do not support and use their backup global settings or the settings associated with the cutobject. The first page of operations properties is probably going to be universal though, basically power, speed, passes, and what operation you're performing.  But, even that might be assuming too much.

Sophist-UK:
Once again I am having trouble understanding what you are saying. Are you suggesting that Pens are some sort of intermediate mapping from elements to operations? Can you point me to an explanation in help pages for other laser software that would help?

tatarize:
I'm saying that pens are a mapping of 1 to many. You have 1 pen. It can have multiple elements. You have multiple elements they have exactly 1 potential pen. Operations and elements are different they are any number of elements for any number of operations. --- So if we wanted we could implement a similar system to other software as such.

There's bound to be some videos on them, but a lot of laser software uses pens.

Sophist-UK:
So a pen can have 0-n elements and be assigned to 0-n operations? It's effectively a grouping of elements?

tatarize:
Elements can be assigned to 1 and only 1 pen. A pen can have any number of elements. It's effectively an operation without the ability to put the element as relevant to more than one pen. These are all predefined and little color swatches in a toolbar (usually bottom). Then you have your settings programmed into those and they remain the same throughout. When you draw an item you select the pen to use and draw the element and that element classifies into that pen. Then if you select an item and you pick a different pen you classify that element into that operation which also removes it from the previous operation. So if you want to fill something two different ways you need to make a copy of it. You could pick up a free version of rdworks and check how that works.

Sophist-UK:
So - it's like a colour but without giving the user the flexibility to e.g. classify several cyan elements into the same two matching operations and then manually move one of them to a third operation of a different colour? 😉

tatarize:
Yeah, but the interesting method of having them pick like color swatches and each does a different thing.

tatarize:
rd-works

Note the colors in the operation list and color in the bottom there. You have those you select them and assign a element a color

Sophist-UK:
Ok - so it's a colour plus an operation type combined - so not only can you not put elements of the same colour into different operations, you also can't put elements of the same colour into different types of operation. So it's more restrictive than I first thought.

I think this would not be great when you import an SVG file you created in e.g. Inkscape and then found that it wasn't quite right and you needed to tweak things in MK.

BUT...

I can see the benefit if we create a whole new part of MK focused on drawing and creation - it might well make sense there to enforce a consistent use of colours.

tatarize:
Yeah, or just give them the option to use something like that. What it loses in general power it does make up for being somewhat user-friendly. And we currently have operations that we mark as "default" what if "default" like operations were sort of pens. I mean you could give them a friendly button bar and when you draw something you make it that color and you put it in that operation. There's already some drawing tools and I really am at the point that I could make literally any number of other really awesome and easy drawing tools.

tatarize:
nothing came of this, full save is mostly done in 0.8.x

PR #730, opened 2022-01-04
Ruidacontrol Rastering

tatarize:
Detect and process rasters for the ruida control emulator.
---
I had to go through and simplify the driver code a bunch to make sure I could figure out how to correct the main issue. These corrections removed a bunch of legacy code and made proper conversions of similar logic functions.

Codewise this *should* work but because it changes the core of how the driver works. This needs testing for:

- [x] vector cuts in all directions.
- [x] rasters
- [x] twitch and twitchless vectors.
- [x] alt raster
- [x] ruidacontrol interfacing
- [x] turning off alt-raster while rastering
- [x] turning on alt-raster while rastering

The ruidacontrol interfacing seemed pretty solid, it let me raster a 1 degree rotated image without going insane and raster normal rasters at solid speed, with the ruidacontrol code.


tatarize:
I guess by popular demand I should push this code back to the front burner. It wasn't too far from working.

Sophist-UK:
👍

tatarize:
This should be an improvement but is not apparently finished. There's a couple seemingly okay modes that fail horribly like trying to raster something at an angle. It goes insane. It actually should be able to be made to work.

If that's sort of required we could push this back a bit. Also, I think there might have been an error at speed that caused a hit. I'll see if there's something notably quick to fix this stuff like forcing a direction shift for the h_switch which would make sense if the direction was somehow wrong.

Sophist-UK:
I would suggest that we tidy up what we have without angled rastering and (ideally with my rastering optimisation finished - I just need to analyse and handle the accelerations) push out another beta or even full release.

tatarize:
I think this might still have a tiny break at the very start point where it goes the wrong way. I tried. forcing the directionality but it seemed like it was still wrong. I think maybe directional angle rastering might be somewhat required since getting things right like that tend to have knock on effects. And the start pip could be something like that.

If it moves at an angle, but does not change the x direction. So we to `__/--` it wrongly inverts that because it expects things to only go `==|`  So I think the same problem here might be manifesting in a couple different ways.

tatarize:
The issue is that it's detecting a raster and thinking because it didn't move up or down by more than 15 and 1 direction of the cardinal directions wasn't seen that that is a raster. But, it fails to account for movements up while moving in a lateral direction.

I tried super-alt-rasters where I do not do any y-changes in a flip around. I just process the y-change whereever I hit one, but I could hear it move where it was a lot softer inside the N-SE switch. But, I think I might just bite that bullet and do it that way anyway since unlike this it would be able to do laterial shifts without inverts.

tatarize:
I couldn't solve it, got mostly grinding out of my laser. It should be plausible not sure what I did wrong. Might need to run some more tests with that. This is certainly an improvement but the starting up hit should be fixed and maybe things in the driver need some rewriting still. I'll try again tomorrow (much later today). 16 hours from now or so. But, this is like 95% of the way there. Turns out +15 is enough to cause most angles to count as non-rasters. Since they tend to go up by more than that over the distance of the image.

Even tiny images at low angles didn't work. Caused some unknown error where it would try to force raster the same direction twice. The enforce thing might have fixed the first hit.

This stuff could certainly be tested a bit, but I think it's still a little wrong and I'd sort of prefer isolating the direction swaps in this thing.

Sophist-UK:
yes - what is commonly used needs to work fully - rare things less so.

meerk40t/device/lhystudios/lhystudiosdevice.py
```diff
@@ -1526,203 +1446,114 @@ def unlock_rail(self, abort=False):
     def abort(self):
         self.data_output(b"I\n")
 
-    def check_bounds(self):
-        self.min_x = min(self.min_x, self.current_x)
-        self.min_y = min(self.min_y, self.current_y)
-        self.max_x = max(self.max_x, self.current_x)
-        self.max_y = max(self.max_y, self.current_y)
```
tatarize, line 1533: All calls to check bounds were removed.

meerk40t/device/lhystudios/lhystudiosdevice.py
```diff
@@ -1526,203 +1446,114 @@ def unlock_rail(self, abort=False):
...
+        self._request_leftward = None
+        self._request_topward = None
+        self._request_horizontal_major = None
+        self._topward = False
+        self._leftward = False
+        self._x_engaged = False
+        self._y_engaged = False
+        self._horizontal_major = False
+
+    def goto_xy(self, dx, dy):
```
tatarize, line 1460: goto_x and goto_y were inlined into goto_xy

meerk40t/device/lhystudios/lhystudiosdevice.py
```diff
@@ -1526,203 +1446,114 @@ def unlock_rail(self, abort=False):
...
-            self.move_bottom(dy)
+    def goto_octent(self, dx, dy, on):
+        if dx == 0 and dy == 0:
+            return
+        if on:
+            self.laser_on()
         else:
-            self.move_top(dy)
-
-    def goto_angle(self, dx, dy):
```
tatarize, line 1551: goto_angle was inlined into goto_octent

meerk40t/device/lhystudios/lhystudiosdevice.py
```diff
@@ -1526,203 +1446,114 @@ def unlock_rail(self, abort=False):
...
-        )
+        return not self._x_engaged and self._y_engaged and not self._topward
 
     @property
     def is_angle(self):
-        return self.is_prop(STATE_Y_STEPPER_ENABLE) and self.is_prop(
-            STATE_X_STEPPER_ENABLE
-        )
-
-    def set_left(self):
```
tatarize, line 1659: set directional elements were inlined

meerk40t/device/ruida/ruidadevice.py
```diff
@@ -173,6 +173,7 @@ def __init__(self, context, path):
         self.filestream = None
 
         self.cutcode = CutCode()
+        self.plotcut = PlotCut()
```
tatarize, line 176: All cutcode entries were merged into plotcode calls. Plotcut is a cutcode element that tracks specific locations and can do calculations like whether this code is a raster or not.

meerk40t/device/ruida/ruidadevice.py
```diff
@@ -1133,6 +1115,11 @@ def process(self, array):
                 # Only seen in Absolute Coords. MachineZero is Ref2 but does not Set Absolute.
         elif array[0] == 0xE7:
             if array[1] == 0x00:
+                if len(self.plotcut):
+                    self.plotcut.settings = self.cutset
+                    self.plotcut.check_if_rasterable()
+                    self.cutcode.append(self.plotcut)
+                    self.plotcut = PlotCut()
```
tatarize, line 1122: Plotcuts are defined at the blockend. Since this is ends the ruida block and different cut objects should have different blocks and thus different plotcuts.

Sophist-UK (approved):
I don't have the ability to test this or the knowledge to know whether it is doing the right thing but the code looks OK.

I recommend that we spit out a full 0.7.5 release before we merge this, and then with the raster optimisations PR spit out another 0.7.6 beta 1.

Sophist-UK:
On reflection, I think this should be merged **before** we spit out a full release.

tatarize:
Okay.

PR #736, opened 2022-01-06
Fix issue #724

### LINKED ISSUE #724: [Crash] SetAttr error in svgio.py when loading previous save
```
MeerK40t crash log. Version: 0.7.5-beta1 git on Windows:AMD64 - 4.1.1 msw (phoenix) wxWidgets 3.1.5
  File "D:\MeerK40t-64\Source\meerk40t_meerk40t\meerk40t\gui\wxmmain.py", line 1663, in <lambda>
    lambda e: self.load_or_open(context.file0),
  File "D:\MeerK40t-64\Source\meerk40t_meerk40t\meerk40t\gui\wxmmain.py", line 1624, in load_or_open
    self.load(filename)
  File "D:\MeerK40t-64\Source\meerk40t_meerk40t\meerk40t\gui\wxmmain.py", line 1762, in load
    results = self.context.load(
  File "D:\MeerK40t-64\Source\meerk40t_meerk40t\meerk40t\core\elements.py", line 6575, in load
    results = loader.load(self.context, self, pathname, **kwargs)
  File "D:\MeerK40t-64\Source\meerk40t_meerk40t\meerk40t\core\svg_io.py", line 292, in load
    result = SVGLoader.parse(
  File "D:\MeerK40t-64\Source\meerk40t_meerk40t\meerk40t\core\svg_io.py", line 410, in parse
    setattr(
AttributeError: can't set attribute
```

tatarize:
Vertical_raster is saved in the test file but setting vertical_raster is not possible because it's a @property. To rectify we go ahead and except the attribute error for the setattr. there.

Fixes #724

### DIFF
--- a/meerk40t/core/svg_io.py
+++ b/meerk40t/core/svg_io.py
@@ -404,15 +404,18 @@ class SVGLoader:
                                             key,
                                             type_v(element.values[key]),
                                         )
-                                    except (ValueError, KeyError):
+                                    except (ValueError, KeyError, AttributeError):
                                         pass
                                 elif type_v == bool:
-                                    setattr(
-                                        op.settings,
-                                        key,
-                                        str(element.values[key]).lower()
-                                        in ("true", "1"),
-                                    )
+                                    try:
+                                        setattr(
+                                            op.settings,
+                                            key,
+                                            str(element.values[key]).lower()
+                                            in ("true", "1"),
+                                        )
+                                    except (ValueError, KeyError, AttributeError):
+                                        pass
                         elements_modifier.add_op(op)
                 except KeyError:
                     pass


Sophist-UK:
This file did not create an error in 0.7.5b1. I am unclear why it started to crash nor why ignoring the error is a fix.

tatarize:
With the code as it was, it would always be the case that setting the attr based on the svg was going to be a problem if it tried setting a function. This will prevent that.

Sophist-UK:
Yes - but why was it trying to do it in the first place - you have cured the symptom not addressed the root of the problem.
1. Why did it work in 0,7.5b1?
2. If it was something in the file, how was that broken file written?

tatarize:
1. Because it didn't have any properties in 0.7.5b1
2. The file is written because everything in the object is written to disk. And the attribute there looks to be a regular boolean and so it gets written but, it's a property and writing it both pointless and problematic because you can't read it and set the property. If one wanted to fix this fully the svg writing routines would be best to avoid anything that is a property. The code for loading and saving Settings in kernel is exactly that sort of thing.

Sophist-UK:
I will fix the save when I do the File Save rewrite.

Sophist-UK (approved):

PR #754, opened 2022-01-20
Reduce travel time 'on' by default

jaredly:
It can save so much time, and it's likely the default most people want.

lmk if I should have gone about this some other way :)


tatarize (approved):
That's accurate.

Issue #753, opened 2022-01-20
Turn on "reduce travel time" and "cut inner first" by default?


jaredly:
Are there important reasons that they're currently off-by-default? I feel like they're what most users expect / want.

reduce travel time turned off (the current default)
![image](https://user-images.githubusercontent.com/112170/150379195-44b26517-8c3f-4911-a639-b5ae27eab50f.png)

tuned on, 🏎🏎🏎🏎
![image](https://user-images.githubusercontent.com/112170/150379222-63969e34-19d7-42ea-b1a9-73d5fb6a7698.png)

tatarize:
Was previously much slower. I think that was most of the reason. Sometimes you could look down on like a couple minutes.

jaredly:
thanks!

Sophist-UK:
Welcome to the Meerk40t Contributors Hall of Fame!!!!!!

PR #755, opened 2022-01-20
Additional optimisation options

Sophist-UK:
1. Burn Complete Subpaths - with this option, closed subpaths can be entered anywhere, but non-closed subpaths will always be burned from one of the ends.
2. Group Inner Burns - when the first inner element has been burned, the rest of the matching inner burns will be burned next, followed by the outer path before moving on to other elements. For a design with multiple parts, this will normally burn each part in turn rather than e.g. doing inner burns for one part then outer burns for another then returning for outer burns for the first. This reduces the risk of a shift ruining all parts by completing one part before moving onto another.

Additionally, minor fixes for:
a. Repeat step and multiple passes without Travel optimisation; 
b. Default values changed for Travel Optimisation and Cut Inner First (see #753)

### DIFF
--- a/meerk40t/core/cutcode.py
+++ b/meerk40t/core/cutcode.py
@@ -232,7 +232,16 @@ class CutObject:
     def generator(self):
         raise NotImplementedError
 
-    def contains_uncut_objects(self):
+    def contains_burned_groups(self):
+        if self.contains is None:
+            return False
+        for c in self.contains:
+            for pp in c.flat():
+                if pp.burns_done > 0:
+                    return True
+        return False
+
+    def contains_unburned_groups(self):
         if self.contains is None:
             return False
         for c in self.contains:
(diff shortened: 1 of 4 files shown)


tatarize (approved):
Did not crash in a giant ball of flames.

Sophist-UK:
> Did not crash in a giant ball of flames.

LoL. I do test these things before I submit them. 

(The burn complete path code only took three attempts, one of which was an infinite loop - no crashes or balls of flame though. The final grouping code was about the 20th attempt to work out how to do it right and do it in an efficient way - so there were the usual typos that caused crashes during coding, and then 19 simulations which didn't crash but didn't work - some had only one subpath burned, a couple had literally no groups, some just did the first group then didn't bother with the rest. So there were many crashes and small balls of flame along the way.)

PR #759, opened 2022-01-21
0.7.5-beta2 Version Bump

tatarize:
Should be version bump for the beta. We can check if there's anything else in the 0.7.5 line before a full release but betas should be pretty doable.


Sophist-UK (approved):
Looks good to me.

Perhaps we should automate this with workflow when a release tag is created and before builds are kicked off. We could then change it to a dev version (which includes a commit reference) whenever a PR is merged.

PR #763, opened 2022-01-22
"Open Project" should first clear everything

jaredly:
There's an 'import file' menu item for if you want to add a file to the current working project :)

### DIFF
--- a/meerk40t/gui/wxmmain.py
+++ b/meerk40t/gui/wxmmain.py
@@ -325,6 +325,17 @@ class MeerK40t(MWindow):
                 if fileDialog.ShowModal() == wx.ID_CANCEL:
                     return  # the user changed their mind
                 pathname = fileDialog.GetPath()
+                gui.clear_and_open(pathname)
+
+        @context.console_command("dialog_import", hidden=True)
+        def import_dialog(**kwargs):
+            files = context.load_types()
+            with wx.FileDialog(
+                gui, _("Import"), wildcard=files, style=wx.FD_OPEN | wx.FD_FILE_MUST_EXIST
+            ) as fileDialog:
+                if fileDialog.ShowModal() == wx.ID_CANCEL:
+                    return  # the user changed their mind
+                pathname = fileDialog.GetPath()
                 gui.load(pathname)
 
         @context.console_command("dialog_save", hidden=True)
@@ -1143,7 +1154,7 @@ class MeerK40t(MWindow):
         # ==========
         self.Bind(wx.EVT_MENU, self.on_click_new, id=wx.ID_NEW)
         self.Bind(wx.EVT_MENU, self.on_click_open, id=wx.ID_OPEN)
-        self.Bind(wx.EVT_MENU, self.on_click_open, id=ID_MENU_IMPORT)
+        self.Bind(wx.EVT_MENU, self.on_click_import, id=ID_MENU_IMPORT)
         self.Bind(wx.EVT_MENU, self.on_click_save, id=wx.ID_SAVE)
         self.Bind(wx.EVT_MENU, self.on_click_save_as, id=wx.ID_SAVEAS)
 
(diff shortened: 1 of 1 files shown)


meerk40t/gui/wxmmain.py
```diff
@@ -1783,14 +1813,6 @@ def load(self, pathname):
...
                 if n != self.context.elements.note and self.context.auto_note:
                     self.context("window open Notes\n")  # open/not toggle.
-                try:
-                    if self.context.uniform_svg and pathname.lower().endswith("svg"):
-                        # or (len(elements) > 0 and "meerK40t" in elements[0].values):
-                        # TODO: Disabled uniform_svg, no longer detecting namespace.
-                        self.working_file = pathname
-                        self.validate_save()
-                except AttributeError:
-                    pass
```
jaredly, line 1793: I moved this because it doesn't make sense to set the working file if we're just importing a file into the current project.

meerk40t/gui/wxmmain.py
```diff
@@ -1828,15 +1850,14 @@ def on_focus_lost(self, event):
         # event.Skip()
 
     def on_click_new(self, event=None):  # wxGlade: MeerK40t.<event_handler>
-        context = self.context
-        self.working_file = None
-        self.validate_save()
-        context.elements.clear_all()
-        self.context(".laserpath_clear\n")
```
jaredly, line 1835: I figured it was a good idea to extract this out so that I could re-use the function in clear_and_open.

tatarize (approved):

meerk40t/gui/wxmmain.py
```diff
@@ -1757,6 +1768,25 @@ def set_file_as_recently_used(self, pathname):
                 break
         self.populate_recent_menu()
 
+    def clear(self):
```
Sophist-UK, line 1771: Can we use a slightly more detailed method name than `clear` here? In my mind at least `clear` has a more generic usage than this.

Sophist-UK (changes requested):

meerk40t/gui/wxmmain.py
tatarize: In wxmain it clears the working file, all the elements and the laserpath in the background. I'm not sure clear is wrong here. There shouldn't be any work left.

Sophist-UK:
I also wonder whether we should rename `File Import` to `File Add` though I have no real opinion one way or the other except that Open vs. Add feels slightly better than Open vs Import.

jaredly:
Hmmmmm. Yeah I guess given that "opening meerk40t" doesn't clear the operations list, that does make sense. Although I've thought it is a little weird that the operations list persists across restart.

Sophist-UK:
Would you like to do the pr to fix this?

Sophist-UK:
Yes - perhaps a `clear_elems_branch` like this:
```
def clear_elems_branch(self):
    self.clear_elements()
    self.clear_files()
    self.clear_note()
    self.validate_selected_area()

def clear_all(self):
    self.clear_elems_branch()
    self.clear_operations()
```

jaredly:
should clicking "new" also preserve operations?

jaredly:
Yeah I guess I'm now remembering my logic for clearing the operations log: "Open" is now the same as "New + Import". Which makes sense to me?

Sophist-UK:
No I think we should just revert the removal of operations on file open. New should continue to remove operations IMO.

Sophist-UK:
> "Open" is now the same as "New + Import". Which makes sense to me?

With the benefit of hindsight, I do not think this should be the case. I think operations should be persistent until you explicitly class then with a file new.

jaredly:
🤷‍♂️ I think it's very confusing in general that operations are thought of as persisting outside of a project, but maybe that's just me.

tatarize:
Well, we borrow the project-like terminology without actually having projects. You can't really open multiple projects at the same time or whatever. Which you would tend to expect for a document-like workflow. Though some work was done checking through this in the 0.8.0 branch currently "project" is more-or-less a fiction.

PR #765, opened 2022-01-22
Fix Save to not open dialog if you opened a project file

### LINKED ISSUE #711: Save button is actually Save As
Even when File/Save does an immediate save, the Save button always pops up the Save dialog box.

jaredly:
This depends on https://github.com/meerk40t/meerk40t/pull/763, to make a distinction between "open" and "import" (because when we import a file, we don't want to save back to it).
fixes #711


tatarize (approved):

meerk40t/gui/wxmribbon.py
```diff
@@ -214,7 +214,7 @@ def open_simulator(v=None):
         )
         toolbar.Bind(
             RB.EVT_RIBBONBUTTONBAR_CLICKED,
-            lambda e: self.context(".dialog_save\n"),
+            lambda e: self.context(".on_click_save\n"),
```
Sophist-UK, line 217: I don't think this will work as there isn't a console command registered for `on_click_save`.

Sophist-UK (changes requested):
Since this includes the changes for #763 I would suggest that you close #763 and update the title and description of this one.

Sophist-UK:
P.S. Apologies for having missed you off the new contributors list on Facebook. I have rectified this.

meerk40t/gui/wxmribbon.py
jaredly: hmm it's definitely working for me, I thought it just keyed off of the method name

Sophist-UK: Yup.

Sophist-UK (approved):
LGTM

tatarize (approved):

PR #768, opened 2022-01-22
Force append \n for normal context calls

tatarize:
Forces \n if one doesn't exist within console `__call__` methods.

### DIFF
--- a/meerk40t/kernel.py
+++ b/meerk40t/kernel.py
@@ -149,6 +149,8 @@ class Context:
         return "Context('%s')" % self._path
 
     def __call__(self, data: str, **kwargs):
+        if len(data) and data[-1] != "\n":
+            data += "\n"
         return self._kernel.console(data)
 
     def boot(self, channel: "Channel" = None):


meerk40t/kernel.py
```diff
@@ -149,6 +149,8 @@ def __repr__(self):
         return "Context('%s')" % self._path
 
     def __call__(self, data: str, **kwargs):
+        if len(data) and data[-1] != "\n":
```
Sophist-UK, line 152: `if len(data) == 0 or data[-1] != "\n":`?

Sophist-UK (approved):
Make your own mind up about my comment. Approving this now in case you think my comment is wrong.

meerk40t/kernel.py
tatarize: if len(data) != 0

The truthy value of len(data) is `len(data) != 0` since len() returns an int and bool(int) is int 0 or non-zero

Sophist-UK: OK - I was suggesting that we append a "\n" if **either** the string is null **or** doesn't already end with it. But on reflection this doesn't seem right either and we should perhaps not bother sending a null string to console in any case.

PR #770, opened 2022-01-23
Add "ConsoleOperation" to allow interrupts to have notes and custom labels

### LINKED ISSUE #764: Add a note to Interrupts?
I'd love to be able to add a note next to interrupts. I'm using them to adjust the power settings, and it'd be great to have listed the power that I want to set it to.

jaredly:
Fixes #764 

Here's my first stab at it! I left in the original interrupt menu item so I could compare behavior, I'm happy to take them out once this is in good shape.

https://user-images.githubusercontent.com/112170/150663972-47504b97-5f61-4295-9001-da3a20b6c793.mp4

### DIFF
--- a/meerk40t/gui/wxmmain.py
+++ b/meerk40t/gui/wxmmain.py
@@ -315,20 +315,6 @@ class MeerK40t(MWindow):
                     context._stepping_force = None
             dlg.Destroy()
 
-        @context.console_command("dialog_load", hidden=True)
-        def load_dialog(**kwargs):
-            # This code should load just specific project files rather than all importable formats.
-            files = context.load_types()
-            with wx.FileDialog(
-                gui, _("Open"), wildcard=files, style=wx.FD_OPEN | wx.FD_FILE_MUST_EXIST
-            ) as fileDialog:
-                if fileDialog.ShowModal() == wx.ID_CANCEL:
-                    return  # the user changed their mind
-                pathname = fileDialog.GetPath()
-<<<<<<< HEAD
-=======
-                gui.clear_and_open(pathname)
-
         @context.console_argument("message", help=_("Message to display, optional"), default="")
         @context.console_command("interrupt", hidden=True)
         def interrupt(message="", **kwargs):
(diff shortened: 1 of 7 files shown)


tatarize:
This is so really robust work. Like it gives me a lot of faith that the api is really navigable if you can figure it out enough to put something this nice together that quickly.

meerk40t/gui/wxmmain.py
```diff
@@ -325,6 +340,7 @@ def load_dialog(**kwargs):
                 if fileDialog.ShowModal() == wx.ID_CANCEL:
                     return  # the user changed their mind
                 pathname = fileDialog.GetPath()
+                gui.clear_and_open(pathname)
```
tatarize, line 343: This line doesn't refer to any current code. And a such would crash.

tatarize (approved):
A little bit rough around the edges, might be better with a simple popup box rather than the larger gui there. But, is generally a highly robust PR. 1 line there wouldn't really work since the other changes as such isn't merged so that's not a command in main.

tatarize:
Already included but some rough edges.

* Double click on a node calls activation which generally loads up specific properties. You have some specific properties these aren't changed on node activate.
* The previous command in the node isn't in the text box on start. One would expect to maybe have a simple text entry box there, with the previous command already present.
* The editing window is probably overbuilt. I can't see using more than maybe the default Text Popup thing in reality.
* Editing. Closing and editing again displays the old value in the window rather than the new value you put in there the first edit.
* Interrupt commands generally are CLI and GUI. So in theory if you ran the code in the commandline it should work, but the command is really only included with a gui popup. Getting that right might be a little harder, the interrupt command for example saves itself in function/interrupt and switches out for text and gui forms. Though that's a very general form since you can spool any console commands there and only the gui lets you put in the console commands.
* There's no way to add in your special commands to the tree without gui. If I wanted to spool this entire thing from command line and use your special node types, there's no set methodology that makes that possible.

Most of these are minor, the `clear_and_open()` being added in there strictly doesn't work but everything else, despite some rough edges is pretty robustly done. Bravo.

tatarize:
* The divide between command and name is weird. You really only need one.

Smoothing out the edges I commented out the stuff and you get 1 text entry box. It took me a bit to figure out what was going on there with the name and commands differences, but largely this is an issue with adapting the code I think. Still waiting on the updated approval but still quite solid work.

meerk40t/gui/wxmmain.py
Sophist-UK: I think it was in @jaredly 's previous PR.

Sophist-UK:
:+1 from me too.

meerk40t/gui/wxmmain.py
tatarize: As do I, but it was sort of a fish out of water change there. And testing it crashed things. So it shouldn't have been included in the PR. There was another small bug where he loaded the `command` as an `int` when he used a string there that didn't really work. But, as far as PRs go this was/is decidedly quite robust. Most of my edge smoothing was commenting out parts of this. And adding in the routine to let you double-click to edit the values.

jaredly:
So I thought it could be cool to have a sicint label in the tree, that's a summary of the message that gets shown. but it's not a dealbreaker.
Thanks for helping smooth out the rough edges 😅

tatarize:
It got confusing as to which part was the command and which part was the unneeded label. It wouldn't be too hard to set the label independently. The label is still there but the editing of the command part updates the label currently. If less confusing it would be fine. Especially since you can make a larger number of commands in the node.

Issue #767, opened 2022-01-22
[Bug] New Group Inner Burns functgionality is missing out some inner burns and doing them at the end. 


Sophist-UK:
![image](https://user-images.githubusercontent.com/3001893/150649146-ed836a24-e902-4300-8a6f-385be53209e4.png)

Sophist-UK:
I will look into this tonight. At this stage I am not sure whether this is an issue with my code or the code which checks whether a path is inside or not.

Also, I would like to optimise groups so that they track the burn status (started, finished) at a group level rather than checking the path segments in the iterative candidate check. This should further improve the optimisation speed.

Sophist-UK:
This particular issue is definitely fixed in 0/.7.5 full.

PR #771, opened 2022-01-23
Fixes Nuke-Beta2 Crash bug.

tatarize:
* Prevents rastercut from being considered by the group expecting code.
* Adds first for rastercut, so it's considered in ordering for distances in new mode.

### DIFF
--- a/.github/workflows/win/file_version.txt
+++ b/.github/workflows/win/file_version.txt
@@ -6,8 +6,8 @@ VSVersionInfo(
   ffi=FixedFileInfo(
     # filevers and prodvers should be always a tuple with four items: (1, 2, 3, 4)
     # Set not needed items to zero 0.
-    filevers=(0, 7, 5, 1),
-    prodvers=(0, 7, 5, 1),
+    filevers=(0, 7, 5, 2),
+    prodvers=(0, 7, 5, 2),
     # Contains a bitmask that specifies the valid bits 'flags'r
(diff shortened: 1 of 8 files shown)


Sophist-UK:
Apologies for the raster bug!!

Sophist-UK:
Do you want to wait for beta3 until I diagnose the other (**non**-crashing) bug?

Sophist-UK (approved):

tatarize:
This should be fairly good enough. I think I got the non-crashing bug. Or rather made it work.

tatarize:
@Sophist-UK was there something else you wanted addressed before pushing out Beta3?

Sophist-UK:
Beta3 is urgent because of my bad. 

The two changes I want to make are not crashing things so less urgent and I cannot get to them until later. They will need to wait until Beta4 I think.

The bug in #767 still exists and will take some tracking down.

And the performance improvements can also wait.

tatarize:
Ah. Okay. I'm fine with additional betas. There's still a good amount of life in the 0.7.x line for features since I'm a bit better at merging in 0.8.0 than I was in 0.6.x 0.7.x divide. So features like ConsoleOp are easier to filter up to the higher levels of code. Just checking mostly against crashing and burning and slightly missed the mark on that one. All's well though.

Sophist-UK:
@tatarize Thanks for fixing this though.

PR #774, opened 2022-01-23
Improve Lhystudios config window

Sophist-UK:
1. Text and tooltips
2. Layout
3. Integer spin boxes


Sophist-UK:
# Before
image

# After
image

tatarize:
No real objections here though the work might be short lived since there's a shift in the config window for 0.8.0 to work more with panes since they are easier to disable and enable and to cover all the different features.

page1controller

page2controller

tatarize:
The shift home position isn't in pixels, it's actually in steps. Though really in 0.8.0 it needs to take a physical unit measurement there to work correctly since raw units are going to be nanometers.

Sophist-UK:
Yes - though some of this will translate to the new settings.

I thought I was thinking ahead by switching from mils to px (by which I meant steps) but I can revert this easily enough.

Sophist-UK:
P.S. I still plan to try to create a more standardised way of creating dialog boxes / panes - and this was part of the learning curve.

tatarize:
There's PropertyPanels these are used here and there and create panels for choices. So if I make a quickie dialog to change values stored in a choice you can call a PropertyPanel to do that. Any more standardized methods I would think should fall into that sort of thing.

tatarize:
Strict isn't a programmed speed mode, it's actually makes sure the directionality of the start of the vector cut comes from the TOP-LEFT always rather than going the typical and correct direction.

Sophist-UK:
I based the Strict description on the previous tooltip. Happy to use any text that helps a non-technical user decide whether to check the option or not. What are the symptoms that they might get that should prompt them to check this option?

tatarize:
https://github.com/meerk40t/meerk40t/blob/tatarize-services/meerk40t/gui/operationpropertymain.py

Might be worth looking at for tooltips. The core idea being to keep them together in larger triple quoted areas rather than stick them together in longer areas. There's a very real chance that somebody won't need to translate the entire tooltip so keeping it together may be for the best.

tatarize (approved):

PR #776, opened 2022-01-24
Fix Group Inner Burns

Sophist-UK:
This PR does three things:

a. Fixes VectorMontonixer having too few polygon points to ensure that small inner burns are not mistakenly considered outside.
b. Fixes Rasters not being considered for inner burns, and treats rasters correctly as being inside even if it extends outside;
c. Improves performance by maintaining burns_done at a group level rather than iterating over segments to find out.

### DIFF
--- a/meerk40t/core/cutcode.py
+++ b/meerk40t/core/cutcode.py
@@ -329,12 +329,6 @@ class CutGroup(list, CutObject, ABC):
             for s in c.flat():
                 yield s
 
-    def is_burned(self):
-        for c in self:
-            if c.burns_done != c.passes:
-                return False
-        return True
-
     def candidate(self, complete_path: Optional[bool]=False, grouped_inner: Optional[bool]=False):
         """
         Candidates are CutObjects:
--- a/meerk40t/core/cutcode.py
+++ b/meerk40t/core/cutcode.py
@@ -159,7 +159,7 @@ class CutObject:
         self.next = None
         self.previous = None
         self.passes = passes
-        self.burns_done = 0
+        self._burns_done = 0
 
         self.mode = None
         self.inside = None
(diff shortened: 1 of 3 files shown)


Sophist-UK:
I have tested this with rasters and images and dots.

VectorMontonizer is still using brute force approach (with 1000 polygon points rather than 100). We should consider using a more intelligent choice for:

a. polygon points - based on path curvature and length. Long straight lines don't need intermediate points, long curves will need intermediate points based on curvature, short path segments will need to be approximated by straight lines, and then points added based on changes of direction and / or maximum deviation (like 3d printer slicers do to reduce stuttering).

b. inner points - based on absolute size or size relative to the outer path.

Sophist-UK:
Performance test results (for a leather pattern with one outer and a couple of hundred inners):
|      | Main | This PR |
| --- | --- | --- |
| Inner | 4.051s | 1.970s |
| Greedy | 3.821s | 0.424s |

Sophist-UK:
Need to fix a poorly chosen algorithm in CutGroup burns_done.

Sophist-UK:
P.S. I think this is a good pre-requisite for my planned enhancement to break dispersed rasters into several smaller images.

tatarize (approved):

tatarize:
We may at some point need to consider the utilization of the plotplanner to cut up the paths. It's actually not that much slower from some of the checks against this with vpype and actually decided very definitional with regards to the shape. This might actually need an implementation of the sort of phase 2 potrace algorithm which would be rather helpful with some of these conversion.

http://potrace.sourceforge.net/potrace.pdf gives the Sellinger used for the straight path line detection metric. And give that we can actually get point by point curve sections generated from the zingl-bresenham line draw algorithm we could utilize a version of this to result in more-or-less straight line sections to approximate the curves we are using. This would be a weird method of linearizing but rather than a sample rate, we'd actually get remarkably linear sections that depend only and entirely on the native resolution at which we are applying the curves. So the zingl-bresenham curve drawing algorithms would give us a series of single step ortho and diagonal steps, and we'd combine these into line sections.

With a clear representation of the shape in linear segments at native resolution we'd have a shape that given your resolution would be identical to the original shape in perfectly sized linear segments. For some of the smarter boards this would be extremely helpful as well, as we could get nearly perfect curves made out of linear segments. Likewise for the vector monotonized this would also work to help provide solid insideness definitions.  Since the only time we'd overlap these points is if they actively intersect. It would give us perfect sets of lines at exactly the resolution we needed.  Since it would derive from the curve drawing algorithm and line plotting exactly from that plotted path.

PR #778, opened 2022-01-24
Fix a couple of translation issues

jpirnay:
a) PO File as ususal
b) A colon at the end of a frame label in the GUI effectly prevents a tranlsation.

### DIFF
--- a/locale/de/LC_MESSAGES/meerk40t.po
+++ b/locale/de/LC_MESSAGES/meerk40t.po
@@ -1,4 +1,4 @@
-# German translations for PROJECT.
+# German translations for Meerk40t.
 # Copyright (C) 2021 ORGANIZATION
 # This file is distributed under the same license as the PROJECT project.
 # FIRST AUTHOR <EMAIL@ADDRESS>, 2021.
@@ -6,10 +6,10 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: PROJECT VERSION\n"
+"Project-Id-Version: Meerk40t 0.7\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
 "POT-Creation-Date: 2021-09-11 02:52-0700\n"
-"PO-Revision-Date: 2021-09-11 03:17-0700\n"
+"PO-Revision-Date: 2022-01-24 03:17-0700\n"
 "Language: de\n"
 "Language-Team: de <LL@li.org, JP>\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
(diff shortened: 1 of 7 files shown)


jpirnay:
Not sure what happens in the background but if you leave the ":" at the string, i.e. _("Units:")
then the corresponding translation in the .po files could not be found, i.e.

> #: meerk40t/gui/panes/position.py:86 meerk40t/gui/settings.py:26
> msgid "Units"
> msgstr "Einheiten"

I tried to fix it in one locale by adding the colon there, that did not work out... So I fixed it in the code, and the colon is superfluous in the label anyway IMHO.
Before:
2022-01-24 22_34_16-Lhystudios-Configuration
After:
2022-01-24 22_33_27-MeerK40t v0 7 5-beta3 src      (0 -_ lhystudios -_ lhystudios)

tatarize:
The colons might be regionally a different thing. They are added there because Mac requirements actually wanted them, but they aren't actually essential to leave in the translated version. If a different language used ; or - or whatever that should totally be valid and they should absolutely be translated in those cases.

tatarize:
Ah, the message id actually needs to be "Units:" in those cases. It was changed but but template file was not updated correctly. These changes were made to fit the Mac Requirements for the AppStore. It doesn't match because it literally doesn't match. When it said "Units" that was the old value it should properly say "Units:" in order to find the match. The translation isn't working because the translation file needed to be updated for the orphan requirements and the removal of the : should be reverted.

Rather than file itself should be update to include these translation elements.

tatarize:
errors-in-current

The current settings for the non-updated version has a few items that migrated into the project according to poboy (translation software for python projects I'm sort of writing on and off when it comes up).

tatarize:
https://github.com/meerk40t/meerk40t/pull/779 did a very basic update there to check things compared with the older stuff seems to have the colon items though.

jpirnay:
I am happy to revert/discard the code changes and work on the po files instead. But then all language templates  need to be updated, I checked with French and Hungarian in the compiled version, they suffer from the same phenomenon.
Pardon my ignorance, please: is there a tool that scans the code and updates the pos files with newly added strings (as there are quite a lot of them outstanding to be included).
I am more than happy to take over this part of the workload if you tell me, what tools to use.
And BTW: I just love meerk40t, it's such a great tool and you discover any moment new functionalities...

jpirnay:
Ahh, on my previous comment: poboy provides the code scan functionality. Hmm, I need to have a closer look to make sure that existing translations are not getting lost.

Sophist-UK:
Perhaps we need a workflow to regenerate the master translation file every time a PR is merged.

jpirnay:
I will close the pull request and rework the po-files

PR #779, opened 2022-01-25
Check and Update Translation information via Poboy

tatarize:


Sophist-UK (approved):

Sophist-UK:
Obviously we cannot automate the translation of new strings in a Github workflow (or could we automate an initial automated translation from workflow using an API?), but can we update the US English PO file from the python files on every PR merge and (if appropriate??) update the untranslated strings in the other languages?

And we probably need workflow to regenerate the .mo files every time a PR is merged that changes .po files.

tatarize:
Pretty sure we could actually merge in the new string each time, we'd need to edit poboy to have the functions we want but it wouldn't be too hard. I could put it up on pip and we could certainly download it and run it in a github script. We currently have most of the functions we'd need augmented with a lot of babel's backend. While I didn't quite plan poboy around that sort of thing it wouldn't be too hard to combine it with a much more extensive CLI front end and much more minor operations. We'd be fairly safe implementing find source strings that aren't in the .pot file, update them. Find .pot strings that aren't in the po and update them. Compile the .mo files for anything that passes the error checks. (eg. printf %s exists in the original but is absent in the translated version).

PR #780, opened 2022-01-25
Extend translation capability

jpirnay:
Extended translation capability to the Node-Headings in the Tree pane


tatarize (changes requested):
Rather than import wx. There you need to do `_ = kernel.translation` Doing things this way will make MeerK40t require wxPython for commandline operations and would fail execution in default mode. 

There's every reason to have translations but the GetTranslation() thing in wx isn't set in non-gui code. The `kernel.translation` call there or `_ = context._`  since we're in the init there. Defaults to a dummy operation if we don't have translations.

jpirnay:
Will close this one and reopen one with just the context._
(Still trying to understand the finer details of how I can make changes to a pull request already submitted...)

PR #782, opened 2022-01-26
Extend translations

jpirnay:
Makeing labels in tree panel translatable


tatarize (changes requested):
This version might not actually catch on the sources catch. I'm unsure specifically but the function must be `_()` and I'm not sure `context._()` actually matches that. The catches for finding the translation strings is anything called with `_()` as a function. It iterates through all the source-code with the python source code iteration thing and picks out all cases where a string is applied in a function like that. I do not know if context._() works the same. I'd feel safer with `_ = context._` before those lines. And calling `_` as a function to make sure it catches in the generate sources code.

jpirnay:
Obsolete

PR #783, opened 2022-01-26
Disallow Drag / Drop of some invalid image combinations

Sophist-UK:
Proposed rules

1. Disallow drop of anything other than an Image onto Image Ops.
2. Disallow drop of images onto a Dot Op.

Do this for drops onto "op" or "opnode" nodes.

Test what happens if drop of single node fails, and where multiple nodes partly success and partly fail.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -190,62 +190,102 @@ class Node:
             if drop_node.type == "op":
                 # Disallow drop of non-image elems onto an Image op.
                 # Disallow drop of image elems onto a Dot op.
-                # Dragging element into operation adds that element to the op.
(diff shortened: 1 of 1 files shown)


Sophist-UK:
Ok - I have written the code, but I have not yet run it to check for syntax errors or tested it properly. But a desk review would be appreciated to confirm I am on the right lines.

Sophist-UK:
Ok - aside from typos ("Dot" vs "Dots") it looks like this works first time. I have tested it with both single and multiple nodes being dropped.

Also fixed special ops not being equal to ops for resequencing.

tatarize (approved):

PR #784, opened 2022-01-27
Version Bump 4003 == Beta 3

tatarize:
Version Bump


tatarize:
```python
        if not getattr(sys, "frozen", False) or platform.system() != "Darwin":
            self.file_menu.AppendSubMenu(self.recent_file_menu, _("&Recent"))
```

Was not not included but the rest of these branched off as their own PRs.

PR #785, opened 2022-01-27
Fix Mac pinch-to-zoom

Sophist-UK:


Sophist-UK:
I don't have a Mac so this is untested. Over to @TheKeeb for testing.

meerk40t/gui/scene/scene.py
```diff
@@ -278,13 +278,15 @@ def on_right_mouse_up(self, event):
...
-        if magnify > 0:
-            self.scene.event(event.GetPosition(), "zoom-in")
-        if magnify < 0:
-            self.scene.event(event.GetPosition(), "zoom-out")
+        magnification = event.GetMagnification()
+        if event.IsGestureStart():
+            self.scene.event(event.GetPosition(), "gesture-start")
+        elif event.IsGestureEnd():
+            self.scene.event(event.GetPosition(), "gesture-end")
+        else:
```
tatarize, line 288: The gesture-start gesture-end stuff might not be needed but could be of some use at some point.

meerk40t/gui/scene/scene.py
```diff
@@ -278,13 +278,15 @@ def on_right_mouse_up(self, event):
...
+            self.scene.event(event.GetPosition(), "zoom {mag}",format(mag=magnification))
```
tatarize, line 289: This can't be used. "zoom " here has been used for the pinch and zoom screen gestures which are already using that functionality and that zoom is active from gesture-start.

1) Touch Screen Pinch-Zoom: ZoomFactor is zoom from the start of the zoom event.
2) TouchPad Pinch-Zoom: Magnify is the instantaneous zoom from the previous zoom event in a range of around [-0.4,0.4].

These two are not compatible. One is compounded the other is not.

meerk40t/gui/scene/scene.py
```diff
@@ -301,7 +303,7 @@ def on_gesture(self, event):
                 zoom = event.GetZoomFactor()
             except AttributeError:
                 zoom = 1.0
-            self.scene.event(event.GetPosition(), "zoom %f" % zoom)
+            self.scene.event(event.GetPosition(), "zoom {zoom}".format(zoon=zoom))
```
tatarize, line 306: This says `zoon=` and will crash.

meerk40t/gui/scene/scene.py
```diff
@@ -1384,7 +1386,7 @@ def event(self, window_pos=None, space_pos=None, event_type=None):
             return RESPONSE_CONSUME
         elif event_type == "lost":
             return RESPONSE_CONSUME
-        elif str(event_type).startswith("zoom"):
+        elif str(event_type).startswith("zoom "):
```
tatarize, line 1389: That space is needed since the zoom-in and zoom-out would fit. The question was is the first active word there actually zoom the hyphen forms are to avoid catching this since these are different zoom styles.

tatarize (changes requested):

tatarize (dismissed):
I went through and made sure this fit the work done with Keeb. Still best that it gets checked.

tatarize (approved):
While these changes seem a bit more impactful the code should be a lot easier to read and make it clear what's going on here. The commented sections could be removed.

Sophist-UK:
@tatarize This synergy is how Open Source should work - so thanks for making this improvement even better.

I think we still need this testing before I merge it.

tatarize:
```
TheKeeb — Today at 12:41 AM
yes
ship it!
```

tatarize:
@tiger12506 Is there any way to add @mjalindahl as a co-author after the fact or something?

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line

Dunno, but this really should show up as his work too.

tiger12506:
Unfortunately git commits are hashed as a whole. To do so, we'd need to back up to that point and rewrite every commit since then. Possible? Yes. A good idea? Well... not really.

In future, note that it's trivial to do so before the branch gets shared in the repo.

Sophist-UK:
I think the best you can do is to note in the PR the specific contribution and add that person as a contributor in the wiki / About window.

And of course the direct expression of thanks to the person which you almost certainly have already done.

PR #792, opened 2022-02-02
Fix bind alias issues

Sophist-UK:
1. Provide update functionality if default bind / alias is updated in code and user has not changed the default.
2. Fix Keybind window blank after `bind default`.

### DIFF
--- a/meerk40t/core/bindalias.py
+++ b/meerk40t/core/bindalias.py
@@ -232,12 +232,12 @@ class BindAlias(Modifier):
         keys.clear_persistent()
         alias.clear_persistent()
 
-        for key, value in self.keymap.items():
+        for key, value in self.context.keymap.items():
             if key is None or len(key) == 0:
                 continue
             keys.write_persistent(key, value)
 
-        for key, value in self.alias.items():
+        for key, value in self.context.alias.items():
             if key is None or len(key) == 0:
                 continue
             alias.write_persistent(key, value)
--- a/meerk40t/core/bindalias.py
+++ b/meerk40t/core/bindalias.py
@@ -1,5 +1,108 @@
 from ..kernel import CommandMatchRejected, Modifier
 
+# The following dicts consist of a tuple of values, the first of
+# which is the current default, with any prior defaults following.
+# Prior defaults are used to keep a users keymap / aliases up to date with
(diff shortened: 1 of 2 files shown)


Sophist-UK:
Hmmm - I seem to have  broken a whole bunch of UnitTests. I will take a look tomorrow and see whether I have actually broken something or whether the tests themselves are using the wrong context.

Sophist-UK:
Ok - it was an easy fix.

meerk40t/core/bindalias.py
```diff
@@ -135,105 +232,58 @@ def save_keymap_alias(self):
...
-        if not len(self.keymap):
+        context = self.context
+        context.keymap.clear()
+        prefs = context.derive("keymap")
+        prefs.kernel.load_persistent_string_dict(prefs.path, context.keymap, suffix=True)
+        if not len(context.keymap):
             self.default_keymap()
+            return
+        for key, values in DEFAULT_KEYMAP.items():
+            if key in context.keymap and context.keymap[key] in values[1:]:
```
tatarize, line 254: Tested and this was insufficient.

I added a key `control+q` equal to `testing`
```python
     "control+q": ("testing",),
```
And ran this code. As control+q is not in the context.keymap it did not add my new key. I'll go ahead and accept since this implies *update* but I think maybe a third thing, where unless they bound that key for something else or specifically set it to none, it should rebind that key. Part of the problem is that new binds also do not get added.

I do not know how you tell a new bind from a bind that previously existed and was then set to none, but it seems like the spirit of the changes should address this problem. If not, say the word and I'll approve this.

tatarize (changes requested):

meerk40t/core/bindalias.py
Sophist-UK: If I understand this correctly, you added a DEFAULT_KEYMAP entry for an entirely new key not currently mapped in your settings in the code and it was NOT added to the bindings?

If this is what you are reporting, I agree - and 2 lines of additional code are required (plus two for the similar alias code).

Sophist-UK:
2 bugs fixed and a few minor tweaks.

tatarize (approved):

PR #793, opened 2022-02-03
Make delete key usage consistent

Sophist-UK:
`tree selected` was actually using emphasised rather than selected. So e.g. deleting an image under an image operation also deleted the image under elements.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -4459,7 +4459,6 @@ class Elemental(Modifier):
                 if n.type not in ("root", "branch elems", "branch ops"):
                     # Cannot delete structure nodes.
                     n.remove_node()
-                    break
             return "tree", [self._tree]
 
         @context.console_command(
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -4452,13 +4452,14 @@ class Elemental(Modifier):
         )
         def delete(channel, _, data=None, **kwargs):
             """
-            Delete node. Due to nodes within nodes, only the first node is deleted.
+            Delete nodes.
             Structural nodes such as root, elements, and operations are not able to be deleted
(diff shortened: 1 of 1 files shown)


tatarize (approved):
Odds are pretty good that'd work.

tatarize:
@tiger12506 does this actually fix the stuff you run into there?

tiger12506:
I believe this fix is correct (It deletes references in Operations without deleting the Elements), but doesn't fix inpain's issue (that I reproduce) of multi-select delete not working as intended.

tatarize:
Improvements even if they don't address other issues are still good to go.

Sophist-UK:
I'll check out multi-select delete before we merge this. What **exactly** are the symptoms?

tiger12506:
Selecting multiple items in the tree and hitting delete only deletes one item.

Sophist-UK:
Good to go again.

meerk40t/core/elements.py
```diff
@@ -4459,7 +4459,6 @@ def delete(channel, _, data=None, **kwargs):
                 if n.type not in ("root", "branch elems", "branch ops"):
                     # Cannot delete structure nodes.
                     n.remove_node()
-                    break
```
tatarize, line 4462: "Delete node. Due to nodes within nodes, only the first node is deleted."

* Break serves the purpose of being lazy.

tatarize (changes requested):
Either leave the break and take your win, or fix the harder issue that tries to delete nodes from the leaves going backwards in a depth-last fashion.

tatarize (approved):
I don't see anything wrong with that solution.

Sophist-UK:
Tested it before (crashed) and after (didn't crash) so should work.

PR #794, opened 2022-02-03
Fix scene zoom on file load

Sophist-UK:
I found it very annoying that the scene on file load was random - so here is a fix.


Sophist-UK:
I did consider zoom to file size - but it gets a bit more complicated if you have multiple files loaded. And generally I think an overview of the file on the scene is better to allow you to decide whether the design is positioned properly and roughly the right size, whereas a zoom into a design which is 0.01mm wide will look great at first sight, but the burn probably won't be what you are expecting.

PR #795, opened 2022-02-03
Raster optimisations

Sophist-UK:
This PR consists of a range of optimisations to allow Rasters to be burned more intelligently - and also some more general code-quality improvements.

1. Rather than generating a single, potentially large, raster image for each operation, with raster sweeps potentially across large swathes of non-burn areas, this groups raster elements within a raster op into groups of overlapping rasters, so each group is disconnected from the others, and creates several smaller raster images, one for each group.

2. If Cut Inner First and Group Inners are **not** both set, we optimise the size of these raster images in order to get the best trade-off in time between sweeping at a fixed speed across a non-burn distance and splitting the tasters into separate groups each having its own acceleration / deceleration time. In essence, we estimate the acceleration / deceleration time by extrapolating from experimental results, and from that the distance that the head would travel in that time - the theory being that for that specific distance between images the time taken to burn as one image sweeping across the empty area is the same as the time needed to do the additional acceleration / deceleration. Then we group raster areas together by extending their margins by half that distance i.e. the distance swept in the time needed to either accelerate or decelerate.

3. If Cut Inner First and Group Inners are both set, then instead we create the smallest possible raster images in order to allow the inner first algorithm to determine which raster areas are inside which closed curves.

    Once the inner first algorithm has done its work, we then combine any of these small raster images that are all inside the same set of outer closed cut paths into a single image on the same basis as 2a. above.

4. These raster optimisations are entirely automated and no additional user options are needed to enable / disable them.

5. The CutPath code object has been split off into a separate file in order to reduce the length of the planner.py file. Many other Codacy issues (mainly code complexity) are also addressed.

6. This code relies on https://github.com/meerk40t/svgelements/pull/166 which causes a particular issue with closed cut sub-paths in combined paths.

Note: A future further optimisation with large benefits is possible to reduce large dimension raster images which have a large empty area inside (such as picture frames) to be split into several smaller images - separate raster images would likely be required for cross-hatch raster operations with this optimisation. Another future optimisation (but without large benefits) is to allow Vert and Horiz rasters where the optimiser can decide whether to burn e.g. a Vert one from T2B or B2T whichever has a closer starting point.


Sophist-UK:
This PR is now ready for review.

These changes are non-trivial and should be given both a decent desk review (which may take a reasonable amount of time) and a good test.

Sophist-UK:
I note that unittests are now failing - I will need to look and see why this is before we merge this - however the main code is definitely ready for review and testing.

tatarize:
@Sophist-UK this became desynced. corrected again to fit with main.

Sophist-UK:
Thanks.

tatarize (approved):
Tested code on actual job, does seem to clearly take some superior pathing at times breaking the raster objects into multiple images when a single image isn't otherwise beneficial.

PR #797, opened 2022-02-04
Fix critical codacy errors

Sophist-UK:
Part 1 of several PRs to fix codacy errors.

Most of these should be obviously transparent, but changes to meerk40t/core/drivers.py for `opt_jog_mode` almost certainly need double checking.

### DIFF
--- a/meerk40t/core/drivers.py
+++ b/meerk40t/core/drivers.py
@@ -74,7 +74,9 @@ class Driver:
         context._quit = False
 
         self.rapid = self.root_context.opt_rapid_between
-        self.jog = self.root_context.opt_jog_mode
+        # # there is a method called jog and the following line overwrites it
+        # # I can find no other references to self.jog as an int in the code.
+        # self.jog = self.root_context.opt_jog_mode
         self.rapid_override = False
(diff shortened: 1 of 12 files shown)


meerk40t/core/drivers.py
```diff
@@ -191,7 +193,9 @@ def _fetch_next_item_from_spooler(self):
             self.spooled_item = element
         else:
             self.rapid = self.root_context.opt_rapid_between
-            self.jog = self.root_context.opt_jog_mode
+            # # there is a method called jog and the following line overwrites it
+            # # I can find no other references to self.jog as an int in the code.
+            # self.jog = self.root_context.opt_jog_mode
             try:
```
tatarize, line 199: This does nothing. Actually self.rapid also does nothing.

tatarize (approved):
Even if there were issues this is so chalked full of marked improvements that I'd take some issues which there shouldn't be any of.

Sophist-UK:
Good to go.

tiger12506 (approved):
Tat already left a glowing review on this. The changes afterward seem minor and look ok to me.

PR #799, opened 2022-02-04
Python version stuff

Sophist-UK:
1. Enable Python 3.10 for Pypi
2. Python 3.6 introduced math.tau so remove unneeded code setting up tau.

### DIFF
--- a/meerk40t/gui/scene/scenewidgets.py
+++ b/meerk40t/gui/scene/scenewidgets.py
@@ -1,8 +1,8 @@
-from meerk40t.gui.scene.scene import Widget
-from meerk40t.gui.wxutils import create_menu
-
 import math
 import wx
+
+from meerk40t.gui.scene.scene import Widget
+from meerk40t.gui.wxutils import create_menu
 
 from meerk40t.gui.laserrender import (
     DRAW_MODE_BACKGROUND,
--- a/meerk40t/gui/scene/scenewidgets.py
+++ b/meerk40t/gui/scene/scenewidgets.py
@@ -1,13 +1,7 @@
 from meerk40t.gui.scene.scene import Widget
 from meerk40t.gui.wxutils import create_menu
 
-try:
-    from math import tau
-except ImportError:
-    from math import pi
-
-    tau = 2 * pi
-
+import math
 import wx
 
 from meerk40t.gui.laserrender import (
@@ -987,7 +981,7 @@ class GuideWidget(Widget):
                     ends.append((x, h - length - edge_gap))
 
                     # gc.DrawText("%g %s" % (mark_point, p.units_name), x, 0, -tau / 4)
-                    gc.DrawText("%g" % mark_point, x, edge_gap, -tau / 4)
+                    gc.DrawText("%g" % mark_point, x, edge_gap, - math.tau / 4)
             x += points
 
         y = offset_y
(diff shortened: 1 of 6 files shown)


meerk40t/gui/scene/scenewidgets.py
```diff
@@ -987,7 +981,7 @@ def process_draw(self, gc):
                     ends.append((x, h - length - edge_gap))
 
                     # gc.DrawText("%g %s" % (mark_point, p.units_name), x, 0, -tau / 4)
-                    gc.DrawText("%g" % mark_point, x, edge_gap, -tau / 4)
+                    gc.DrawText("%g" % mark_point, x, edge_gap, - math.tau / 4)
```
tatarize, line 984: This code is 3.6 code. We currently still support 3.5 but don't really need to.

tatarize (changes requested):

meerk40t/gui/scene/scenewidgets.py
tatarize: This can be corrected if you want to edit main.py to change to throwing the error at 3.6 instead of 3.5. We'd get access to fstrings (which are great but do not translate), underscores in binary numbers, and tau.

Sophist-UK: I did check this before I removed it - the Pypi config says Py3.6 upwards.

Indeed, both Py3.5 has been out of support for almost 2.5 years and Py3.6 has been out of support for over a year.

If you want I will switch it to math.pi/2.

Sophist-UK: > This can be corrected if you want to edit main.py to change to throwing the error at 3.6 instead of 3.5. We'd get access to fstrings (which are great but do not translate), underscores in binary numbers, and tau.

I will do this. Personally I have no issues with allowing 3.6 even though it is unsupported.

Sophist-UK:
Good to go.

tiger12506 (approved):
Looks like Tat already approved the main changes here. Nothing looks strange to me. Sophist justified them fairly. Looks good in my book.

Sophist-UK:
@tiger12506 There really is no problem with you having merged this, but the protocol is (apparently) for PRs from people who have merge rights is to let the originator merge it (I guess in case they have last minute changes of mind about anything - we have all been there, right?).  😄

tiger12506:
@Sophist-UK I've asked Tat about it several times, and he has said that he has no issue with me merging other people's items, it's his personal preference to not merge and allow the originator to merge. If you would like for me to let your merge your own, that's fine.

Sophist-UK:
It really is not that big a deal - but on balance I would prefer to merge my own PRs only because sometimes having had it approved I think "oh - I could have done that slightly differently" and throw in another tweak. Of course, I can always open a new PR if it has already been merged, so it is not a big deal either way really.

tiger12506:
Seems like you ought to trick yourself into thinking it was reviewed, just so you get that "done slightly differently" inspiration *before* you put out the pull request for review... Not that I care about it too much, but it is a little jarring to review/approve of something only to have github reject the review because you made changes afterward.

I'll make a mental note not to merge yours.

Sophist-UK:
@tiger12506 I do try hard not to request a review until I believe it is finished - but sometimes as you sleep or in the shower you get a brainwave about how you could have done something better. I know it then makes the review (and the reviewers time) a waste, but that is life.

PR #800, opened 2022-02-05
Divvy up node classes

tatarize:


Sophist-UK:
I fear that, of all the changes you have made for 0.8.0, this one is going to prove painful. Because merging in any changes in 0.7.x for elements.py is now going to be a right PITA.

Sophist-UK:
I should have said that I think this change is a good thing, just that I think that the approach will have consequences.

In fact, you might actually find it a better strategy to revert this completely, do the same thing in main and merge that into 0.8.0 in order to keep some consistency between the two branches.

tatarize:
You might be a little bit right. This is just leading to a bigger set of changes to divide up LaserOperation into the different operations it uses. This change was not a serious problem it worked fairly well and wasn't too breaking. The following change of making `opnode` to `refelem` was fine.

--

Dividing up LaserOperation tho is very breaking. I do not want to break main. Like ever. This simplifies a lot of thing but it should never be applied to 0.7.x if anything we should start to correct things in 0.8.x and replace main with it and send 0.7.x to legacy7 or whatever. 

While this is actually really nice. Since all nodes have `.type` and only the op node types have `.operation` this seems to simplify a lot of code. It makes a lot of other simplifications possible, like the drag nodes logic is all in Node. And that could be made much more reasonable. It opens up the door to maybe doing the same thing to the elem node. So rather than ElemNode for SvgImage, Path, Polygon, etc I could make each of those their own nodes too. Which would simply that logic too. Though I might hold off on that one since it's not important just yet.

Sophist-UK:
I agree. Splitting up an over large file into several smaller ones with essentially the same code is relatively low risk IMO and is safe for 0.7.

Re-architecting nodes is a bigger risk (though not as large as rearchitecting the kernel) and should be left to 0,8 and left until late on in 0.8 so we do not either restrict changes in 0.7 over long or give ourselves a headache making the same changes in 0.8 as well.

I still recommend backing out this change, re-doing it in 0.7 and then rolling it into 0.8. You could try merging this commit into 0.7 and fix any merge conflicts, but I suspect that and bugs you then need to fix in 0.7 because it needs to be different may then create incompatible merges back to 0,8. But I am not a git expert.

tiger12506:
I recommend finalizing 0.7.x line and moving on to 0.8 line. What was the hold-up again?

Sophist-UK:
If we do that then any further rearchitecting would need to wait for 0.9.

I just recommend as before - revert this and reapply it to 0.7 and then merge it in to 0,8 - and then finish the re-architecting and move to make 0.8 a stable release AFAP.

Sophist-UK:
> What was the hold-up again?

Balor and Grbl apparently. 😃

tatarize:
I would think if we did the same exact changes to 0.7.x we could potentially keep them similar. At first I thought Sophist meant the full set of changes there but there are some really sorts of harsh changes but simply dividing up that class into smaller subclasses actually went off without a hitch.

I think there is one last final architectural change that might be needed for 0.8.0 and it might not be too problematic. Basically we need to finalize the coordinate system.

1) Physical bed width.
2) Physical bed height.
3) Origin point location within the physical bed space scaled to the unit square.
4) X-scale.
5) Y-scale.
6) X-flip.
7) Y-flip
8) Rotation

The problems are that coordinate systems like grbl and balor don't share enough in common with the MK coordinate system. If we move the home position we basically need most things to switch up heavily to match this. I think if we map these parameters out for whatever drivers we should be able to map whatever coordinate system onto another coordinate system. At the least it's enough information and the various values don't require that we change the other values when we change things. so if we say the bed is bigger and the origin point is located at `1,0` (upper left hand corner), we don't have to change our other properties. Same if we somehow home into a different corner, we can say the device is in position `0,0` and treat that like it's really `(bedwidth,0)` within the device coordinate system.

We load an svg and we can map the SVG space into MK space. (they are very similar) and we can load up a gcode file or dxf file and those have an origin point in the lower left hand corner, which we could map correctly into the space based on the same principles Those however are origin point within the physical bed space scaled to the unit square. So our origin point is located in the lower left, and that implies an offset that is `(0,bedheight)`, then we have the flipped y direction for them so the +y after that offset is located up and to the left of the origin point. And we scale everything within MK at the size of physical units, and call the unitless values as equal to pixels 1/96th of an inch. We could also perhaps hard define this based on user preferences of their unit types. Basically `move 20 20` becomes completely ambiguous when we're dealing with mapped spaces like this.

We then scale for native as something that is done within plot planning. And push any conversions to device specific units into the plot planner. We say cutcode is non-scaled and in the device coordinate system. This gives us the levels of separation required to convert these units fairly arbitrarily. Things like offsets can be done in device specific units. But, we should be able to map the different spaces that way. However there are some things with galvos where the field can become flipped or rotated. But, these are always really going to be tau/4 turns though with this many different values we might actually just end up being better off with an affine transformation matrix that is generated from these different value, or just apply a black box sort of mapping between one field and the other field, after all we could just calculate the transformation matrix based knowing these properties. And if we need to apply a different operation we could just map from location to location assuming we're still dealing with standard Euclidean space.

Sophist-UK:
I am not a git expert (I think your first name might have to be Linus to qualify), but I fear that applying a change to 0.8.0 and then a similar change to main is **not** going to be the same as reverting this change and then applying a very similar one to main and merging it into 0.8.0. Specifically I worry that it will then be difficult to merge main with one version of this code into 0.8.0 that already contains a slightly different version of the code under different commit references.

PR #801, opened 2022-02-05
Bindalias and keybind improvements

Sophist-UK:
This has been prompted by issues with mappings for e.g. alt-f/alt-h which conflict on windows (and randomly override) menu shortcuts.

### Reserved keys
Regardless of the platform used by individual developers, IMO bindalias needs to conform to BOTH of the following:
1. Mac - common shortcuts defined in https://support.apple.com/en-us/HT201236
should be followed where appropriate and avoided if not. This is important
for MK to be accepted into the Apple app store.
Where keys defined in other sections on this page make sense for this application,
these should also be used by preference.
2. Windows - alt-single letter should be avoided as these correspond to menu
shortcuts (e.g. in english locale, alt-f should activate the file menu).
Common Windows shortcuts can be found at
https://support.microsoft.com/en-us/windows/keyboard-shortcuts-in-windows-dcc61a57-8ff0-cffe-9796-cb9706c75eec

In addition, where it does not conflict with the above, any synergy with keys commonly used in popular SVG / image editors (e.g. poaint.net, inkscape) may be beneficial.

## Fixes included
1. More definitive recognition of modifier keys and special keys esp. on Mac.
2. Keymap window list is now nicely sorted
3. Code simplification through lookup tables
4. "control+" is now "ctrl+" for consistency with menus
5. ctrl+g is now equivalent to the gigantic green Go button. (Suggestion by @tiger12506)
6. on a mac a left click whilst control is pressed is converted to a right click. (Suggestion by @joerlane)
7. Keybinds on mac now show correct mac keys
8. Ability to update default keymaps and have them applied iff user has not remapped the key already (by retaining details of old mappings).
9. bind and alias console lists are now sorted and columns aligned, and alias also shows a keymap if that exists.
10. keymap execute is now a common bindalias.py function call, and it can execute multiple commands like alias can. reset_bind_alias is now eliminated as an alias because it is no longer needed


tatarize:
```control-z should be reserved for future Undo
(though reset is a close equivalent right now unless we mapped it to file reload).```

This should actually call a command called undo currently. And just map that command (currently) to calling whatever we need. The redo command is `shift+control+z` rather than y.

**Comment by tatarize** [comment]
I think the Undo/Redo test went fairly well. I needed a method of copying the entire operation tree. But, beyond that the main part of the idea worked fairly well and it stacked somewhat nicely.

**Comment by tiger12506** [comment]
I don't have strong opinions on the structure or what the keybinding keys are, but I do think that we should take a hard look at the defaults and see if there are improvements that can be made.

For instance:
1) The shapes in 0.8.0 need new bindings that reflect units, otherwise they are tiny in nano units. Also, I think adding 'stroke red classify' is a good idea.
2) Dunno if my "Go" keybind is implemented as a default? Doesn't seem like it. Less important now because of Laser Panel for the public, but still handy.
3) There are a million things I don't know about meerk40t and a lot of these can be exposed as example commands. And the most common place I see those are in keybinds... `outline 1mm`   -- Holy shit! Wish I had seen that before!  We really ought to come up with some serious command documentation, "recipes", if you will.
4) escape should be estop NOT pause, pause (or p) should be pause.

**Comment by Sophist-UK** [comment]
> I don't have strong opinions on the structure or what the keybinding keys are, but I do think that we should take a hard look at the defaults and see if there are improvements that can be made.
> 
> For instance:
> 
> 1. The shapes in 0.8.0 need new bindings that reflect units, otherwise they are tiny in nano units. Also, I think adding 'stroke red classify' is a good idea.

I agree. I guess it probably is useful to do a circle / square etc. because you can then resize these using the Position pane.

> 2. Dunno if my "Go" keybind is implemented as a default? Doesn't seem like it. Less important now because of Laser Panel for the public, but still handy.

Yes - I would like this!!!! I'll take a look.

> 3. There are a million things I don't know about meerk40t and a lot of these can be exposed as example commands. And the most common place I see those are in keybinds... `outline 1mm`   -- Holy shit! Wish I had seen that before!  We really ought to come up with some serious command documentation, "recipes", if you will.

Go for it!! The Wiki is waiting with baited breath for your text...

> 4. escape should be estop NOT pause, pause (or p) should be pause.

The pause key is already pause. However, IMO the escape key is too frequently used to e.g. cancel typing text etc. to abort something like a burn outright - if you are preparing the next job whilst the current one is executing, then the escape key is all too easy to hit by accident.

**Comment by tiger12506** [comment]
The command run for LaserPanel Start.
planz clear copy preprocess validate blob preopt optimize spool0
(for the Go binding)

**Comment by tiger12506** [comment]
inpain's rant today about key mappings is relevant to this pull request.  Do we already have some structure in place for platform-specific keybinding defaults? If not, there probably should be. His example of Ctrl being right-click is especially troubling for mac/win compatibility. I know it's been discussed forever by inpain and Tat, but a lot of that gets internalized and not documented in a way that helps out issues like this that happen in the future. So mac needs to be considered here in this remapping work.

**Comment by Sophist-UK** [comment]
wx already converts command into ctrl on macs. there is a rawctrl you can test which should work like menu.

**Comment by Sophist-UK** [comment]
Ok - mac code is now tested and working. This is (I hope) good to go,

**Comment by Sophist-UK** [comment]
@tatarize, @tiger12506 @joerlane  - Ready for review again.

**Review by tatarize — APPROVED** [review]
Probably still might need inpain's review on the mac parts

**Comment by Sophist-UK** [comment]
I have requested @joerlane 's review some time back and he has not done so. I can't see this making Mac stuff worse, so if he doesn't review it soon, I suggest we merge it and we can iterate when he does use it if it needs further work.

**Comment by tatarize** [comment]
I likely need to take a look at the differences in modifiers between the #825 and #801 here and see if they can be less of a broken-home when both of these are merged.

**Comment by Sophist-UK** [comment]
IMO there is not a conflict between modifiers between these two PRs which are addressing different things.

The concern I have is that there is (generally - not just in #825) a confusion about how we handle events - should we use the wx model of events whereby MK code hooks into mouse and keyboard events as they require and wx handles the hierarchy of event processing to ensure that the correct piece of code (frame, tree, scene, scenewidget etc.) handles a specific event, or for genuine reasons of correlating events which need to be stateful should we centralise the handling of mouse and keyboard events and have MK parse these and call relevant modules to process them. (Note: I have no view on which is correct, only a hunch that it may not be a good idea to mix and match).

If I felt that this PR (#801) was contributing to this confusion I would be the first to put it on hold, but since it only makes existing code more comprehensive and does not change how events are handled, I would suggest that this one should be approved and merged.

Similarly, although #825 does contain material new wx event handling, I still recommend that it is approved and merged, and  that subsequent to both these merges, we take a look at event handling and see if it needs to be rationalised.

**Comment by tatarize** [comment]
WxPython handling of such things is fine. It comes down to how much code could/should be reused. And here we could imagine writing something that could maybe use some of bind/alias in a different gui system. We could also see maybe using scene stuff only in a wx context. So really neither of these is that useful in a CLI or different GUI system. Centralizing things is sometimes useful but sometimes less useful depending on whether doing that creates large amounts of coupling between the different systems. We could maybe merge bind/alias stuff a bit deeper into scene so we deal with binds as a keybind widget within the scene. Or we could try to remove all the specific wx parts from the keybinds so that it's plausible to use keybinds in a CLI context. Though neither of these seems to be logically right.

If I had to guess I would say keybinds should be available to widgets, and our main keybinds should catch whatever our various keystrokes and if not caught it should pass it down the chain. This isn't substantively different than allowing keybinds to caught at the wx layer and then propagate down into scene if they aren't otherwise dealt with, and the way this happens depends a bit on the current system and how scene should deal with hotkeys.

I'd think we'd be best to integrate #825 keystroke stuff to be a generalized set of modifiers that would catch whatever sorts of binds we might want. Then if the keybinds get consumed at the outermost level by something like bind-alias code it should rather defer to scene in this situation. Same for other things like joystick values which could, like mouse and keyboard, matter for some particular widgets and should be there for the scene operations. It depends a bit on how it currently works and what's the best method of generalizing these interactions. 

If we have a hotkey binds qua bindalias and we have a scene. We need to have a hotkey matter to both, and the chief method for this to deal with the bindalias first and then the scene, depending on how possible that is within the current system that would inform the changes we make in the system itself.

**Comment by Sophist-UK** [comment]
Before considering this, we need to note that Mouse and Scene events are slightly different in scope. A keyboard event can apply anywhere in the Frame - but unless you are e.g. allowing drag of elements from scene to tree (I am not recommending this but dragging elements from scene to an Opnode might make some sense) or from tree to scene (makes no sense whatsoever) then mouse events are going to be limited to the scene. 

My gut reaction is that we should handle both Keyboard and Mouse events in Frame, Panes and Scene and not in SceneWidgets - and that this would create the correct balance of centralisation. I think that this is essentially the current approach, so any cleanup would not require wholesale rework. We need to ensure that there is consistency over handling of Keyboard events - I note that if there is e.g. an alt-f keybind, then (on Windows) this prevents using the accelerator to open the File menu if you last clicked in Scene, but doesn't disable it if you last clicked in Tree - this is inconsistent and I think the Tree has it wrong.

**Comment by jpirnay** [comment]
On another note: the wxutils.get_key_name routine may need some more lines to acknowledge KeyUp for Shift and Ctrl:
```
    if keyvalue == "":
       if key == wx.WXK_SHIFT:
            keyvalue = "shift+"
        elif key == wx.WXK_CONTROL:
            keyvalue = "ctrl+"
        elif key == wx.WXK_ALT:
            keyvalue = "alt+"
```
these are currently not recognized

Sophist-UK:
> On another note: the wxutils.get_key_name routine may need some more lines to acknowledge KeyUp for Shift and Ctrl:
> ...
> these are currently not recognized

There has not previously been a use case for recognising these keyup cases, which is why no one has spotted that they are not reported until now. But you are right that on key-up we should report these too, though there may need to be extra code (particularly in keymap) to deal with this situation.

Since we have not heard from @joerlane I am going to merge this and put this on my to-do list (unless you would like to pick it up Jens).

jpirnay:
I think it's getting rather messy with the treatment of multi-modifier-keypresses:
a) Scenario 1: Try ctrl-shift-o for outline of an object, works like a charm. Now press Shift-O, keep them pressed and add the ctrl-key - nothing happens
b) In the keybind-window press Ctrl-Shift and release the Shift-Key: nothing
c) In the keybind-window press Ctrl-Shift-O and release the Shift-Key, add or remove other modifiers as you want but keep the O-pressed

Not sure whether this is the right approach, but I think we need to distinguish a couple of things:
1) key-down: one solution could be to keep track which keys are still pressed and then compose the get_key_name event appropriately
2) key-up: if you release i.e. in the above example (ctrl-shift-o) the o-key then you get back what you expected: ctrl-shift-o. If you release the ctrl key only then what: imho you should you get at least a key-up event for ctrl and what about the still pressed shift-o (sometimes an immediate follow-up event for keydown gets raised, sometimes not). 
...

Sophist-UK:
I think this depends on whether you trigger things on a non-modifier keydown or keyup.

Either way you are going to use the modifiers in place when the non-modifer key goes down or up and AFAIK the current get_key_name handles this correctly in both situations. But the Keymap window might not i.e. if you do Ctrl then P then shift it may not show the correct thing. I would need to check.

Sophist-UK:
#853 seems to do it.

Non-modifier should be processed on keydown. So Ctrl-P-Alt should correctly show Ctrl-P whereas Ctrl-P-Alt-P should show Ctrl-Alt-P.

PR #804, opened 2022-02-06
[Proof of concept] Table driven lookups rather than endless `elif`s

Sophist-UK:
When I converted this code, there were several places where methods or opt variables referred to did not exist, and several pieces of code that simply hid errors without reporting them.

Even when I commented out the lines requiring methods or variables that don't exist I still got hundreds of messages of the following format:
```
drivers.py: unknown command: <generator object home at 0x0000023BD69CCB30>, args ()
drivers.py: unknown command: <generator object CutCode.generate at 0x0000023BD69CCB30>, args ()
```

This is beyond my ability to debug - so I am creating a PR only as a proof of concept, but of course anyone suitably knowledgeable can resurrect it or use this code as the basis of their own PR.

### DIFF
--- a/meerk40t/core/drivers.py
+++ b/meerk40t/core/drivers.py
@@ -1,3 +1,4 @@
+from functools import partial
 import platform
 import os
 import time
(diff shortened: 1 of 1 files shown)


tatarize:
This code goes away in 0.8.0 and is actually built for speed. The timing of a lookup dict is slower than a giant block of elifs. There previous to 3.11 weren't case statements for python and when you used a dictionary things went slower so I just got in the habit of making a big block of ifs.

Sophist-UK:
I would have thought that a dictionary lookup (done on a very small number of P-code instructions (with the loop running in C) would be faster than if statements which would have several P-code instructions for each test.

Sophist-UK:
@tatarize 

The big issues uncovered by this PR are:
a. Missing methods that would create run-time crashes - hopefully Codacy will catch these
b. Code which simply ignores errors created elsewhere

P.S. Can you please change the Codacy quality settings to failing checks on Medium severity rather than Minor severity.,

tiger12506:
No, dictionary lookup is slower if there are a small number of items. (somewhere in the less than a hundred range, I think?), remember there is the hashing calculation which touches each item.... but really the difference is marginal, and I think readability plays a role too.
Pretty sure I recall Tat actually tested it (of course).

Definitely one of those "premature optimization" things to replace a dictionary with a string of if/elif   lol.

Sophist-UK:
I have no knowledge of whether Tat tested it or not, but IMO it would be completely in character and I have absolutely zero reason to disbelieve it when he says he has tested it. Whether it was a good decision depends on just how time critical the code is. The lower level in the stack you are, the more verbose the control language and the more frequently and time critical the code is. And this code is in driver.py so pretty low level.

But equally, code elsewhere is becoming more compute intensive - e.g. the optimisations - so there is more reason for users to run the code on faster machines so that optimisation happens in a reasonable time, and this then benefits these low level routines too.

Bottom line is that I am perfectly happy that this example didn't work. But I am definitely concerned by code which hides errors because that leads to unreliability.

PR #805, opened 2022-02-07
Reenable old-style Github issues.

Sophist-UK:
Experts need more flexibility to submit issues.


tiger12506 (approved):
This is a good thing because Tatar can continue using issues as he has always done. The issue templates are a bit too strict for internal use, but are probably handy for end-users.

Issue #802, opened 2022-02-06
[Feature] Laser Panel Arm checkbox


tiger12506:
### Summary Description

Laser panel should have an Arm checkbox.
Preference should exist (default off) to disable this feature (by disabling arm checkbox and setting it permanently armed)

LaserFireButton.Enabled = ArmCheckbox.Checked
if (pref.checked) {
  armcheckbox.checked = true
  armcheckbox.enabled = false
}

### MeerK40t Version

main

### MeerK40t Type

Git

### Your Operating System

Linux

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #806, opened 2022-02-07
Add LaserPanel Arming #802

tatarize:
This should be an implementation of #802. LaserPanel Arming. This is fairly easy to turn off either with right click or in the preferences menu. All relevant UI switchings happen.


tiger12506 (approved):
I tested this out and it looks great! Thanks for the preference setting!

PR #811, opened 2022-02-09
Fix / Extend missing translations

jpirnay:
Fix / Extend missing translations
a) DE-Locale
b) Fix in Simulation-Window
c) Add missing tooltip

### DIFF
--- a/translate.py
+++ b/translate.py
@@ -1,0 +1,22 @@
+import os, polib
+
+def create_mo_files():
+	data_files = []
+	localedir = './locale'
+	po_dirs = [localedir + '/' + l + '/LC_MESSAGES/'
+	for l in next(os.walk(localedir))[1]]
+	for d in po_dirs:
+		mo_files = []
+		po_files = [f
+					for f in next(os.walk(d))[2]
+					if os.path.splitext(f)[1] == '.po']
+		for po_file in po_files:
+			filename, extension = os.path.splitext(po_file)
+			mo_file = filename + '.mo'
+			po = polib.pofile(d + po_file)
+			po.save_as_mofile(d + mo_file)
+			mo_files.append(d + mo_file)
+		data_files.append((d, mo_files))
+	return data_files
+	
+data_files=create_mo_files()
--- a/translate.py
+++ b/translate.py
@@ -13,9 +13,18 @@ def create_mo_files():
 		for po_file in po_files:
 			filename, extension = os.path.splitext(po_file)
 			mo_file = filename + '.mo'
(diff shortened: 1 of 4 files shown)


translate.py
```diff
@@ -0,0 +1,31 @@
+import os, polib
```
Sophist-UK, line 1: Please split this line into two separate imports.

translate.py
```diff
@@ -0,0 +1,31 @@
...
+				doit = 0 
+			if doit == 1:
+				print("Translate " + d + po_file)
+				po = polib.pofile(d + po_file)
+				po.save_as_mofile(d + mo_file)
+				mo_files.append(d + mo_file)
+		data_files.append((d, mo_files))
+	return data_files
+	
+data_files=create_mo_files()
```
Sophist-UK, line 31: I think this should just be `create_mo_files()`.

translate.py
```diff
@@ -0,0 +1,31 @@
...
+	for l in next(os.walk(localedir))[1]]
+	for d in po_dirs:
+		mo_files = []
+		po_files = [f
+					for f in next(os.walk(d))[2]
+					if os.path.splitext(f)[1] == '.po']
+		for po_file in po_files:
+			filename, extension = os.path.splitext(po_file)
+			mo_file = filename + '.mo'
+			doit = 1
```
Sophist-UK, line 16: `doit` should use `True` and `False` rather than 1 and 0.

Sophist-UK (changes requested):

tatarize (approved):
The translate.py thing is fine, it's mostly a non-integrated standalone script and doesn't really need to exist, but it's fine if it does. polib isn't really a requirement here and, in theory, it would fall to the wayside if `poboy` got finished up to a nice and workable state.

I think it's fine to just hit the script with black and isort and merge it in normally.

translate.py
jpirnay: Will amend

translate.py
jpirnay: Right...

translate.py
```diff
@@ -0,0 +1,31 @@
+import os, polib
+
+def create_mo_files():
+	data_files = []
+	localedir = './locale'
+	po_dirs = [localedir + '/' + l + '/LC_MESSAGES/'
+	for l in next(os.walk(localedir))[1]]
```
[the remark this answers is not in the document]
jpirnay: I've 'black'ed the script...

Sophist-UK:
Closing in favour of #812

PR #812, opened 2022-02-09
Fix / Extend missing translations

jpirnay:
Fix / Extend missing translations
a) DE-Locale
b) Fix in Simulation-Window
c) Add missing tooltip


tatarize (approved):
Not sure how this differs from the other pr.

Sophist-UK:
@jpirnay is not yet a Git expert - we all know just how hard it can be to get to grips with git - hence this is another version with my code comments addressed.

PR #813, opened 2022-02-09
Version to 0.7.4010 -- Beta4

tatarize:
Version Bump.

As this is a pre 0.7.5000 it should be considered beta and published as a minor version.

### DIFF
--- a/.github/workflows/win/file_version.txt
+++ b/.github/workflows/win/file_version.txt
@@ -6,8 +6,8 @@ VSVersionInfo(
   ffi=FixedFileInfo(
     # filevers and prodvers should be always a tuple with four items: (1, 2, 3, 4)
     # Set not needed items to zero 0.
-    filevers=(0, 7, 5, 2),
-    prodvers=(0, 7, 5, 2),
+    filevers=(0, 7, 4010, 0),
+    prodvers=(0, 7, 4010, 0),
     # Contains a bitmask that specifies the valid bits 'flags'r
     mask=0x3f,
     # Contains a bitmask that specifies the Boolean attributes of the file.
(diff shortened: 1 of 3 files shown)


tatarize:
Should just be the version thing. I'd say we should either publish the beta or full version in short order either before or after the ruidacontrol and (less sure still) raster optimizations.

Sophist-UK:
I have no idea why we are switching from 0.7.5.x to 0.7.4010. I was not involved in any discussions about this, though I am guessing it may be Apple Store related.

I have no idea whether this is a good idea or not - is it in line with the Semantic Versioning standard? What would the consequences of this change be (if any)? For example, how can the updater (when I get around to finishing it) distinguish between full releases and beta releases?

I think the rationale needs to be documented in the first comment and any potential downsides considered before we make this change.

tatarize:
@Sophist-UK Pushed this to 0.7.5000 for full release. It's compatible with semantic versioning since it's just very minor changes in the minor version bit. We tend to do full and beta version. And this way we could do more full-ish version and just minor bump the version. It's probably best to do a full version before we move on to #730 and #795 adding. And I'll restore the beta tags after this version so it should be like 0.7.5001 beta-1 sorts of things again. Otherwise you're right that this could get somewhat confusing. I have no qualms with full releases that aren't directly aligned at the 1000 place or whatever, and that wouldn't be very clear without the whole beta things.

tatarize:
* As a system we can still call anything a beta. But, nice round numbers for full version markers seems fine.
* We can bump version much more often and for easier reasons and likely release a lot more beta versions if we want.
* The version numbering lets us call 4003 or whatever as 0.7.5-beta3 which lets things count up to actual 5000 rather than doing what we were doing before and incrementing the super-minor version.
* It was sort of useful for the mac store thing where paiin needed to burn a bunch of version numbers for very little reason.

There's a good set of reasons to do this and since it's still at the very least monotonic with other versions it will not cause any issues. Since 5000 is certainly greater than 4. We'll keep adding the beta tags which you're right might get a little confusing if we didn't somewhat continue that.

PR #819, opened 2022-02-11
Add portuguese Language files to Workflows

jpirnay:


tatarize (approved):
Good catch.

tatarize:
Well, they are somewhat blank. I think we also have some basic machine done japanese in a PR somewhere.

PR #820, opened 2022-02-11
Update translate 2

tatarize:
Salvage some of the work from the last effort before we decided that we should likely correct things using `{format}` data rather than printf for the sake of translations.


tatarize:
Okay, this seems pretty solid. Ports in the work from the previous attempt and corrects the directories to actually use the files consistently.

Sophist-UK (approved):

PR #821, opened 2022-02-11
Improve pulse_grouping

Sophist-UK:
Previous code would convert 0101 to 0011 and 1010 to 1100 but it would not handle 11011011 to 11100111.

This code handles the above and also improves performance by using deque instead of list.

### DIFF
--- a/meerk40t/core/plotplanner.py
+++ b/meerk40t/core/plotplanner.py
@@ -392,7 +392,7 @@ class PPI(PlotManipulation):
 class Shift(PlotManipulation):
     def __init__(self, planner: PlotPlanner):
         super().__init__(planner)
-        self.shift_buffer = deque([], 6)
+        self.shift_buffer = deque([], 4)
         self.shift_pixels = 0
 
     def __str__(self):
@@ -407,16 +407,10 @@ class Shift(PlotManipulation):
     #   the input patterns until all shifted bits have moved to at least the first position
     #   i.e. shifted bits will not be shifted again.
     PIX_SHIFT = {
-        0b010010:   0b001100,
-        0b010100:   0b001100,
-        0b010101:   0b001101,
-        0b010110:   0b001110,
-        0b010111:   0b001111,
-        0b101000:   0b110000,
-        0b101001:   0b110001,
-        0b101010:   0b110010,
-        0b101011:   0b110011,
-        0b101101:   0b110011,
+        # 0b0100:   0b1000,
+        0b0101:   0b0011,
+        0b1010:   0b1100,
+        # 0b1011:   0b0111,
     }
 
     def process(self, plot):
@@ -441,7 +435,7 @@ class Shift(PlotManipulation):
             self.process_add(x, y, on)
 
             # When buffer is full start popping off values.
-            if len(self.shift_buffer) >= 6:
+            if len(self.shift_buffer) >= 4:
                 yield self.process_pop()
 
     def process_add(self, x, y, on):
@@ -450,7 +444,7 @@ class Shift(PlotManipulation):
         self.shift_pixels &= (1 << len(self.shift_buffer)) - 1
         if on:
             self.shift_pixels |= 1
-        assert self.shift_pixels <= 63
+        assert self.shift_pixels < 16
         if self.shift_pixels in self.PIX_SHIFT:
             self.shift_pixels = self.PIX_SHIFT[self.shift_pixels]
 
--- a/meerk40t/core/plotplanner.py
+++ b/meerk40t/core/plotplanner.py
@@ -440,27 +440,30 @@ class Shift(PlotManipulation):
                 yield x, y, on
                 continue
 
-            # Shift() is on.
-            self.shift_pixels <<= 1
-            if on:
-                self.shift_pixels |= 1
-            self.shift_pixels &= 0b111111
+            self.process_add(x, y, on)
 
-            self.shift_buffer.insert(0, (x, y))
-            if self.shift_pixels in self.PIX_SHIFT:
-                self.shift_pixels = self.PIX_SHIFT[self.shift_pixels]
-            if len(self.shift_buffer) >= 6:
-                # When buffer is full start popping off values.
-                bx, by = self.shift_buffer.pop()
-                bon = (self.shift_pixels >> 5) & 1
-                yield bx, by, bon
+            # When buffer is full start popping off values.
+            if len(self.shift_pixels) >= 6:
+                yield self.process_pop()
+
+    def process_add(self, x, y, on):
+        self.shift_buffer.insert(0, (x, y))
+        self.shift_pixels <<= 1
+        self.shift_pixels &= (1 << len(self.shift_buffer)) - 1
+        if on:
+            self.shift_pixels |= 1
+        assert self.shift_pixels <= 63
+        if self.shift_pixels in self.PIX_SHIFT:
+            self.shift_pixels = self.PIX_SHIFT[self.shift_pixels]
+
+    def process_pop(self):
+        x, y = self.shift_buffer.pop()
+        on = (self.shift_pixels >> len(self.shift_buffer)) & 1
+        return x, y, on
 
     def flush(self):
         while self.shift_buffer:
-            self.shift_pixels <<= 1
-            bx, by = self.shift_buffer.pop()
-            bon = (self.shift_pixels >> 5) & 1
-            yield bx, by, bon
+            yield self.process_pop()
         self.clear()
 
     def flushed(self):
--- a/meerk40t/core/plotplanner.py
+++ b/meerk40t/core/plotplanner.py
@@ -1,3 +1,5 @@
+from collections import deque
+
 from meerk40t.tools.zinglplotter import ZinglPlotter
 
 from ..device.basedevice import (
@@ -392,7 +394,7 @@ class PPI(PlotManipulation):
 class Shift(PlotManipulation):
     def __init__(self, planner: PlotPlanner):
         super().__init__(planner)
-        self.shift_buffer = []
+        self.shift_buffer = deque([], 6)
         self.shift_pixels = 0
 
     def __str__(self):
@@ -401,6 +403,23 @@ class Shift(PlotManipulation):
             str(self.shift_buffer),
             bin(self.shift_pixels),
         )
+
+    # Notes:
+    #   As output pattern is shifted to the left and new bits added it will not match any of
+    #   the input patterns until all shifted bits have moved to at least the first position
+    #   i.e. shifted bits will not be shifted again.
+    PIX_SHIFT = {
+        0b010010:   0b001100,
+        0b010100:   0b001100,
+        0b010101:   0b001101,
+        0b010110:   0b001110,
+        0b010111:   0b001111,
+        0b101000:   0b110000,
+        0b101001:   0b110001,
+        0b101010:   0b110010,
+        0b101011:   0b110011,
+        0b101101:   0b110011,
+    }
 
     def process(self, plot):
         """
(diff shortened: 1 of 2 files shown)


tatarize (dismissed):

tatarize:
Tried patterns at 5 since 6 is actually a lot more extreme.

```python
    PIX_SHIFT = {
        0b01000: 0b10000,
        0b01001: 0b11000,
        0b01010: 0b11000,
        0b01011: 0b11100,
        0b01100: 0b11000,
        0b01101: 0b11100,
        0b01110: 0b11100,
        0b01111: 0b11110,
        0b10001: 0b11000,
        0b10010: 0b11000,
        0b10011: 0b11100,
        0b10100: 0b11000,
        0b10101: 0b11100,
        0b10110: 0b11100,
        0b10111: 0b11110,
    }
```

This code does the work over 5 rather than 4. And we take whatever pattern given and we turn it into all the ones on the left and zeros on the right. 

We do not process anything that starts with 00 or 11. And we don't process entries that would replace values with themselves.

```
Group Pulse Test
================
Test 001100110011001100110011001100110011001100110011001100
Grpd 011110000111100001111000011110000111100001111000011000
```

```
Group Pulse Test
================
Test 10101010101010101010101010101010101010101010101010
Grpd 11000111000111000111000111000111000111000111000010
```

```
Group Pulse Test
================
Test 110111011101110111011101110111011101110111011101
Grpd 111111100011111111100011111111100011111111100001
```

While the results here are a bit more extreme they are also much more clumpy, which is sort of the point. We can also very likely expand this to a generalized solution. Sort the bits over the window-space.

meerk40t/core/plotplanner.py
```diff
@@ -421,35 +440,34 @@ def process(self, plot):
...
-            if len(self.shift_buffer) >= 4:
-                # When buffer is full start popping off values.
-                bx, by = self.shift_buffer.pop()
-                bon = (self.shift_pixels >> 3) & 1
-                yield bx, by, bon
-        # There are no more plots.
+            self.process_add(x, y, on)
+
+            # When buffer is full start popping off values.
+            if len(self.shift_pixels) >= 6:
```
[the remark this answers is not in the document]
Sophist-UK: You are absolutely correct - it even says so in the comment above.

Sophist-UK:
```python
    PIX_SHIFT = {
        0b01000: 0b10000, # Although this is a legal move, why would we shift this bit?
        0b01001: 0b11000, # This is shifting the rightmost bit 3 places! 
        0b01010: 0b11000, # This is shifting the rightmost bit 2 places!!
        0b01011: 0b11100, # This is shifting the rightmost bit 2 places!!
        0b01100: 0b11000, 
        0b01101: 0b11100, # This is shifting the rightmost bit 2 places!
        0b01110: 0b11100, # This doesn't seem necessary!
        0b01111: 0b11110,
        0b10001: 0b11000,
        0b10010: 0b11000,
        0b10011: 0b11100,
        0b10100: 0b11000,
        0b10101: 0b11100,
        0b10110: 0b11100,
        0b10111: 0b11110,
    }
 ```
I have a feeling you think that this array does the bit shifting one place to the left as well, but it doesn't.

**Comment by Sophist-UK** [comment]
Just reread this and understand what you are suggesting. Here is a counter example...
```
10011111111111
10001 -> 11000
110001 -> 111000
1110001 -> 1111000
```
Serious continue shifting to the right and ones to the left indefinitely.

It is preventing this from happening but ensuring that the shifted bits never match a pattern as they are shifted left that makes this so difficult to get your head around.

**Review by tatarize — CHANGES_REQUESTED** [review]
Likely requires UI changes to register the degree of pulse-grouping without breaking the lighter pulse grouping mechanism (or rewriting it to being a more general manner). So rather than True False. The pulse_grouping (which would require a new parameter name to avoid type-clashing) would give you 0, 1, 2 or whatever for modes.

I note this set of changes is a bit extreme but enhancements shouldn't break already existing functionality but rather enhance that. While this scheme is solid enough and we can fiddle with the dictionary replacements this doesn't so much enhance the pulse grouping as replace it with a different one wholesale. It needs to permit both original and enhanced.

**Comment by Sophist-UK** [comment]
Most of this discussion has been about double or triple bit shifting. However the scope of this PR is only to make the existing single-bit shifting algorithm more comprehensive -  because it still only a choice between no grouping and single-bit grouping, the existing Pulse Grouping checkboxes are sufficient to allow this selection.

However, I do wonder whether a different 4-bit algorithm would suffice without allowing bits to continue to shift after their first move, as follows:

#### Current mapping
0101->0011
1010->1100

With these there are no doubles in the before column, and always doubles in the after column so it requires a minimum of 3 shifts before it can match again and the first bit has not previously changed so it is impossible for bits to move more than once.

However thinking about it again, I believe that it might be possible to add two more rules to the 4-bit and still preserve the above three rules: A) no doubles in the first 3 input bits, B) always last 2 output bits the same, C) don't change bit 4 - and providing that you preserve these 3 rules, you will always require 3 shifts before another match and the new first bit will never have previously shifted.

0100->1000
1011->0111

The example sequences that were never optimised - 010010 and 101101 - would now be optimised as follows (though we need to know what the next digit is to work it out (first 4 bits, then shifted left 3 before another possible match hence 7 bits):

0100100->1001000
0100101->1000011
1011010->0111100
1011011->0110111

If you agree with analysis, I can rework this to handle it in 4 bits as before.

**Comment by Sophist-UK** [comment]
@tatarize Bump

**Comment by tatarize** [comment]
> @tatarize Bump

Still has changes requested. I need to be able to set the mode of pulse grouping so that the default is still none, and the original type of pulse grouping can be set. So you can choose the degree of pulse grouping. Currently this breaks the ability to optionally set old style pulse grouping. It just makes pulse grouping stronger it does not necessarily make it better. An option to use stronger pulse-grouping *would* make it better but that's not what this is. This takes the only form of pulse grouping and just makes it stronger.

**Comment by Sophist-UK** [comment]
I'll look at this next in preference to raster optimize for 0.8.

**Comment by Sophist-UK** [comment]
Test results existing legacy code:
```
Test 0100010110101011
Grpd 0100001111000111
Reduced transitions from 11 to 5

Test 0000000100100011010001010110011110001001101010111100110111101111
Grpd 0000000100100011100000110110011110001001110001111100110111101111
Reduced transitions from 31 to 23

Test 1111111011011100101110101001100001110110010101000011001000010000
Grpd 1111111011011100011111001001100001110110001110000011001000010000
Reduced transitions from 31 to 23

Test 11101101011011000001111010010000101110001001100101111001111111010011111001011100
Grpd 11101110011011000001111100010000011110001001100011111001111111100011111000111100
Reduced transitions from 37 to 25

Test 10100000011001001101101011101011000100010111111000001111100110101001100011000000
Grpd 01100000011001001101110011110011000100001111111000001111100111001001100011000000
Reduced transitions from 37 to 28

Test 01000001001110001000100111110100111111110101111000000100011010111111011000011000
Grpd 01000001001110001000100111111000111111111001111000000100011100111111011000011000
Reduced transitions from 32 to 26
```

Test results with 6-bit version:
```
Test 0100010110101011
Grpd 0100001111001011
Reduced transitions from 11 to 7

Test 0000000100100011010001010110011110001001101010111100110111101111
Grpd 0000000011000011100000110110011110001001110001111100110111101111
Reduced transitions from 31 to 21

Test 1111111011011100101110101001100001110110010101000011001000010000
Grpd 1111111100111100011111001001100001110110001110000011001000010000
Reduced transitions from 31 to 21

Test 11101101011011000001111010010000101110001001100101111001111111010011111001011100
Grpd 11110011100111000001111100010000011110001001100011111001111111100011111000111100
Reduced transitions from 37 to 23

Test 10100000011001001101101011101011000100010111111000001111100110101001100011000000
Grpd 01100000011001001110011011110011000100001111111000001111100111001001100011000000
Reduced transitions from 37 to 28

Test 01000001001110001000100111110100111111110101111000000100011010111111011000011000
Grpd 01000001001110001000100111111000111111111001111000000100011100111111011000011000
Reduced transitions from 32 to 26
```

Test result with additional 4-bit strings as discussed above:
```
Test 0100010110101011
Grpd 1000001111000111
Reduced transitions from 11 to 4

Test 0000000100100011010001010110011110001001101010111100110111101111
Grpd 0000001001000011100000101110011110010001110001111100101111011111
Reduced transitions from 31 to 23

Test 1111111011011100101110101001100001110110010101000011001000010000
Grpd 1111110110111100011111010001100001101110001110000011010000100000
Reduced transitions from 31 to 23

Test 11101101011011000001111010010000101110001001100101111001111111010011111001011100
Grpd 11011110010111000001111100100000011110010001100011111001111111100011111000111100
Reduced transitions from 37 to 25

Test 10100000011001001101101011101011000100010111111000001111100110101001100011000000
Grpd 01100000011010001011110011110011001000001111111000001111100111010001100011000000
Reduced transitions from 37 to 28

Test 01000001001110001000100111110100111111110101111000000100011010111111011000011000
Grpd 10000010001110010001000111111000111111111001111000001000011100111110111000011000
Reduced transitions from 32 to 25
```

Of the three, I still prefer the output of the 6 bit version because it makes fewer unnecessary changes. But I am happy whichever of the above is chosen.

File: meerk40t/core/plotplanner.py [hunk 2]
```diff
@@ -402,6 +402,17 @@ def __str__(self):
...
 
+    # Notes:
+    #   As output pattern is shifted to the left and new bits added it will not match any of
+    #   the input patterns until all shifted bits have moved to at least the first position
+    #   i.e. shifted bits will not be shifted again.
+    PIX_SHIFT = {
+        0b0100:   0b1000,
+        0b0101:   0b0011,
+        0b1010:   0b1100,
+        0b1011:   0b0111,
```
tatarize, line 413: 4 changes

meerk40t/core/plotplanner.py
```diff
@@ -421,35 +432,34 @@ def process(self, plot):
...
-            self.shift_pixels <<= 1
-            if on:
-                self.shift_pixels |= 1
-            self.shift_pixels &= 0b1111
-
-            self.shift_buffer.insert(0, (x, y))
-            if self.shift_pixels == 0b0101:
-                self.shift_pixels = 0b0011
-            elif self.shift_pixels == 0b1010:
-                self.shift_pixels = 0b1100
```
tatarize, line 434: 2 changes.

meerk40t/core/plotplanner.py
```diff
@@ -421,35 +432,34 @@ def process(self, plot):
...
+
+    def process_add(self, x, y, on):
+        self.shift_buffer.insert(0, (x, y))
+        self.shift_pixels <<= 1
+        self.shift_pixels &= (1 << len(self.shift_buffer)) - 1
+        if on:
+            self.shift_pixels |= 1
+        assert self.shift_pixels < 16
+        if self.shift_pixels in self.PIX_SHIFT:
+            self.shift_pixels = self.PIX_SHIFT[self.shift_pixels]
```
tatarize, line 449: This code bit is rather critical and run a zillion times. There's absolutely going to be better methods of optimizing this. Please use one. Two dictionary lookups of an absolutely tiny dictionary to find 1 entry seems problematic. This could be done with `get` with a `default` value or a list-based hash-table with value->value keys. You could also possibly time the entire set to show that this code is somehow on par with methods that are almost certain to be 2x the speed at least.

tatarize (changes requested):
Does seem to have extra tweaks that might result in non-legacy manipulations (which is fine). But, it also needs a bit of hyper-optimization at the critical sections. eg. Why are we shifting the buffers in at the 4th position or so when we could totally shift the last pixel in from the right for little endian code without checking `(len(self.shift_buffer)) -1` code or whatever.  Or rather, we apparently do do that, but rather than `& 15` we calculate the mask every single time?

Usually hyper-optimizing doesn't matter but it in this section it might, you can show that it doesn't or simplify the code.

meerk40t/core/plotplanner.py
Sophist-UK: Agreed that this should be optimised.

PR #823, opened 2022-02-12
Transformation panel

jpirnay:
- Extend the functionality of the transformation panel by recognizing right clicks on the individual icon which will perform the same operation but with a bigger scale (i.e. the Rotate Icon will rotate the selected element by 5° when clicked with the left mouse button and by 90° when clicked with the right mouse button)
- Reordered  / tidied up the scale / skew / translate buttons to make them slightly more understandable (i.e. with more comprehensive tooltips)
1. You can now enter new values into the scale / skew / translate fields. 
2. The scale fields are been interpreted as factor to the original values, alternatively the can be provided as %age, so 0.5 or 50% will both have the same effect and reduce the size by half
3. The skew factors can be entered as pure numbers (which represent the trigonometric tangens of the to be applied angle) or as regular angle (i.e. 15deg, 30rad, 0.5 turn ...) which then will be applied
Notabene: the matrix still represents the regular [SVG transformation matrix](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform).
4. The translation value represents the distance in mils, but understands the usual length syntax (i.e 2in, 3cm, 12.5 mm...)
NB: The changes will only be applied if you press the Enter-Key in one of the fields
- Applied some fixes to sync the panel with the selection properly


meerk40t/gui/panes/navigationpanels.py
```diff
@@ -998,34 +1021,33 @@ def on_reset(self, event=None):  # wxGlade: Navigation.<event_handler>
...
+        angle = -90.0
+        self.rotateit(angle)
+
+    def on_rotate_cw_90(self, event=None):  # wxGlade: Navigation.<event_handler>
+        angle = 90.0
+        self.rotateit(angle)
 
     def on_text_matrix(self, event=None):  # wxGlade: Navigation.<event_handler>
         try:
+            # Not yet working as kernel reports a syntax error...
```
Sophist-UK, line 1041: Can you comment out lines 4107, 4130-4131 in elements.py and try to see what is creating a ValueError which is propagated as a SyntaxError?

Sophist-UK (changes requested):

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -998,34 +1021,33 @@ def on_reset(self, event=None):  # wxGlade: Navigation.<event_handler>
...
-        ]
-        x = input_driver.current_x if input_driver is not None else 0
-        y = input_driver.current_y if input_driver is not None else 0
-        self.context("rotate %fdeg %f %f\n" % (-5, x, y))
-        self.matrix_updated()
+        angle = -5.0
+        self.rotateit(angle)
 
     def on_rotate_cw(self, event=None):  # wxGlade: Navigation.<event_handler>
-        spooler, input_driver, output = self.context.registered[
```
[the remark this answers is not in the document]
jpirnay: Done

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -872,12 +882,19 @@ def __do_layout(self):
         grid_sizer_2.Add(self.button_rotate_ccw, 0, 0, 0)
         grid_sizer_2.Add(self.button_translate_down, 0, 0, 0)
         grid_sizer_2.Add(self.button_rotate_cw, 0, 0, 0)
+        # Add some labels to make textboxes clearer to understand
+        grid_sizer_2.Add(wx.StaticText(self, wx.ID_ANY, _("Scale")), wx.HORIZONTAL)
+        grid_sizer_2.Add(wx.StaticText(self, wx.ID_ANY, _("Skew")), wx.HORIZONTAL)
+        grid_sizer_2.Add(wx.StaticText(self, wx.ID_ANY, _("Translate")), wx.HORIZONTAL)
+        # First X
         grid_sizer_2.Add(self.text_a, 0, 0, 0)
-        grid_sizer_2.Add(self.text_c, 0, 0, 0)
```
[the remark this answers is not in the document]
jpirnay: Ther is still one minor flaw, but the life of me I can't figure out how to amend the tab order... Any pointers in the right direction would be appreciated

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -872,12 +882,19 @@ def __do_layout(self):
         grid_sizer_2.Add(self.button_rotate_ccw, 0, 0, 0)
         grid_sizer_2.Add(self.button_translate_down, 0, 0, 0)
         grid_sizer_2.Add(self.button_rotate_cw, 0, 0, 0)
+        # Add some labels to make textboxes clearer to understand
+        grid_sizer_2.Add(wx.StaticText(self, wx.ID_ANY, _("Scale")), wx.HORIZONTAL)
```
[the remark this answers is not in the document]
jpirnay: Done

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -815,12 +815,22 @@ def __init__(self, *args, context=None, **kwds):
...
-        self.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix, self.text_b)
-        self.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix, self.text_d)
-        self.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix, self.text_f)
+        #self.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix, self.text_a)
+        #self.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix, self.text_c)
+        #self.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix, self.text_e)
+        #self.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix, self.text_b)
+        #self.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix, self.text_d)
+        #self.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix, self.text_f)
+        self.text_a.Bind(wx.EVT_TEXT_ENTER, self.on_text_matrix)
```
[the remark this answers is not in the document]
jpirnay: Done

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -842,9 +852,9 @@ def __set_properties(self):
         self.button_translate_left.SetToolTip(_("Translate Left"))
         self.button_reset.SetToolTip(_("Reset Matrix"))
         self.button_translate_right.SetToolTip(_("Translate Right"))
-        self.button_rotate_ccw.SetToolTip(_("Rotate Counterclockwise"))
+        self.button_rotate_ccw.SetToolTip(_("Rotate Counterclockwise by 5° / by 90° on left / right click"))
```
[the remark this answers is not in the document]
jpirnay: Done

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -842,9 +852,9 @@ def __set_properties(self):
...
         self.button_translate_down.SetToolTip(_("Translate Bottom"))
```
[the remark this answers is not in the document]
jpirnay: Done

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -950,24 +966,31 @@ def on_scale_down(self, event=None):  # wxGlade: Navigation.<event_handler>
...
-        self.context(
-            "scale %f %f %f %f\n"
-            % (
-                scale,
-                scale,
-                x,
-                y,
-            )
-        )
+        self.context("rotate %fdeg %f %f\n" % (angle, x, y))
```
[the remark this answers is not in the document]
jpirnay: Done

meerk40t/gui/panes/navigationpanels.py
jpirnay: See [PR#824](https://github.com/meerk40t/meerk40t/pull/824) - I was splitting these two PRs intentionally, as this PR is more of a cometic nature (although it fixes some nasty bugs as well, like the transformation matrix not updated after dragging / resizing etc.). If you want me to combine these then let me know, as this PR is rather useless without the aforementioned.

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -923,12 +945,15 @@ def update_matrix_text(self):
...
-            self.text_e.SetValue(str(matrix.e))
-            self.text_f.SetValue(str(matrix.f))
+            # You will get sometimes slightly different numbers thean you would expect due to arithmetic operations
+            # we will therefore 'adjust' those figures slightly to avoid confusion by rounding them to the sixth decimal (arbitrary)
+            # that should be good enough...
+            self.text_a.SetValue('%.6f' % matrix.a) # Scale X
+            self.text_b.SetValue('%.6f' % matrix.b) # Skew Y
+            self.text_c.SetValue('%.6f' % matrix.c) # Skew X
+            self.text_d.SetValue('%.6f' % matrix.d) # Scale Y
+            self.text_e.SetValue('%.1f' % matrix.e) # Translate X, thats in mils, so about 0.25 mm...
```
Sophist-UK, line 955: There are about 40 mils in 1mm (1000 mils in 25.4mm), so 1 mil is c. 0.0254mm.

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -957,12 +982,10 @@ def scaleit(self, scale):
         x = input_driver.current_x if input_driver is not None else 0
```
Sophist-UK, line 982: x and y are no longer needed.

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -882,31 +892,43 @@ def __do_layout(self):
...
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("X:")), wx.HORIZONTAL)
+        grid_sizer_3.Add(self.text_a, 0, 0, 0) # Scale X
+        grid_sizer_3.Add(self.text_c, 0, 0, 0) # Skew X
+        grid_sizer_3.Add(self.text_e, 0, 0, 0) # Translate X
         # Then Y
-        grid_sizer_2.Add(self.text_d, 0, 0, 0)
-        grid_sizer_2.Add(self.text_c, 0, 0, 0)
-        grid_sizer_2.Add(self.text_f, 0, 0, 0)
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("Y:")), wx.HORIZONTAL)
+        grid_sizer_3.Add(self.text_d, 0, 0, 0) # Scale Y
```
Sophist-UK, line 909: Boxes are added row 1 left to right then row 2.

d needs to be after b.

Sophist-UK (changes requested):

tatarize (approved):

tatarize:
I corrected the ordering of #824 and merged it, the code here was the only code that used it and it's set in the normal svg ordering so no since introducing inconsistencies by fixing matrix to use the other somewhat reasonable ordering as such.

It's correctly set to want sx, kx, ky, sy, tx, ty now.

tatarize:
Merged in the main branch to test the matrix changes, and yeah, that all seems to work perfectly fine. The Transform bit there has always been the weakest and least helpful so making it more helpful does a lot to justify its existence.

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -957,12 +982,10 @@ def scaleit(self, scale):
...
         y = input_driver.current_y if input_driver is not None else 0
         self.context(
-            "scale %f %f %f %f\n"
+            "scale %f %f \n"
             % (
                 scale,
-                scale,
-                x,
-                y,
+                scale
```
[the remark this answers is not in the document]
tatarize: Style-wise we should likely just apply black and call it a day.

As code thing it reduces bickering over style bits. Since it always transforms everything into the same thing anyway.

jpirnay:
- Applied some cosmetic changes to the layout
- Added angle conversion to skew fields, so you may provide an angle, but it will then be represented as its tan
- 'Blacked' the code

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -36,6 +36,10 @@
 )
 from meerk40t.gui.mwindow import MWindow
 from meerk40t.svgelements import Length, Group
+from math import (
```
Sophist-UK, line 39: If you are willing can you address this codacy issue in your PR?

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -882,31 +892,43 @@ def __do_layout(self):
...
-        grid_sizer_2.Add(wx.StaticText(self, wx.ID_ANY, _("Translate")), wx.HORIZONTAL)
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, ""), wx.HORIZONTAL)
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("Scale")), wx.HORIZONTAL)
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("Skew")), wx.HORIZONTAL)
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("Translate")), wx.HORIZONTAL)
         # First X
-        grid_sizer_2.Add(self.text_a, 0, 0, 0)
-        grid_sizer_2.Add(self.text_b, 0, 0, 0)
-        grid_sizer_2.Add(self.text_e, 0, 0, 0)
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("X:")), wx.HORIZONTAL)
```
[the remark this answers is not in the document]
Sophist-UK: I think you need to adjust the `matrix_sizer.ADD(grid_sizer_2/3` statement to include a bit of left margin on grid_sizer_2.

Or maybe make grid_sizer_2 4x3 also with the same sized first (empty) column.

meerk40t/gui/panes/navigationpanels.py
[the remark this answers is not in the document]
Sophist-UK: I have a feeling that the only way to solve this is to make each column a separate vsizer and then add these four vsizers to the grid_sizer_3.

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -933,104 +1029,148 @@ def matrix_updated(self):
...
+
+    def on_rotate_ccw_90(self, event=None):  # wxGlade: Navigation.<event_handler>
+        angle = -90.0
+        self.rotateit(angle)
+
+    def on_rotate_cw_90(self, event=None):  # wxGlade: Navigation.<event_handler>
+        angle = 90.0
+        self.rotateit(angle)
+
+    def skewed_value(self, stxt):
```
Sophist-UK, line 1120: Because `self` is not used inside this method, codacy is recommending that you add the `@staticmethod` decorator immediately before and remove `self` from the parameters.

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -933,104 +1029,148 @@ def matrix_updated(self):
...
+                valu = tan(angle)
+            elif stxt.endswith("turn"):
+                angle = tau * float(stxt[:-4])
+                valu = tan(angle)
+            elif stxt.endswith("%"):
+                angle = tau * (float(stxt[:-1]) / 100.0)
+                valu = tan(angle)
+            else:
+                valu = float(stxt)
+        except:
```
Sophist-UK, line 1143: An `except` statement without one or more exceptions explicitly listed is really not a good idea.  

I would remove the try/except from here and wrap it around the statements that call `skewed_value` below - and if there is an exception, NOT do the transformation but instead indicate an error someone by e.g. red colour.

Sophist-UK (changes requested):
We are really getting close to a great PR here. Keep going.

Sophist-UK:
P.S. Can you update the very first comment box at the top to include all the improvements that are now in this.

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -872,24 +944,43 @@ def __do_layout(self):
...
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("Skew")), wx.HORIZONTAL)
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("Translate")), wx.HORIZONTAL)
+        # First X
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("X:")), wx.HORIZONTAL)
+        grid_sizer_3.Add(self.text_a, 0, 0, 0)  # Scale X
+        grid_sizer_3.Add(self.text_c, 0, 0, 0)  # Skew X
+        grid_sizer_3.Add(self.text_e, 0, 0, 0)  # Translate X
+        # Then Y
+        grid_sizer_3.Add(wx.StaticText(self, wx.ID_ANY, _("Y:")), wx.HORIZONTAL)
+        grid_sizer_3.Add(self.text_d, 0, 0, 0)  # Scale Y
```
Sophist-UK, line 961: I think this needs to come after text_b.

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -872,24 +944,43 @@ def __do_layout(self):
...
+        grid_sizer_3.Add(self.text_b, 0, 0, 0)  # Skew Y
```
Sophist-UK, line 962: I think this needs to come before text_d.

jpirnay: We need to take a decision here:
A) If we want to mirror the SVG matrix definition then the fields need to be:
SC_X SK_X TL_X
SK_Y SC_Y TL_Y
(as the wey were previously) - then we should add another row below Y to show "Skew Scale Translate" or omit the headers row to avoid confusion
B) If we want to hide the internal matrix representation and make the fields loook more 'logically' to your average Joe/Jane (and where the majority mous probably has no clue about SVG matrices in the first place, and I count myself to that group as of a couple of days ago...) then we should keep the representation as in this PR

Well, my preference is B)   ;-)

Sophist-UK: My vote is 👍 for option B).

tatarize: Well, the fields match the SVG provision. We don't *really* expose this fact, the `matrix` command is, after all a console command. It's fairly well hidden. The text boxes in navigation give that information but it doesn't much matter a it usually doesn't come up. The SVG style for the matrixes matches the wxPython style for the matrixes. For example when you pinch and zoom in the scene, that's using the SVG style matrix to do that and transfers that over to wxPython matrix things in the zmatrix or whatever that is, and it's just the same values in the same order.

tatarize: There's not really that much of a *logical* order to those values if they aren't a direct rotation or have a skew or something or sometimes are scaled and rotated. The values of the matrix are sometimes helpful to know but mostly you'd need to make some guesses when decomposing the matrix into a series of scale/rotate/translate operations. Your skew values for angles get a little weird when you're moving the points around, and you just can't quite read it as independent operations in a way that is helpful even when you know the math behind it.

tatarize (approved):

meerk40t/gui/panes/navigationpanels.py
jpirnay: I agree the matrix gets pretty weird soon, still it has become my favourite tool to exactly place / scale something (I do a lot of on-the-fly rects...) and there it helps :-)

meerk40t/gui/panes/navigationpanels.py
jpirnay: On the text-element thingy... I would prefer to detach the matrix-textfield part from the graphical button part and expose it as a panel of it's own. This would bring more flexibility IMHO. Happy to do it (in a separate PR).

tiger12506:
I did a quick test of this, and it is looking awesome!

Unlock pane, undock, and note that minimum size is too small, the textboxes get clipped.

Likely this will be different on different platforms and needs testing.

jpirnay:
@tiger12506 On Windows there's no such effect, everything is fine. I assume you are referring to MacOs? Could you provide me with a screenshot please?
(Just compiling wxPtyhon on a Linux machine to see how it behaves there)

tiger12506:
@jpirnay This is the minimum size that I see on Windows 10, 4K screen with 150% DPI scaling (maybe that's the difference?)

image

jpirnay:
@tiger12506 Great, thank you - I can reproduce this behaviour on my Linux machine as well:
But this seems to be a problem that is true for the other two panels as well... 
grafik

meerk40t/gui/panes/navigationpanels.py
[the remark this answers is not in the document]
Sophist-UK: Alignment is still a (minor issue).

image

I think you either need to put the transformation matrix into a separate panel as you previously suggested, and then align the panels in the Navigation window, or combine the two grid sizers for the buttons and the matrix into a single 4x6 grid sizer.

Also, the 0.000000 skew value is too wide for the box - I am not sure if you can reduce the inside margins of the box or whether you need to make the boxes slightly wider.

Sophist-UK:
> @jpirnay This is the minimum size that I see on Windows 10, 4K screen with 150% DPI scaling (maybe that's the difference?)
> 
> image
> 
> And on my Linux machine: image

This is actually a broader issue of minimum pane sizes and I am not sure whether we should allow the user to reduce size below that needed to display everything or not.

But the **default** pane and window sizes need to show everything.

Sophist-UK (dismissed):
Three new issues:

1. Should we execute the transformation when the user leaves a field, or just on enter?
2. Do we need an execute button to go with the input fields as we have on the position and move-to fields?
3. Images in the enclosed test file disappear to the top left corner when I execute a transformation. (I note that they also do this on a `reset` console command.)
[Element classification test.zip](https://github.com/meerk40t/meerk40t/files/8079644/Element.classification.test.zip)

tatarize (approved):

tatarize:
@Sophist-UK We should just merge this, as generally done. None of the defects are due to any changes made by jpirnay. If we opt to fix the other bugs which are in the pane sizes elsewhere for the FloatSize() or whatever then that can be done in another PR. 

This code here breaks nothing that wasn't already broken. And fixes a number of things.

meerk40t/gui/panes/navigationpanels.py
```diff
@@ -933,104 +1044,154 @@ def matrix_updated(self):
...
+    def on_rotate_ccw_90(self, event=None):  # wxGlade: Navigation.<event_handler>
+        angle = -90.0
+        self.rotateit(angle)
+
+    def on_rotate_cw_90(self, event=None):  # wxGlade: Navigation.<event_handler>
+        angle = 90.0
+        self.rotateit(angle)
+
+    @staticmethod
+    def skewed_value(stxt):
```
tatarize, line 1136: This code could just be replaced with `Angle.parse(stext)` but other than that this whole thing is fine. And I'd be happy to merge it. What errors exist did so before. This is certainly a net gain.

tatarize: Well, actually given you accepted the whole meerkitten invite. I'd be happy to see it merged, you could merge it yourself.

Sophist-UK:
> @Sophist-UK We should just merge this, as generally done. None of the defects are due to any changes made by jpirnay.

This PR is certainly functional however (to be factually correct) some of the minor visual issues were in fact due to changes made by Jens. The addition of X and Y to the left of the transform input fields was the cause of the alignment changing and the boxes moving closer together etc. I don't say this to point any finger of blame, just to correct the facts.

I genuinely believe that we should encourage PRs to be as perfect as possible, subject to the author being willing to make them so, because otherwise we are storing up minor issues that need to be fixed later and because IME minor UI imperfections disproportionately impact the reputation of software (i.e. users look at the UI imperfections and believe that the core functionality is equally imperfect). 

But if the author is not willing to do more (which is their choice - we are all volunteers) and no one else is willing to fix it, and the choice is therefore between merging something with minor issues or losing all the work that has been done, then I certainly would prefer to see the PR merged and **not** to see the effort wasted.

Sophist-UK:
I have been playing around with your code to see how easy it would be to make it do the transform on a tab. I have a version of the code which I would like to share with you. Do you have an email I can use? Or can I encourage you to join our Discord server?

jpirnay:
a) There were some minor tweaks and code cleaning applied (thank you @Sophist-UK - appreciated :-) )
b) As there is unused screen estate in the transformation window I added a resize panel - feedback appreciated

Sophist-UK:
> As there is unused screen estate in the transformation window I added a resize panel - feedback appreciated

If you need more real-estate to deliver useful functionality then just annex your next door neighbours field.

I will get to reviewing this when I can - perhaps this evening.

tatarize (approved):

Issue #719, opened 2021-12-27
Allow scene selection of elements whilst avoiding unwanted moves.


Sophist-UK:
As I understand it there has been a deliberate decision to only select the entire scene with a mouse click to avoid unwanted selections and moves. Individual elements (or multiples) can now only be selected using a selection drag-box.

The problem is that this makes it much more difficult to select elements and use right-click functions to change them - and this is one of MK's most powerful features. Ideally we would find another way to prevent moving subsets of the whole without this being a deliberate choice by the user.

With `Undo` functionality this may become unnecessary.

Another possible approach is that if only part of the design is chosen, moves are disabled by default and you have to select Move Enable in a right-click menu to set a flag to enable moves.

tiger12506:
jpirnay's #825 should help this situation, but it's not enough. It's not possible to select more than one item from the scene that are not adjacent, even with #825. Other programs handle this with holding shift or control (Windows general is ctrl, Inkscape uses shift, etc.)

With #825, and above paragraph fixed, it would be possible to actually use selection in scene, but it would still be nice to see the "click = select all" gone. It should become a preference item default on.

jpirnay:
I believe this can be closed now: with [PR1058](https://github.com/meerk40t/meerk40t/pull/1058) you can select a single element with a single click and maintain the existing selection if you are holding down the shift-key.

PR #825, opened 2022-02-12
Visually distinguish selection modes

jpirnay:
Make the different selection mechanisms more visibly distinguishable by introducing different color and selection schemes.

### Selection
I am distinguishing now between the different sectors of mouse movement and allow every single sector to have a different selection criteria (touch, surround). There is a draft visual separation between these.  In addition a short helptext is displayed in the Status-Bar of the main-window.
grafik

The implementation as in this PR establishes
Touch     | Surround
---------+------------
Touch    | Surround

- If you press the Shift-Key while selecting then the selected elements will be added to the previous selection 
grafik

- If you press the Ctrl-Key while selecting the selected element will flip their selection status (use case: select a bunch of element first then deselect them with Ctrl+Selection)
grafik

### Alt+Drag = Copy
I have added an experimental copy feature: if you move a selection while pressing the Alt-Key then a copy is being created. NB: the Alt-Key is only checked at the time of the leftdown event, so any keyboard-events while dragging (i.e. a later alt or a release of alt before the end of the drag) are ignored


meerk40t/gui/scene/scenewidgets.py
```diff
@@ -668,6 +668,12 @@ def process_draw(self, gc):
             y0 = self.start_location[1]
             x1 = self.end_location[0]
             y1 = self.end_location[1]
+            # Determine Colour on selection mode: standard (from left top to right bottom) = Blue, else Green
```
Sophist-UK, line 671: I am unclear about:
a. What the difference in selection modes is TL->BR from all three other directions?
b. Why we only have TL->BR and not 2 of the 4 directions?

jpirnay: As far as I can see there is a different selection scheme: in leftup you establish whether a object to be selected neds to be fully encompassed by the selection rectangle (right_drag = sx <= ex and sy <= ey) or whether it suffices that the selection rectangle overlaps the boundary of the object (and i believe there is still a bug there)
The color selection just follows this logic

jpirnay: Okay, fixed intersection check

Sophist-UK: There are 4 possible directions to draw the rectangle - AFAICT, two do one type of selection, and two do the other.

Sophist-UK: I have just tested this code and it seems actually to select differently to before. I think that the way it used to work was better so please revert.

Sophist-UK (changes requested):

jpirnay:
@Sophist-UK I am happy to revert the intersection check (= the non-standard select) although I have to admit I cannot really get the logic behind the algorithm. 
From a different piece of software (LB - [link to docu](https://lightburnsoftware.github.io/NewDocs/ZoomPanSelect.html)):

> The red rectangle will disappear when you release the mouse button, and all the shapes enclosed within it will be selected. This is called an enclosing selection - only things fully contained in the red enclosing rectangle will be selected.
> 
> You can drag from right to left instead, and this will create a green rectangle, which will select anything that it crosses:
> 
> DragSelect-Crossing
> 
> In this case, all three items will be selected even though they are not fully contained by the selection rectangle.
I was assuming Meerk40t would do the same, but actually I still need to figure out when it selects items while dragging the selection rectangle.
Addmittedly there is one flaw in my current implementation, it should just look for the left to right selection on ignore the up/down element - I will fix that for consistency.

Sophist-UK:
@jpirnay Jonathan - I am very happy with the concept of the different selection methods, and when I try it on `main` these appear to work in a similar way (though I haven't tested that they are exactly like Lightburn). But when I tested your PR it had stopped working this way.

meerk40t/gui/scene/scenewidgets.py
```diff
@@ -618,7 +618,7 @@ def event(self, window_pos=None, space_pos=None, event_type=None):
                 sy = self.start_location[1]
                 ex = self.end_location[0]
                 ey = self.end_location[1]
-                right_drag = sx <= ex and sy <= ey
+                right_drag = sx <= ex # and sy <= ey
```
tatarize, line 621: Without sy <= ey, isn't this purely right drag. So you right click draw a purely horizontal line to the left, and everything in the scene within that line horizontally would be selected? This seems weird. I was told other graphics utilities used this sort of right-up drag function to select objects and left-down as a functional and useful difference but I've not really heard of it without also checking y. That seems like it would be weird.

jpirnay:
@tatarize I've tried to demonstrate my problems with the current algorithm: [Quick screen recording](https://youtu.be/cvcNX3WtQFo)
I would have expected the current implementation to select an element regardless of the statrting point / shape of the sleection rectangle as long as it crosses / touches the elements boundary. Alas this is not the case. So I  tried (mostly for my own benefit) to make the process more appealing to my logic and show when which method would be used.

meerk40t/gui/scene/scenewidgets.py
tatarize: Yes, we treat selection going BR->TL differently. Rather than needing to select the entire outlines it only needs to cross the path of the bounding box.

tatarize:
@tiger12506 You also have some notably complaints with the selections in MK especially with regard to groups. The above video does a pretty good job highlighting other seeming issues. Should we maybe revisit this thing overall or just look for a better way to fix this and add a checkbox for your whole select individual things complaint?

meerk40t/gui/scene/scenewidgets.py
tatarize: Tested in actual code. This actually works pretty well.

tatarize (dismissed):

Sophist-UK:
I think there needs to be a discussion about the best approach to meeting the multiple objectives for the second PR.

a. Make selection of individual elements in scene easy.

b. Make selection of all elements equally easy.

c. Prevent accidental moves of only some elements e.g. when you move something and don't realise that you don't have anything selected or e.g. when you nudge the mouse with your elbow combining a click on a random object with a drag move.

(And I defy anyone to say that the last example hasn't happened to them at some point in their life!!!!)

jpirnay:
This is not yet for publication as we need to agree on the selection method. 
I am distinguishing now between the different sectors and allow every single sector to have a different selection criteria (touch, cross, surround). There is a draft (and ugly) visual separation between these. I would have hoped for a way to change the main-statusbar - but I have not found a way to do so without touching more code (and most probably I have missed something) - that would be hepful for those with color-identification challenges to have a message in the statusbar as well describing the selection method.

The implementation as in this PR establishes
Cross | Surround
------+---------
Touch| Surround

and that's most probably too confusing to have three different ways of selecting. I just wanted to have all three present to help experiment and decide which ones to choose.

Sophist-UK:
@jpirnay Jonathan - It is fine to experiment. You can mark this as "draft" to show that it is not ready for merge, and you can request reviews when you want some input about the different types of selection mechanism.

jpirnay:
@Sophist-UK Thank you, for introducing me to the finer details of github workflows. And btw it's Jens not Jonathan - although related ;-)

tiger12506:
I love this idea as a power user. There are established precedents in other programs which do this... Fusion360 is a big one.

I think most people who are observant enough to recognize this behavior will be most familiar with Touch for II and III quadrant, and Surround for I and IV quadrant. I don't know that there are any advantages to Cross...

That said... this code seems quite buggy to me and I can't get Cross or Touch to behave correctly. Please see attached file and try to Cross or Touch only one hexagon. @jpirnay 
Aristotle-Hexagons-meerk40t-dev

tiger12506:
@tatarize I agree with @Sophist-UK that the select-all needs to be a separate pull request/issue from this one. This stands well enough on it's own as a concept.

jpirnay:
@Sophist-UK Thanks for the file and yes there was a bug - fixed. Please have another look at it.

tiger12506:
@jpirnay Yes, looks like the bug has been fixed and the functionality is as described. I believe that my earlier opinion still holds for me, that quadrant II and III should be touch and quadrant I and IV should be surround. I don't care for cross, but I my opinion could certainly be overridden if anyone wishes.

jpirnay:
Update:
- I've amended the modes according to the feedback of the reviewers
- New: Touch left down will leave the previous selection untouched, while touch right down discards the previous selection (default behaviour) - for the time being no visual representation is associated but would be simple to implement
(and I have updated title and comment on the top)

jpirnay:
On a sidenote: I was tempted to import pynput or keyboard to listen to the shift key, but that would require afaik a couple of root rights at least on Darwin and maybe Linux as well...

meerk40t/gui/scene/scene.py
```diff
@@ -141,6 +144,20 @@ def signal(self, *args, **kwargs):
...
+            self.scene.isShiftPressed = True
+        if keycode == wx.WXK_CONTROL:
+            self.scene.isCtrlPressed = True
+
+    def on_key_up(self, evt):
+        keycode = evt.GetKeyCode()
+        if keycode == wx.WXK_SHIFT:
+            self.scene.isShiftPressed = False
+        if keycode == wx.WXK_CONTROL:
+            self.scene.isCtrlPressed = False
```
Sophist-UK, line 159: My gut reaction is that this is **NOT** the way to go. IMO the raw event that wx executes should be queried to find out what modifiers are pressed **at the time of the leftup event** (and every other event) and these modifiers should be passed as parameters to the various classes that processes the event. There is also a routine in wxutil which you should use to check for modifiers because there are some variances between o/s and this routine handles these.

meerk40t/gui/scene/scenewidgets.py
```diff
@@ -583,12 +583,38 @@ class RectSelectWidget(Widget):
...
+        wx.GREEN,
+        wx.BLUE,
+    ]
+    # Visual help for selection rectangle (to help those with visual challenges i.e colorblindness)
+    selection_line = [wx.PENSTYLE_DOT_DASH, wx.PENSTYLE_DOT, wx.PENSTYLE_SHORT_DASH]
+    selection_text = [
+        "Select all elements that the selection rectangle touches",
+        "Select all elements that the selection rectangle crosses in at least one dimension",
+        "Select all elements that the selection rectangle fully encloses",
+    ]
```
Sophist-UK, line 599: 1. Please translate these strings i.e. `_("text")` rather than `"text"` .
2. I like that these strings exist. Check the wxmain.py file as there is now a method by which you can send this to the status bar when the selection box is started and you can determine the type of selection based on the direction of movement (though I guess this can change if you start going NW and then go East and turn it into a NE drag. 
3. Can we please have versions for Control and Shift (which need to react if these are pressed or released whilst left-down?
4. There are three strings here, but I can see only two modes (RTL and LTR) when I test.

meerk40t/gui/scene/scene.py
Sophist-UK: Also, I think that Ctrl-drag and Shift-drag and Ctrl-Shift-drag are all valid actions, and we should decide what each of these does. I have a feeling that Shift is usually additive, Control is usually toggle.

Sophist-UK (changes requested):
This PR is getting better and better, but it still has a few things that need improving.

In addition to the above comments, there is one major functional issue: in the enclosed test file, try selecting nothing and then a red RTL selection of the middle emblem. Instead of it selecting every element that the red-selection box touches in the emblem, it selects everything and this doesn't feel right to me.
[Element classification test.zip](https://github.com/meerk40t/meerk40t/files/8079433/Element.classification.test.zip)

Also, I know this PR is about the selection box, but there is another issue which I think is similar to the above, and which relates to #719. I can click anywhere inside the outermost bounds of the design and it selects everything, even if I don't click a specific element and instead click the background. This is counter-intuitive, **but** it could be used to advantage i.e. clicking on background selects all, clicking on an element selects that element (or maybe something even more intelligent - clicking on an element selects that element and anything that has a fill and overlaps (recursively) or any element without a fill whose path contains or crosses any of the filled elements (recursively).

But I say again, this is really going somewhere - so please keep going.

Sophist-UK:
> @tatarize I agree with @Sophist-UK that the select-all needs to be a separate pull request/issue from this one. This stands well enough on it's own as a concept.

See my recent comment - I now believe that there is cross-over between #719 select all and this PR.

meerk40t/gui/scene/scene.py
jpirnay: I get your reaction to the treatment of Ctrl & Shift - as stated that has been a quick and dirty attempt to see whether this is beneficial or not (and after having used now for some hours I personally like it a lot) and needs to be refined. It's definitely not up to MK code standards yet.

meerk40t/gui/scene/scenewidgets.py
jpirnay: > There are three strings here, but I can see only two modes (RTL and LTR) when I test.
The current test does no longer use cross-selecting, as the feedback so far indicates that touch is more intuitive.

Sophist-UK: > The current test does no longer use cross-selecting, as the feedback so far indicates that touch is more intuitive.

That's fine - we either need to leave the code to do the cross-select in the code base but commented out (if we think we might ever want to use it) or remove all traces including the string.

jpirnay: >In addition to the above comments, there is one major functional issue: in the enclosed test file, try selecting nothing >and then a red RTL selection of the middle emblem. Instead of it selecting every element that the red-selection box >touches in the emblem, it selects everything and this doesn't feel right to me.
Actually it does what it is supposed to do ;-) It selects all these elements but as well as it overlaps with the big box outside that one as well, so the biggest surrounding box is used to highlight the selection.
On second thought i think it would be better if we ignore selections fully inside a bigger object - have changed the code accordingly
I beleive though it might be more log

Sophist-UK: I'll wait to see the code to see if it is a good algorthm.

tiger12506:
@Sophist-UK You're very good at taking simple positive changes and turning them into very large unwieldy scope-creep changes that ultimately have great difficulty in going through.

We already have established in the past that "click = select all" is a big disagreement and sticking point, and I don't see why it's necessary to drag it into this topic. There are definite changes in this pull request that are great and would help, even if they are developed iteratively and shouldn't be held back.

There is a valid question though, of whether Touch should be selecting inner elements as much as possible, or outer elements as much as possible. I am of the strong opinion that Touch should select inner elements, and Surround should select outer groupings because that is what would make each feature the most useful. @jpirnay 
Note: I haven't tested latest commits.

tiger12506:
@jpirnay Also, as I have stated before, @tatarize or @joerlane needs to weigh in with any modifier changes that are being made. Shift and Ctrl key modifiers in the scene have cross-platform considerations to be made, and are already pretty overloaded as it is. Does shift/ctrl modifying view scrolling still work with these changes?

jpirnay:
Okay, a major update (at least for my limited python / MK experience):
Internal changes: 
The KeyPress-Events are being dealt with as any other event in the Scene and passed down the widgets list. The only one listening now to that is the RectSelectWidget, but could be extended to others (i.e. handling Ctrl-Key plus translate = translated copy)
Selection-Wise:
The strange behaviour for partial selects should be fixed (as stated above). For fun I have added a Control-Modifier that inverts the selection state of all hit elements
I could definitely use some help to tell me how to address the wxmain.main_statusbar with in the widget classes... 
(Everything would be so much simpler if there would be a command in the console/kernel that would set / reset the StatusPanel 0)

jpirnay:
@tiger12506 On the selection bit within a given selection: try to press the Alt-Key in parallel to the Ctrl or Shift Key. This will bypass the move-Action of the Selection Widget

tiger12506:
Ok, but why can't holding Ctrl or Shift do that directly?

Sophist-UK:
> @Sophist-UK You're very good at taking simple positive changes and turning them into very large unwieldy scope-creep changes that ultimately have great difficulty in going through.

@tiger12506 - and there is certainly a valid argument that this is the case, and if I am brutally honest with myself, I am often guilty of throwing too much into my own PRs. But on this occasion I don't think this is the case. I was against including the select-all issue right up until I tested this and discovered that there was an issue with this drag selection inside a bbox of a group or file or element branch node, which is pretty much the exact same issue as the click to select issue - and at that point it seemed a possibility that there was some synergy. **If** there is some synergy then it makes sense to include it, and if no synergy then not. The rest of what you are calling scope creep is IMO more about my perfectionism - at this stage I think comments are a good thing, and the quality is great as a consequence, but if it ever gets to the point that it gets stuck for any reason, then we always have the option of merging it and iterating from there. There are often downsides to multiple small iterative PRs rather than iterating a single PR - one is that you risk confusing users if a release is spun off part way through this process, another is that you are much more likely to get a piecemeal end solution.

jpirnay:
@tiger12506 If you have some time pls check again - there's no need for the alt-key anymore - on the othe hand if you want to move a selection you need to click and hold still for about 0.3 seconds - only then will the move start. There might be a better way but that was the best I could think of

tiger12506:
@jpirnay - I took a look at this as requested. I'm not sure I understand the need for the 0.3 seconds, I'd rather it be decided based on whether or not the Ctrl/Shift is held immediately -- but I'm not familiar with how the mouse events are structured in meerk40t, tbh (I'm guessing meerk40t calcs the decision to drag happen first and then checking ctrl/shift? I believe it should be the other way around, key modifiers checked first!)-- so that may be difficult to achieve. If it were @tatarize, I'd badger him to do the refactoring that would be needed to support that use. Maybe he has an opinion.

It seems like the current commit causes a bug with the move feature. To reproduce, do a Surround selection and try to drag to move it.
(Although the Touch selection seems to work great)

@Sophist-UK We're already plenty deep in it now. Might as well get what you want out of it. Could have just been a simple positive change, now it's a whole feature with edge cases and bugs and potential reworking how meek40t works etc etc etc. As with all issues you weigh in on. Sigh. I liked it before the pomp and circumstance. Fast iteration was better. Talking is annoying.

Sophist-UK:
@tiger12506 said: "I'd rather it be decided based on whether or not the Ctrl/Shift is held immediately"

IME with other software, the action is normally chosen at the end on mouseup, but it is also used when you handle a mouse move during the drag. (In other software like Windows file explorer, you e.g. get a different mouse icon depending on what modifiers you have down during the drag, and if you change these modifiers mid-drag, then the icon changes. Here you get a different colour or design of the selection box, but the principle is the same.) In theory you should note if the modifiers change without moving the mouse (i.e. process the modifier keydown/keyup changes as well as noting their status during a drag mouse move and at mouseup, but this is not essential and in the interests of avoiding scope creep could be left out.

tatarize:
The mouse events in wxPython store the keyboard modifications. This is why zoom holding shift acts differently. 

```python
        if event.GetWheelAxis() == wx.MOUSE_WHEEL_VERTICAL and not event.ShiftDown():
            if event.HasAnyModifiers():
                if rotation > 1:
                    self.scene.event(event.GetPosition(), "wheelup_ctrl")
                elif rotation < -1:
                    self.scene.event(event.GetPosition(), "wheeldown_ctrl")
            else:
```

Assuming these keypresses don't need to get caught earlier in the scene to process the hotkeys it's acceptable to hand these events off to the widgets. It seems like jpirnay has a pretty reasonable understanding of what's happening there.

The typical modifiers for keyboard tend to already exist stored in the mouse events. Though the mouse events aren't handed down to the hit-chain, these are currently given an event name with the position as such. There's some events that include a space and some extra information (`magnify` for example) so it wouldn't be too weird to hand it "kb_down shift" and "kb_up shift". However, it might be preferred to just hand it all of the keycodes in those cases, since what we're doing in that case is handing the data down to the widgets. While this upgrade might not care, other widgets that people write might find this information useful.

---

Though not documented very well, there is a fantastic set of widget/scene visual api setup that makes programming the GUI extremely easy. The scene does events at particular points in the scene. The position information is then used to calculate the hitchain. This is the list of widgets being hit by this event. So that when you select an element in the scene and click and drag it, you get a different result if you're moving an object, using the selection widget, or if you're drawing a rectangle or circle (both work in 0.8.0). It does this by calculating the hitchain set of widgets which contained within that area and calling their event routine which will process the event widget by widget until one actually deals with that event.

The widgets can call `HIT`, `DELEGATE`, `DELEGATE_AND_HIT` or `HIT_AND_DELEGATE` depending on whether they should be added to the hit chain, or if they should be added but only after all of the children etc. Then the event calls the hitchain which has response:
*  `CONSUME`, eats the event
*  `ABORT`, kills this event and the entire chain -- Don't process anything else, movements or releases within the same associated events.
* `DROP`, removes the current widget from being considered in future associated event hitchains
* `CHAIN`, processed the events but lets additional widgets under it *also* process those events. 

The system here took a while to perfect and comes from an android app I wrote and is fairly robust. But, this will allow a widget to have a position, get delegated events based on position and process them. Which is why I spent a couple hours and managed to write a full set of rather useful tools in a fairly short period of time the other day. It's entirely possible for widgets to draw themselves and do anything you want. So if you wanted to make widgets that draw themselves or any sorts of visualizations that apply only to them or during particular events or animate things in the scene, all of this is rather easily doable. It's one of those super-powered features that isn't used to the full extent of it's abilities.

This brings us back to this PR which is great, but currently the key_down and key_up checks should potentially be checked as to whether or not this interferes with the binds or use the mouse events which already store those modifiers. Keep in mind this system is *not* set in stone. The **shift** key currently sends different **wheelup_ctrl** events (note: shift != ctrl) this is for panning rather than zooming. This events system is not perfect or highly refined by any stretch, it's very very powerful but the api here is not fully realized. This touches on the keyboard events and what widgets might need to do with them.

Take, for example, a desired feature that ties very closely with this PR. Control+Click to add selections. We can detect mouse events as to whether the control is held or not, at the time of the mouse event. So we don't have to complicate these event_types with things like `wheelup_ctrl` which, frankly, does not scale for all the events and all the different modifiers. We could add in a modifiers tag to event which isn't a giant refactor and does scale to all these events. But, also we could maybe have specific hotkey sequences that could be processed by the widgets themselves. So, for example, we could have widget-context sensitive hotkeys. Which could let you control the entire scene with the keyboard with basically infinite combinations, assuming the binds will be allowed to propagate from the main window (which I don't know if they would, off the top of my head) (though we could make a hotkey widget that would catch them all and run them normally if that was a required thing).

---

So do we need to know the kb events without the mouse events? These are position based anyway so the mouse clicks and positions generate the hitchain and keyboard events do not have a spacial location. jpirnay uses the last position in space to calculate these currently.

Should we add a modifiers to the `event` command to allow processing of these event types with potential modifiers but only for the mouse events?

Should we permit the fine-grain keyboard up and down events as part of the processed hitchain. So we'd be processing a `move` event with modifiers of holding the shift, or we'd get a keyboard `kb_down shift` event being sent if that keyboard key gets pressed during processing or we could do both. However, press shift, drag-select would not get a `kb_down shift` during the normal processing of the chain since it could occur before the processing of the event and be part of the hitchain for a `hover` event (movements without pressing any mouse button).

Should we try to expand the sequence to send all hotkey presses to the widgets. So control+b would do different things depending on the hitchain?

jpirnay:
Okay, I made a couple of changes without having read through all the valuable comments... So some of my changes may come across as if I have wanted to ignore the advice of those who have a way better understanding of whats happening inside all the Widget/Scene classes, which is definitely not the case.
In a nutshell, deltas to last time:
- No need for a wait to drag
-  Ctrl+Drag for a copy of the selected elements
- Therefore the modifiers for selections are shift for additional selection and alt for inverted selection

I really would like to make sure those changes are working as expected before adding additional bells and whistles (says the guy who has just added yet another feature)

tiger12506:
@jpirnay -- Personally I'd prefer to swap the alt, ctrl. I think ctrl is more strongly tied to individual selection (toggle feature) in Windows/Linux (at least). The drag to copy feature sounds good in theory, might have too many features in one place? We do have Ctrl-C,Ctrl-V.... but I haven't yet tested. If it works intuitively, then I'm for it.

@tatarize - Just "yes" to your style of thinking. You're my hero. Should the keyboard modifiers be accessible without mouse events? Yes. The statusbar messages are a great example of this. You hold shift, and the statusbar gives you helpful message that Shift+Drag = Additive selection, all without mousemove needing to happen.

Further, Ctrl+Click is an excellent workaround for 90% of my hatred for click = select all. All I have to do is hold Ctrl to disable the select all feature! Yes!

@Sophist-UK - I agree that this is getting out-of-hand. Someone else should comment on jpirnay's latest work in addition to me and we should merge it before iterating further.

Sophist-UK:
I have just tested this again.

1. I really like the + symbol in the corner next to the mouse - much better than the previous cross, but there are some margins inside the dashed blue/red line and the cross is not being completely shown (and indeed the - is missing a bit too).

2. With something already selected, and no modifier keys pressed, it seems to act like you have toggle/Ctrl pressed down. With toggle selected by default I cannot tell whether other odd results I am receiving are due to this or are separate issues. I will retest once this is sorted.

Sophist-UK:
> I agree that this is getting out-of-hand. Someone else should comment on jpirnay's latest work in addition to me and we should merge it before iterating further.

@tiger12506 - Sorry, but I disagree - getting something working perfectly is **not** feature creep - it is quality management. I believe this should be merged once it is working in a way that we are all happy with (or if Jens finds that they get stuck) - but not whilst it is not working in a way that is consistent with itself, consistent across various platforms and with generally accepted usability on these platforms, and consistent with other parts of MK.

Sophist-UK:
P.S. Is it just me who finds it annoying that the DIFFS includes Black modifications of entirely unrelated code?

tatarize:
@Sophist-UK it would be less annoying if we applied black more often or required that PRs pass black to be merged. Vpype has that as a thing, unless your code is unmodified by applying black you cannot merge it.

tatarize:
@Sophist-UK if you want I could apply black and isort to #831 which is all codecy fixes. And then we merge that into main and can merge those changes into PRs like this so it would not change those bits.

Sophist-UK:
> @Sophist-UK it would be less annoying if we applied black more often or required that PRs pass black to be merged. Vpype has that as a thing, unless your code is unmodified by applying black you cannot merge it.

@tatarize I think if we applied black automatically after every PR merge, then we would stop having these problems.

jpirnay:
@Sophist-UK 
> I really like the + symbol in the corner next to the mouse - much better than the previous cross, but there are some margins
> inside the dashed blue/red line and the cross is not being completely shown (and indeed the - is missing a bit too).

Fixed

jpirnay:
@Sophist-UK 
> However I am seeing different results on two different test files when (with nothing currently selected) selecting blank areas in the middle of several objects without enclosing even part of any objects. In one file everything is selected, in the other nothing.

Are you able to share these with me, please?

tatarize:
I merged in the black+isort from the main so it will no longer highlight changes coming from applying black but only real changes.

jpirnay:
Can we please agree on the Ctrl / Alt debate:
- Ctrl belongs to drag-copy, Alt to selection
- Alt belongs to drag-copy, Ctrl to selection

(There aren't any polls here, are there?)

Sophist-UK:
Alt on windows activates the main menu. So Alt-f for example, opens the File menu. In other languages it might be a different letter. Alt should probably not be used.

joerlane:
None of those options work on macOS. 

Option/Alt is used to drag copy in the OS, and many other places. Control will always bring up the right click menu though; even in MeerK40t. Apple/Command is used to select individual items (selection).

Sophist-UK:
> None of those options work on macOS.
> 
> Option/Alt is used to drag copy in the OS, and many other places. Control will always bring up the right click menu though; even in MeerK40t. Apple/Command is used to select individual items (selection).

@joerlane wx / meerk40t automatically translates the Mac Command key to windows/linux control, so when we talk about Ctrl you need to read it as Command.

However, we probably need to do something in tooltips and status-bar help to translate "Ctrl" to "Cmd".

joerlane:
Well; that needs to be fixed in Keymap too. Telling users things are set at control+ when it's not is pretty annoying. Just tested it.

Sophist-UK:
My bindalias PR which is in draft form does fix some Mac-related issues already. But yes, the Keybind UI needs to be fixed in this way. Leave that with me to add to the existing PR.

joerlane:
Alt belongs to drag-copy, Ctrl to selection:

Would be correct assuming the OS remapping difference for macOS.

Sophist-UK:
> Are you able to share these with me, please?

Here they are. But they may simply be a consequence of it showing the [-] with no modifiers pressed.

[Element.classification.test.zip](https://github.com/meerk40t/meerk40t/files/8096144/Element.classification.test.zip)
[Selection box test.zip](https://github.com/meerk40t/meerk40t/files/8096162/Selection.box.test.zip)

jpirnay:
> Here they are. But they may simply be a consequence of it showing the [-] with no modifiers pressed.

As far as I can see everything behaves as expected - as least by me ;-)
I will make one final commit to change alt and ctrl (or alt and command), so ctrl/command inverts the selection, alt+drag will copy the elements.

meerk40t/gui/scene/scenewidgets.py
```diff
@@ -680,6 +935,53 @@ def process_draw(self, gc):
             gc.StrokeLine(x1, y0, x1, y1)
             gc.StrokeLine(x1, y1, x0, y1)
             gc.StrokeLine(x0, y1, x0, y0)
+            matrix = self.parent.matrix
+            delta_X = 15.0 / matrix.value_scale_x()
+            delta_Y = 15.0 / matrix.value_scale_y()
+            # Draw very ugly indicator...
+            if self.key_shift_pressed:
+                if (
```
tatarize, line 943: You could probably just gc.StrokeText("+") or whatever.

meerk40t/gui/scene/scenewidgets.py
```diff
@@ -583,19 +655,72 @@ class RectSelectWidget(Widget):
...
 
+    # selection_method = 1 = hit, 2 = cross, 3 = enclose
+    # Color for selection rectangle (hit, cross, enclose)
+    selection_colors = [
+        wx.RED,
+        wx.GREEN,
+        wx.BLUE,
+    ]
+    # Visual help for selection rectangle (to help those with visual challenges i.e colorblindness)
+    selection_line = [wx.PENSTYLE_DOT_DASH, wx.PENSTYLE_DOT, wx.PENSTYLE_SHORT_DASH]
```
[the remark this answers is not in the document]
tatarize:
```python
SELECTION_TOUCH = 0
SELECTION_CROSS = 1
SELECTION_ENCLOSE = 2
```
It's mostly style to make sure the code is more readable for future people since code tends to get read more times than it's written.

tatarize (approved):
The `-` means toggle is a bit weird. Seems like printing a ^ there might be a bit more cryptic. but correct. 

Also, when you press control you take over the meaning of the mouse but the hover stuff still drills down to changing the mouse cursor, I don't love that.

Sophist-UK:
I have been playing around with your code to see how easy it would be to make it do the transform on a tab. I have a version of the code which I would like to share with you. Do you have an email I can use? Or can I encourage you to join our Discord server?

meerk40t/gui/scene/scene.py
```diff
@@ -142,6 +145,44 @@ def signal(self, *args, **kwargs):
         """
         self.scene._signal_widget(self.scene.widget_root, *args, **kwargs)
 
+    # Indicator for Keyboard-Modifier
```
Sophist-UK, line 148: Sorry - but IMO you need to use get_key_name in wxutils.py to check modifiers. This is the single place in Meerk40t that we sort out modifiers in a way that will be consistent across Windows, Linux and Mac.

@tatarize gets the final say, but personally I think this is an essential change.

meerk40t/gui/scene/scene.py
```diff
@@ -701,6 +746,39 @@ def event(self, window_pos, event_type=""):
             previous_top_element = self.hit_chain[0][0]
         except (IndexError, TypeError):
             previous_top_element = None
+
+        if event_type in (
```
Sophist-UK, line 750: See above comment.

meerk40t/gui/wxmmain.py
```diff
@@ -1691,6 +1693,9 @@ def on_export_signal(self, origin, frame):
             obj.image_height = image_height
             elements.add_elem(obj)
 
+    def on_update_statusmsg(self, origin, value):
+        self.main_statusbar.SetStatusText(value, 0)
```
Sophist-UK, line 1697: `self.update_statusbar(value)`

Sophist-UK (changes requested):
Sorry - but it is too late and I am too tired to review the detailed changes to the event processing. I can see that a hell of a lot of thought and work has gone into this, and I know from experience just how hard it is to get to grips with this stuff.

But here are my comments on the current version. The key comment is about how the modifiers are accessed - please see if you can use get_key_name because this is where we deal with all (or most) of the mac related key mapping issues (use of the control key being one of them - for example for macs you need to check that rawctrl is not pressed when you check that ctrl is checked - and by centralising this stuff we keep it contained to a single place where we make this Mac stuff work).

jpirnay:
@Sophist-UK 
> But here are my comments on the current version. The key comment is about how the modifiers are accessed - please see if  you can use get_key_name because this is where we deal with all (or most) of the mac related key mapping issues (use of the control key being one of them - for example for macs you need to check that rawctrl is not pressed when you check that ctrl is checked - and by centralising this stuff we keep it contained to a single place where we make this Mac stuff work).

I think I disagree: get_key_name in wxutils is translating only few of the relevant key events, i.e. all the Shift/Ctrl/Up-Events are missing. I see that you are amending the keybind-logic in your [PR#801](https://github.com/meerk40t/meerk40t/pull/801) but at least as of now they don't include such events either.
In addition we would need agreement of how these different Control-flavors (ControlDown() RawControlDown()) need to be used consistently.
So I would actually break the code IMHO if I would use get_key_name now - this will change of course if it will get the features need and I will happily comply to that new logic then.

tatarize (approved):
As a minor criticism, when you drag up->left. It blocks the thing in the corner. You can't see the symbol. Also ! is more often associated with NOT which isn't strictly XOR and isn't the operation being performed.  Since `!select` will absolutely select everything you select if it wasn't selected before.

Sophist-UK:
> think I disagree: get_key_name in wxutils is translating only few of the relevant key events, i.e. all the Shift/Ctrl/Up-Events are missing. I see that you are amending the keybind-logic in your [PR#801](https://github.com/meerk40t/meerk40t/pull/801) but at least as of now they don't include such events either.
> In addition we would need agreement of how these different Control-flavors (ControlDown() RawControlDown()) need to be used consistently.
> So I would actually break the code IMHO if I would use get_key_name now - this will change of course if it will get the features need and I will happily comply to that new logic then.

Using the Keymap window I have tested how `get_key_name` works in both current main and #801, and it appears that Ctrl-Shift-Up works on key-up as I expect it to in both branches. Except for the repeating keybind/aliases that start with +/-,  on keydown you are generally only interested in modifiers because for any keys something only happens on keyup.

The changes to `get_key_name` in #801 are A) adding additional keys to be recognised and named, B) adding functionality to send modifiers only on key-down used purely for the Keymap window. Aside from "control+" becoming "ctrl+" I don't see how this would materially change how your code would work.

jpirnay:
@Sophist-UK You are right if you talk about shif-/ctrl-/alt.. whatever PLUS another letter. But if you don't have that, so if you just press these keys on their own then you don't get any result. That's changed in your PR (provided you use it with return_modifier=True). The Keyup-Event only provides a value for Alt but not for Shift or Control.
I will create a mockup in my code that mirrors your PR to adhere to the Mac-Users needs and will replace that by a call to your Wxutils-Version as soon as it is merged.

Sophist-UK:
`return_modifier` was added to support displaying the modifier keys in the Keymap window when they were pressed down before the other key (like UP) was pressed and if you want the previous behaviour you don't use this flag. I have just merged the bind PR, so you can test your code with that now. And I have noted your comment on that PR about keyup not returning the modifiers when they are released and once I have finished what I am currently working on (coloured console text for use in help etc.) I will get to it if no one else has addressed it in the meantime.

PR #828, opened 2022-02-14
Fix outline statement

jpirnay:
Offset X and Y were always put to zero - fixes [#829](https://github.com/meerk40t/meerk40t/issues/829)


tatarize (approved):
Yeah, this was a thing we found once and never got around to fixing. It's merging very old code scheme with modern versions and doing it wrong. args() will not have any values there, and will set both to zero.

PR #831, opened 2022-02-14
Fix Codecy Checks with Pylint Disables

tatarize:
https://pylint.pycqa.org/en/latest/user_guide/message-control.html

Using fine grained pylint to disable issues with dynamically created attributes


Sophist-UK (dismissed):

tatarize:
Sleep changes codecy wanted might be a bit weirder, should allow better understanding of typing but might require the pylint turned off.

tatarize:
Fine grain disable didn't seem to work.

tatarize:
Nah, we can just disable the bits. It doesn't say it's corrected. But it didn't refind the error. So it's sort of lag or something.

tatarize:
I can't tell if codecy managed to disable these things with the pylint calls. Seems like it should have but I don't see them removed as issues.

tatarize:
@Sophist-UK This would need re-approval, then I could merge those changes upstream. It also helps keeping versions like services branch nicely together.

Sophist-UK (approved):
I am approving this only on the basis that it is Tat's as I have no idea what it does and cannot really review it properly. Also I have concerns that some of the black reformatting may result in other PRs failing to merge.

tatarize:
@Sophist-UK I updated the blindalias pr to merge.

PR #838, opened 2022-02-16
Permit swapping of branch elems and branch ops

tatarize:
* This might be a bit worthless. But, lets you drag operations and swap it with elements.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -182,7 +182,7 @@ class Node:
         return other is self
 
     def is_movable(self):
-        return self.type not in ("branch elems", "branch ops", "root")
+        return self.type != "root"
 
     def drop(self, drag_node):
         drop_node = self
@@ -286,6 +286,10 @@ class Node:
                     drop_node.add(e.object, type="opnode")
                     some_nodes = True
                 return some_nodes
+        elif drag_node.type == "branch elems" and drop_node.type == "branch ops":
+            drop_node.insert_sibling(drag_node)
+        elif drag_node.type == "branch ops" and drop_node.type == "branch elems":
+            drop_node.insert_sibling(drag_node)
         return False
 
     def reverse(self):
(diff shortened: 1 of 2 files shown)


tiger12506 (approved):
I tested this. Works as advertised, and I couldn't get it to cause an error.

Sophist-UK:
@tiger12506 said: "Is there anyway this could be made persistent?"

:+1

tatarize:
Seems like the bigger question is, "is this worth it"? I mean I could certainly have switching the order there set a setting and just load with the right settings. So swapping it would set the value and it would persist, is this too minor to care? Or should it just do the little bit of extra work and go ahead and merge? After making them expand (they seem to lose their children during the swap and need to be expanded).

It also tends to let you drag `branch op` and `branch elem` nodes which I guess need to be tested that they don't do anything weird if you drag them into other node types.

Sophist-UK:
If you hadn't started this I probably wouldn't have considered it important enough to think about. At this stage you have 3 options all of which are IMO equally good.

A. Drop the idea.
B. Merge this as-is and see if anyone complaints that they have to set it each time they open MK.
C. Add a persistent setting and merge.

I would be happy ... personally having used MK 7 for 8-9 months, I am used to the current tree and probably wouldn't swap anyway.

Sophist-UK:
Many years (decades!!) ago in late March I was on a Scuba Diving expedition, and we were in a restaurant having an end of week celebration when suddenly everyone's diving watches started beeping within seconds of one another. Turned out that we were all using the Casio dive watch and some chip designer had coded in an April Fools joke.

I was thinking that perhaps MK should have some Easter Eggs and that this sort of trivial but unusual user action is exactly the sort of place to hook in such things.

Another possible (tongue in cheek) example: a pop-up window for 10% of the users on Tat's birthday.

tatarize:
Too minor and I'm not interested enough in writing the option to maintain this between uses.

PR #839, opened 2022-02-17
Virtual Arrays: Step Repeat as Node-Type

tatarize:
Proof of concept.

* Add Special Operation: Virtual Array.
* Virtual Array currently limited to 2x2 50% 50% for bed size.
* Virtual Array does not simulate accurately.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -1572,7 +1572,9 @@ class VirtualArrayNode(Node):
         self.y_gap = y_gap
 
     def __repr__(self):
-        return "VirtualArray({rows}, {columns}, {x_gap}, {y_gap})".format(rows=self.rows, columns=self.columns, x_gap=self.x_gap, y_gap=self.y_gap)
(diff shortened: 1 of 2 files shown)


Sophist-UK:
Not sure I particularly like the name `VirtualArray` - `RepeatGrid` might be a better name.

If I read the code correctly, you create an Opnode which is a VirtualArray anywhere in the op tree and this holds details of the repeats and steps) and when you execute, the existence is noted and an additional step added at the end of the plan to do the repeat step - in which case it would do the step repeat on the optimised cut-code.

I am not sure whether using a special Opnode is the way to go - in my Design / Prepare / Burn philosophy (Design being tools to manipulate elements, Prepare being tools to set up the operations, Burn being Execute & Simulate combined, then I might want to be able to set the step-repeat values in Prepare (where they would be saved as special ops when I save the SVG file) or alternatively to dynamically add the step-repeat in fields in a panel on the Burn screen.

tatarize:
This isn't an opnode since those are references to elems, it's just an operation subnode specifically a `virtualarray` node type.

Issue #814, opened 2022-02-10
[Suggestion]: Console/CLI colors.


tatarize:
### Summary Description

Use RichText in gui Console to convert ANSI color coding. This would also permit ansi text colors in console. The console would convert the ANSI color coding to actual colors and bold. And they would be generally displayed in CLI mode.

### Current functionality

Currently all text is unformatted plain text.

### Future functionality

Color that attacks the eyes like angry fruit salad that hates you.

### MeerK40t Version

0.7.4010

### MeerK40t Type

Git

### Your Operating System

Windows

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Sophist-UK:
Functionality delivered in #849 with minor fixes in #855.

PR #849, opened 2022-02-20
Rich-text console + ansi o/s terminal window + bbcode

Sophist-UK:
This PR provides the following enhancements:
1. Colour capable rich-text console panel
2. Ansi enabled o/s terminal windows
3. BBcode to define colours, highlights etc.
4. Echo and cls commands to echo to the window (for test purposes) and clear current console panel contents
5. Coloured echo of commands and red command error messages.

The primary intended use of these enhancements is for improved help text, which will be delivered in a future PR.

The supported BBcode commands are as follows:
* Colours - ANSI colours - [black], [red], [green], [yellow], [blue], [magenta], [cyan], [white]
* Background colours - [bg-*]
* Highlights - [bold] / [/bold], [italic] / [/italic], [underline] or [underscore] / [/underline] or [/underscore]
* Negative - swap text and background colours - [negative], [positive]
* Raw mode - bbcode between [raw] and [/raw] will not be processed
* Reset - [normal]

Note: Raw mode is used for displaying the "echo" command as-is i.e. "echo [red]text" shows the command with the bbcode and no red text, and then outputs a line with the word "text" in red..

Note: This PR is restricted to providing support for coloured output, however it should be noted that the RichTextCtrl does enable display of images, and so it should be possible to have a future enhancement to provide commands to display images in the wx-based GUI Console panel (though not the ansi o/s terminal window).


tatarize:
Keep in mind this should replace these values with ANSI text and then have the GUI replace that with rich text of a color. So if this is printed to the CLI it gets marked red. And if it gets printed to the Console Window that would replace the ANSI text mark-up with red richtext mark-up.

Sophist-UK:
Or just translate it differently depending on the output device.

Sophist-UK:
I am thinking I will use BBCode for the hints.

tatarize:
The issue is you'd need to know the output device if you changed it based on the context. Whereas if you did it in ANSI color then printing to normal CLI would be fine. And the Console window could use the ANSI stuff to trigger the formatting, per #814.

win7

So red is '\033[31m' Then 32 for green, yellow, 33, blue 34, purple 35, 36 light cyan.  With '\33[0m' to end such things. 

https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_(Select_Graphic_Rendition)_parameters

These would, I suppose, get double replaced in the Console window whereas they'd only get single replaced outside of the Console window. But, you convert `[red]` to `\033[31m` and `[/red]` to `\33[0m` or whatever somewhat BBCode looking stuff you'd end up with. Then you could quickly implement something vaguely ANSI standardish in the Console Window itself. With some rich text swaps conversions in there that from my cursory check seemed like it would be pretty reasonable. Then regardless where this text prints out it would be correctly formatted.

My reservations is that I'm not sure how much of this would be really useful or how much would be used, or how much would start looking like angry fruit salad.

Sophist-UK:
If something doesn't work on the console panels I would prefer to see the original controls than have also characters displayed or possibly not displayed.

Sophist-UK:
My intention is **not** to add colours to individual text messages, but instead to add them to the help methods that output this text to console etc. so that headings come out in one colour and details in another etc. and so that command echoes are a different colour and error messages another etc. So hopefully we will not get a confusing rainbow of colours.

Sophist-UK:
I believe that the core functionality of this POR is now working if anyone wants to test it.

I just need to apply colours to command echo, errors and help (though I may leave this until I rework the help command), and this PR can be merged,

Sophist-UK:
Unit tests failed and there are two minor coach comments I need to fix.

Sophist-UK:
Should be good to go now.

jpirnay:
Looks good to me :-)
I can confirm translated help text items recognize the new bb-codes as well:
image

Sophist-UK:
> Looks good to me :-)

Can you approve it then please?

> I can confirm translated help text items recognize the new bb-codes as well:

It would do - the code is at a very low level in the display routines and applies to all console output via channel.

jpirnay (approved):
Looks fine to me

Sophist-UK:
@tiger12506 Apologies if I was a bit quick off the mark and didn't leave yo enough time to comment.

tatarize:
What's with the very extensive panel->control refactor in a broad swath of different files. 33 files modified but most of those are renaming a panel to call it "control" which seems a lot of file changed for a lateral move at best.

Sophist-UK:
@tatarize @tiger12506 I will create another PR to revert that specific commit and fix it up to work with just Console having common naming.

Sophist-UK:
Reverted in #855

tatarize:
Lateral moves don't really need reverting. It's just a lot of little changes or whatever.

tatarize:
It only somewhat matters since I'll have to merge this stuff into tatarize-services too and manipulating other files means I need to pick through those for which changes are reasonable.

Sophist-UK:
Goo reason to merge the revert then.

PR #851, opened 2022-02-22
Fix #850

### LINKED ISSUE #850: [Bug?] wxmscene registers a console. Is this correct?
https://github.com/meerk40t/meerk40t/blob/3dd762b0a33d5889487ce00748b4c71b698d103d/meerk40t/gui/wxmscene.py#L36 has the same registration as  https://github.com/meerk40t/meerk40t/blob/3dd762b0a33d5889487ce00748b4c71b698d103d/meerk40t/gui/panes/consolepanel.py#L27 . Is this correct or a bug?

Sophist-UK:
This does not add a pane to the Panes menu if it has no name or caption (because errors occur if there is no string to use as text when you add a menu item).

However, if a name or caption is ever added to e.g. wxscene then it will start to appear in the Pane menu so in the future we might need an alternative way to distinguish panes that should be shown from those that shouldn't.

### DIFF
--- a/meerk40t/gui/wxmmain.py
+++ b/meerk40t/gui/wxmmain.py
@@ -1037,15 +1037,14 @@ class MeerK40t(MWindow):
 
         def toggle_pane(pane_toggle):
             def toggle(event=None):
-                pane_obj = self._mgr.GetPane(pane_toggle)
+                pane_obj = self._mgr.GetPane(pane_toggle.name)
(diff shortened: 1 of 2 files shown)


Sophist-UK:
Good to go.

tatarize (approved):
Heavier on the changes than I'd like but they all look right.

PR #852, opened 2022-02-22
Selection-Rectangle: optics + make Ctrl-Key more MacOS-aware

jpirnay:
a) Optics: Draw two indicators in the selection rectangle to make it clearer to see even if mouse select left up (was covered by the mouse pointer before), change symbol from ! to ^
b) Keypress-behaviour, anticipates Sophist-UKs PR for keybinds, Ctrl+ should be more consistent with other keys on MacOS

### DIFF
--- a/meerk40t/gui/scene/scene.py
+++ b/meerk40t/gui/scene/scene.py
@@ -14,8 +14,7 @@ from meerk40t.gui.zmatrix import ZMatrix
 from meerk40t.kernel import Job, Module
 from meerk40t.svgelements import Matrix, Point, Viewbox
 
-# Comment this line out as soon as PR805 gets merged
-# from meerk40t.gui.wxutils import get_key_name
+from meerk40t.gui.wxutils import get_key_name
 
 MILS_IN_MM = 39.3701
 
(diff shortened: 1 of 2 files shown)


meerk40t/gui/scene/scene.py
```diff
@@ -14,6 +14,9 @@
 from meerk40t.kernel import Job, Module
 from meerk40t.svgelements import Matrix, Point, Viewbox
 
+# Comment this line out as soon as PR805 gets merged
+# from meerk40t.gui.wxutils import get_key_name
+
```
tatarize, line 19: PR #801 was merged. #805 is a github issue template thing

tatarize (changes requested):

meerk40t/gui/scene/scene.py
tatarize: The use of get_key_name there is thus a bit redundant as code or whatever. Might take merge from main

jpirnay: As indicated in the comments, these were just proxy routines to be replaced those in the recent PR from Sophist
@Sophist-UK : I will have a look at the code in wxmain to think about the key-up event. There are some challenges around i.e. Ctrl-Shift pressed simultaneously and then released... I can easily work around this in my code, but I think it deserves a 'real solution'

Sophist-UK:
I indicated how I would approach this in the comments on the other PR.

tatarize (approved):

PR #853, opened 2022-02-23
Improve keymap key display still further

Sophist-UK:


jpirnay:
Press Ctrl+Shift at the same time: get_key_name provides down and **ctrl+shift+**
a) Release both: **shift+ctrl+**
b) Release only one of the two: (i.e. shift-Key): get_key_name provides up **shift-+ctrl**+ followed by a up **ctrl+**

Try to press Ctrl+Shift+p. get_key-name provides down with **ctrl+shift+p**
Release the shift key (while having ctrl+p pressed): down with **ctrl+shift**
(if youl release the p key, the even if ctrl + shift are still pressed, down with **ctrl+shift+p** is provided)

Sophist-UK:
If you read the wx documentation Shift is a special case on keyup.

Sophist-UK:
This PR does not materially change how `get_key_name` works - the small changes are only as recommended by wx documentation i.e. to use `GetUnicodeKey` first and `GetKeyCode` only when `GetUnicodeKey` shows it is some sort of special key. So this PR is not going to change the behaviour of `get_key_name`.

What it does change is how the Keymap UI behaves. Non-modifier keys normally register on keydown - so that is how Keymap now works. Any modifier changes after that are ignored until you either press another non-modifier or release all keys and press a modifier again. This feels to me to be consistent, easily understandable by the user and using clean easily-understood code.

jpirnay:
@Sophist-UK Again, I have no fundamental objections - what I am trying to say though is that code that relies on a key-up event needs may need to apply some additional logic, the outcome of get_key_name **alone** may not suffice.

Sophist-UK:
If there is a reason to track shift-up and we decide to reverse the wx decision to work this way (which presumably has some logic to it) then there is no reason we can't do that - though the logic will require persistence and persistence implies storing stuff in either an object or a global variable - so it would probably need to be done in a calling object which has persistence rather than in a utility function which doesn't.

jpirnay (approved):

jpirnay:
As there is only one piece of code that looks for the release of modifier keys and has circumvented this very specific use-case I think this is good to go

PR #854, opened 2022-02-24
Validate Length (draft)

jpirnay:
Meerk40t is quite lenient when it comes to checking the validity of Length parameters. Actually it converts invalid statements to a 0.0. I.e. outline blablabla is translated to outline 0.0. 
I think we should introduce some basic validity checking to address this.

The code attempts to do some basic stuff at the level of the Length class in svgelements and provides a new property isValid. I just used this as a POC in elements.py in the 'outline' command.
image

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -3658,7 +3658,7 @@ class Elemental(Modifier):
         @context.console_command(
             "outline",
             help=_("outline the current selected elements"),
-            input_type=(
+            input_type=(            
                 None,
                 "elements",
             ),
@@ -3679,6 +3679,13 @@ class Elemental(Modifier):
             """
             if x_offset is None:
                 raise SyntaxError
+            elif not  x_offset.isValid:
+                raise SyntaxError(_("This is not a valid length"))
+            if not y_offset is None:
+                if not  y_offset.isValid:
+                    raise SyntaxError(_("This is not a valid length"))
+            
+
             bounds = self.selected_area()
             if bounds is None:
                 channel(_("Nothing Selected"))
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -3210,8 +3210,13 @@ class Elemental(Modifier):
                 raise SyntaxError
             if x is None:
                 x = Length("100%")
+            else:
+                if not x.isValid:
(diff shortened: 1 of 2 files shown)


Sophist-UK:
1. I applaud you adding error checking code for the validity of constant arguments provided to console commands. This is a good thing.
2. You can make changes to svgelements.py here in Meerk40t, but they will be overwritten whenever this is updated (by a manual copy of the file) from the svgelements repo. So once you have finalised the changes here you need to submit a PR to svgelement repo for the same changes in the hope that they will become permanent there.
3. Of course, `Length` is only one of several similar classes (like Point which takes coordinates, and Angle) which are measurement classes rather than elements. So you may want to think about including those in  the same PR.

meerk40t/core/elements.py
```diff
@@ -3658,7 +3658,7 @@ def element_fill(command, channel, _, color, data=None, filter=None, **kwargs):
         @context.console_command(
             "outline",
             help=_("outline the current selected elements"),
-            input_type=(
+            input_type=(
```
Sophist-UK, line 3661: A very minor point, but you have added trailing spaces here - you might want to set your editor to remove trailing spaces on save.

jpirnay:
I added the validity checks for all Lengths in elements.py.
On Angles: as Angle.parse is used I think this already covered although the error message shown will mean nothing to the ordinary user. Will have a look...

meerk40t/svgelements.py
```diff
@@ -558,6 +558,7 @@ class Length(object):
     """
 
     def __init__(self, *args, **kwargs):
+        self.isValid = False
```
tatarize, line 561: isValid is camelCase and svgelements is snake_case and for various reasons it's really hard to change svgelements code since it's a stand alone project itself.

tatarize (changes requested):

tatarize:
The process of making svgelements changes is a bit more involved, they should be made at https://github.com/meerk40t/svgelements and need to have unittests, correct casing, and probably have the .valid check as a `@property` rather than something that runs for every single length parsed. Which isn't far fetched, but can't really cope with any breaking change but just has a lot tighter restrictions since it's a used-elsewhere standalone library written for meerk40t.

jpirnay:
Will add a new Draft PR

PR #857, opened 2022-02-26
0.8 add settings classes

Sophist-UK:
The purpose of this PR is to introduce various Setting classes so that Services settings can be programatically distinguished from Services class attributes even if they can be used identically.

Before writing this code, this PR splits the existing 0.8 Kernel into several smaller class files and fixes several bugs.


Sophist-UK:
For the record this was attempting to merge into the wrong branch - hence the lengthy list of commits.

PR #858, opened 2022-02-26
0.8 add settings classes

Sophist-UK:
The purpose of this PR is to introduce various Setting classes so that Services settings can be programatically distinguished from Services class attributes even if they can be used identically.

Before writing this code, this PR splits the existing 0.8 Kernel into several smaller class files and fixes several bugs.


Sophist-UK:
@tatarize As per our Discord discussion, can I please have a decision on whether you are going to accept the idea of having the kernel split in this way or not?

If you agree to go this way, then (of course) I will fix up the changes needed for commits you have made so far (which is another reason why I need a prompt decision) and I will address any previous points made on Discord and any review comments.

Sophist-UK:
@tatarize  AFAIK this now meets all the criteria you have set to accept it. The kernel should be completely stand-alone and imports should be backwards compatible. I have even changed the imports in the kernel to be relative (an obvious requirement to be able to lift the package and use it elsewhere unchanged).

Sophist-UK:
I suggest that we merge this as-is and I start a new PR for the Settings changes.

P.S. This approach would appear to work for svgelements, enabling us to split svgelements whilst providing backwards compatibility for existing applications.

tatarize (approved):

PR #863, opened 2022-02-28
Add checks for valid lengths

jpirnay:
**Attempt #2**
Meerk40t is quite lenient when it comes to checking the validity of Length parameters. Actually it converts invalid statements to a 0.0. I.e. _outline blablabla_ is translated to _outline 0.0_.
I think we should introduce some basic validity checking to address this.

The code attempts to do some basic stuff by extending the Length class in elements.py (without touching svgelements and hopefully avoiding some of the pitfalls of manipulating this publicly available class) and provides a new property is_valid_length.


tatarize (approved):
A quite solid method of checking that.

tiger12506 (approved):
Tat already approved before the typos. I believe the typo fixes are correct, so I'm approving to let this out of limbo.

PR #864, opened 2022-03-01
Reverts Clear Ops on Open Project introduced by #763

Sophist-UK:
This PR reverts the Clear Operations on opening a project introduced in #763. Open Project is still different from File Import in that it clears everything else (as you would expect a project to do).

At present, I have not changed the Recent menu functionality to be consistent with this, Recent files are treated as File Imports and are loaded alongside existing files rather than as a replacement for them. I am unclear whether the Recent menu should continue to operate like a File Import, or instead operate like an Open Project or whether we might actually need two recent menus, one which acts like Open Project and one which acts like File Import. On the whole, I favour:

a. Renaming "Recent" to "Recent Projects";
b. Treating opening a recent file as if it was an Open Project. 
c. Only populating the recent file list from Open Project and Save and not from File Import.

I think this PR could include the above changes once there is consensus on approach. 

Can reviewers tagged (any anyone else interested) please give their views if you wish to contribute?

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -6088,15 +6088,18 @@ class Elemental(Modifier):
         pass
 
     def clear_elements_and_operations(self):
-        self.clear_elements()
         self.clear_operations()
+        self.clear_elements()
 
-    def clear_all(self):
+    def clear_project(self):
         self.clear_elements()
-        self.clear_operations()
         self.clear_files()
         self.clear_note()
         self.validate_selected_area()
+
+    def clear_all(self):
+        self.clear_operations()
+        self.clear_project()
 
     def clear_note(self):
         self.note = None
--- a/meerk40t/gui/wxmmain.py
+++ b/meerk40t/gui/wxmmain.py
@@ -1810,15 +1810,17 @@ class MeerK40t(MWindow):
                 break
         self.populate_recent_menu()
 
-    def clear_project(self):
+    def clear_project(self, clear_ops=True):
         context = self.context
         self.working_file = None
         self.validate_save()
-        context.elements.clear_all()
+        context.elements.clear_project()
+        if clear_ops:
+            context.elements.clear_operations()
         self.context(".laserpath_clear\n")
 
     def clear_and_open(self, pathname):
-        self.clear_project()
+        self.clear_project(clear_ops=False)
         if self.load(pathname):
             try:
                 if self.context.uniform_svg and pathname.lower().endswith("svg"):


tiger12506:
I agree the revert should go through. Operations should not clear on open project. One thing that makes this feature super jarring is it is a full clear, rather than a restore of "previous" operation list.

I believe recent files should keep list of both files opened as project and as import file (as they do now).

The only idea that I can muster is if recent files could remember which option was used for the recent file -- if it was opened as project or as an import, and the recent file menu again opened it the same way -- that would probably feel very intuitive, IMHO.  My line of thinking here is you have a project open, you want to add a small hole cutout, you import file -- it shows up in recent list, you go to recent list and you don't really want it to replace whole project, you want to import it again, as you did before!

Sophist-UK:
@tiger12506 I am unclear whether opening a recent file should work like opening a project i.e. clear other files, or like importing another file i.e. keep existing files.

Of course we could present the recent file list twice, once as "Recent Project" and again as "Import Recent" or we can pick one or the other (in which case IMO it should act like Open Project rather than Import File). As an alternative we could maintain a list of Opens/Saves and a separate list of Imports and have separate submenus, but then you wouldn't be able to load a previously imported file as a project or a previously saved file as an Import. I have no issues in including imports in the Recent file list, I just though as a minor point that it might be more logical to only have Projects in the Recent list.

tiger12506:
So what do you think of my suggestion to have one recent file list that maintains whether the file was opened as a project or a file import?

Sophist-UK:
@tiger12506 I don't think much of that idea at all - but not in the way that sounds. Just because AFAIK that is how it works right now. (Just tested it and it already seems to do this.)

Sophist-UK:
At least, an Imported File makes it into the list - are you suggesting that we flag each file in the list as either a project or an import and repeat the operation if you select that recent file?

tiger12506:
I don't like the clutter of the separate lists. And that also could cause user confusion. (What does it mean to be a project vs. imported file? And then you have to explain it).

> I are you suggesting that we flag each file in the list as either a project or an import and repeat the operation if you select that recent file?
Yes. I think it will be more intuitive and feel more natural than perhaps you are expecting..

tiger12506:
btw, just for documentation purposes, inpain is against the separation entirely. He says there is already a New feature that clears items, Open should always just do an "import file".

Furthermore, he brought up the excellent question of what happens when you drag in files. The answer is: it does an import, and adds to recent list.
We believe this is correct behavior. Just mentioning because it's relevant.

Sophist-UK:
A bit more input regarding drag and drop:

A. we could probably use Control and / or Shift keyboard modifiers to give a choice about whether Drag and Drop is treated as a Project Open or File Import; but this just makes the UI more complicated to understand for novices.

B. the fundamental choice seems to be between A) simplicity where the two actions from the file menu are Clear the bed and Add a file (if you want to remove one of multiple files you would have to right click on the file node and select remove or click on it and press delete) - and Drag and Drop of a file adds it in exactly the same way OR B) slightly enhanced convenience (whereby you can clear the bed and add a file by treating it as a project, this saving one action) at the expense of a much more complex set of choices (Open Project, Recent Projects, Import File, Recent File - possibly with only one recent menu but with a more complex UI).

My gut reaction is to go for the simplicity. A file Menu that has Add File (alt-fa, Ctrl-O), Recent Files (alt-fr), Clear Bed (alt-fc, Ctrl-C), Clear All (alt-fn, Ctrl-N) so I put this up as a straw man for discussion.

tiger12506:
> My gut reaction is to go for the simplicity. A file Menu that has Add File (alt-fa, Ctrl-O), Recent Files (alt-fr), Clear Bed (alt-fc, Ctrl-C), Clear All (alt-fn, Ctrl-N) so I put this up as a straw man for discussion.

Mine too. IMHO, this is the way to go (which is basically what it was before the original pull request with just better naming)

Sophist-UK:
Obviously need to show a bit more time for other people to comment but if this is consensus...

Will probably be a few days before I can get to coding this but I will add it to this PR.

joerlane:
At a quick glace; I can't find another application that uses the File menu to contain any form of a Clear operation. I presume that belongs in an Edit menu.

tiger12506:
> At a quick glace; I can't find another application that uses the File menu to contain any form of a Clear operation. I presume that belongs in an Edit menu.
> 
> Though; Close is a pretty standard operation in a File menu. Perhaps it's an issue of accepted nomenclature for at least one of the two.

"Close All Files" would probably be understood as appropriate nomenclature in File menu?

jaredly:
🤷‍♂️ if y'all are agreed on this, I won't fight it. I think the current model of having operations be persistent across projects is strange, but obviously it's working for current users.

The model that I would like to have is: (which mirrors inkscape)
- "open file" clears everything, adds to the recents list
- "import file" doesn't clear anything, but adds the image to the recents list
- "open recent file" clears everything, regardless of whether it was imported or opened.

we already have the ability to save & restore operations lists, although it could potentially get some tweaking to make it more discoverable/usable.

but given that the people spending the most time with this project don't want stuff to be cleared, 🤷‍♂️ it's fine.

as mentioned elsewhere, aligning behavior with e.g. inkscape might make more sense once (if?) meerk40t supports having multiple "documents" open at the same time. (because in inkscape, selecting something from the recents list doesn't actually clear everything, it just opens a new window).

Sophist-UK:
> if y'all are agreed on this, I won't fight it.

I can't speak for anyone else but personally I am not wedded to this. What I want to see is something that is simple and consistent and easy for users to understand. Right now "this" is the simplest and easiest to understand **but** if we later start to support loading designs simultaneously to support burning on multiple lasers then I don't think it will work as well as the "project" metaphor. The trouble is that we don't get have the consensus on the long term vision, and that may mean we instead choose the solution which fits the current model instead.

I genuinely haven't made up my mind yet.

Of course, there is nothing to say that we can't go with the simpler sorry term solution now and change again later.

jaredly:
certainly. I'm totally open to a simple solution now & more investment later.
Would "supporting multiple lasers" be the main motivation for allowing multiple projects to be open at the same time? I for one would love to be able to have one burn going, and use the wait time to work on tweaking another project, fwiw.

tatarize:
So long as the data is sent to the spooler, you can go ahead and fiddle with anything you want. Move things around and open a different project etc. So its not really that needed to have multiple projects like that. There's only one design space. At that rate we could probably set it to show the simulation for the currently spooled job if that was needed. Then drop cutcode as it gets processed. A second window wouldn't let you manipulate anything other than a second scene or whatever and set that up.

I'm not saying it's a problem, it's certainly not, but that's not something this feature would guarantee. I think such workflow changes might be better after 0.8.x when the tools exist for people and ask about the various workflows people have. The data sending stuff is not happening in the scene. And you can, after it's sent to the backend mess with it all you want. The only risk is doing something laggy and the sending taking longer or something.

Sophist-UK:
> So long as the data is sent to the spooler, you can go ahead and fiddle with anything you want. Move things around and open a different project etc. So its not really that needed to have multiple projects like that.

Sorry @tatarize , but I don't think that having a single project will cut it with multiple lasers. For example if you are doing several burns of project A on machine 1 and several burns of project B on machine 2 you would not want to be swapping three project loaded back and forth. I was hoping that the services architecture would make it easy to hold several projects in memory at once.

> I think such workflow changes might be better after 0.8.x when the tools exist for people and ask about the various workflows people have.

I was absolutely assuming that!

> The only risk is doing something laggy and the sending taking longer or something.

Unless the python folks make multi-threading scheduling more predictable (e.g. by implementing the GIL design I proposed to them) threading will never balance workload correctly, and even if they do implement it I can't guarantee that it will be perfect.

I think we should give serious consideration to breaking out the actual burn code into a separate process so that CPU intensive workloads like simulation cannot impact it.

jaredly:
> So long as the data is sent to the spooler, you can go ahead and fiddle with anything you want.

And even with a single laser going, I frequently execute in multiple passes that I execute in succession so I can monitor & tweak as needed, so it wouldn't work for me to blow away a project that I'm in the middle of cutting to start working on something else.
It's great to hear that multiple workspaces is likely to be more doable in the near future!

Sophist-UK:
Having had some more time to contemplate this, I think we should keep a weather eye on potential future plans and not move away from a Project approach i.e. stick with both Open Project and Import File, the only difference being that Open Project clears the existing design before doing the import. I think that the recent files menu should contain only the Projects opened and saved and not files imported, on the basis that if you are importing several files and you think you will want the same combination again, then you are likely to save it as a combined file and it will then appear in the recent files list.

But I am not wedded to this either - so I plan to leave it for a few days so that everyone has the chance to comment further.

PR #865, opened 2022-03-01
Add a sponsor menu

Sophist-UK:
I appreciate that @tatarize may find this controversial and may not want to have it added. Please feel free to close this if this is the case.


Sophist-UK:
@tatarize I assume that you don't like asking for donations. Should we delete the branch in this case?

PR #866, opened 2022-03-01
Add statusbar help for menus and buttons

Sophist-UK:
1. Add status bar help for main menu items
2. Align ribbon button tooltips to same help messages
3. Eliminate all code relating to View/Hide Tree. Tree is now in a Pane which can be shown or hidden. If you hide the Tree from the View menu it isn't immediately hidden, but when you restart it is, but then you get crashes when you try to load a file.


tiger12506:
There are some Panes menu items that have "(no help text)" appear in the statusbar. Should that be addressed?

tiger12506:
Also, on my machine, the statusbar that shows these messages is kinda short, so many of them get cut off. I would like to see the message area in the status bar get larger, but I don't know what the other sections of the statusbar are doing...

Sophist-UK:
One shows USB status, One controller status. Not sure off hand what the last one does. But A) they don't need to be 25% each, and B) we could probably "multiplex" the other stuff either swapping it every few seconds or using a marquee scroller.

I didn't consider this too deeply when I added the status bar help section because IMO we need to rework the whole status information anyway - users keep asking for better progress information during a burn and there is a reasonable amount of work needed to provide both better status bar info and better Spool window information. So this needs a significant PR IMO.

tiger12506:
That's fair.

tiger12506 (approved):
Okay with me.

Sophist-UK:
At this point, I am just trying to provide some basic GUI help to make the UI easier for new users.

We also need to rework most of the language strings at some point - and if we want decent translations, that needs to be sooner rather than later. I still need to do help strings for the right-click menus (ideally ideally for 0.7.5 - just not sure how easy they would be to merge into 0.8 at this point - which is why I am holding off on these until I have reviewed whether they would merge easily). Then we need to rework the format strings from `"%s" % string` to `"{s}".format(s=string)`. And then we shouldn have stable strings for translations.

Sophist-UK:
Now that the measurements are working in 0.8, I think we are really quite close to having release candidates. Which is why I have been pumping out these small 0.7.5 PRs today with an idea that we could have a 0.7.,6 Beta 1 release soon as a holding position for users whilst we get 0.8 ready, and then stabilise the 0.7.x line to bug fixes only and make all new enhancements go to 0.8.

Sophist-UK:
@tatarize Is this OK with you?

Sophist-UK:
P.S. Not sure what happened to the comment about Panes items not having status bar help - but I only did the wxmeerk40t and wxmribbon files, so there is no doubt a lot of other help which could be added.

tiger12506:
Looks like the (no help text) for Panes menu items is present before these changes, my comment was unwarranted.

Sophist-UK:
The panes menu is programmatically added from the caption names of panes. I will need to add an attribute to the object and go through each pane file to add it. Best left to v0.8 because of the no of files that are changed.

tatarize (dismissed):

Sophist-UK:
@tatarize Sorry - but I needed to make a tiny change to a variable name to avoid a Codacy medium error.

tatarize (approved):

PR #867, opened 2022-03-01
Make twitchless mode default for 0.7.6

Sophist-UK:


tatarize (approved):
I don't love the name twitchfull (I don't have a better one).

PR #868, opened 2022-03-01
Extend grid and add circ_copy, radial and shape

jpirnay:
In preparation for some graphical tools I've extended 
## The existing grid -command 
by a new parameter origin, that describes where the orginal element(s) shall be located in respect to the grid . The parameter needs to be provided in the format:
````
grid ... --origin ox,oy    or       grid ... -o ox,oy
````
so e.g. 
````
grid  3 3 -o "2,2"
````
will create a 3x3 grid with the original elements right in the middle.
image

## A new circ_copy - command
that takes some parameters to create (potentially rotated) copies on a circular arc around the orginal element(s)
image
```
circ_copy <copies> <radius> [ <startangle> <endangle> <rotate> <deltaangle> ]
```
The optional parameters can be provided like:
--rotate --startangle 0deg etc.
image
Notabene: normally you will equally distribute the amount of copies between startangle and endangle. If you omit the endangle and provide the optional _deltaangle_, then the amount of copies will be created starting from the _startangle_ every one _deltaangle_ apart:
grafik

## A new radial- command
that takes some parameters to create (potentially rotated) copies on a circular arc with the orginal element(s) being one of these
```
radial <repeats> <radius> [ <startangle> <endangle> <rotate> <deltaangle> ]
```
The optional parameters can be provided like:
--rotate --startangle 0deg etc.
image

Notabene: While circ_copy is creating copies around the original elements, radial is creating all the copies around a center just -1*radius to the left. So the original elements will be part of the circle.

## A new shape - command
that takes some parameters to create a [regular polygon](https://en.wikipedia.org/wiki/Regular_polygon) or regular star polygons (see below)
image

```
shape <corners> <x> <y> <radius> <startangle> <inscribed>
```
- \<corners\> defines the amount of vertices of the polygon. 
These lie all on a circle around _x_, _y_  with radius _radius_ (see also _inscribe_ below)
- \<startangle\> (optional) defines where to start with the polygon: <br>
image

- \<inscribed\> (optional) as indicated above the vertices lie on a circumscribing circle of the polygon. If you want to have the polygon outside of the inscribing circle you can use the _inscribed_ parameter :
````
shape 3 2in 2in 1in --inscribed
````
image

If you do provide the additional parameter "_radius_inner_" then every other point will lie on the inner circle providing a more star-like figure:
```
shape 8 5cm 2cm 1cm --radius_inner 50%
````
grafik
You may specify actually the alternating sequence with the _alternate_seq_ option in conjunction with the aforementioned  _radius_inner_ option. This will create somewhat gear-like shapes:
Gears

If you provide the optional parameter "--density" then you will create a [regular star polygon](https://en.wikipedia.org/wiki/Star_polygon) 
```
shape <corners> <x> <y> <radius> [ <startangle> ] --density num
```
image

- \<corners\> defines the amount of vertices of the polygon. 
- \<density\> define the line to draw from one vertice to the one \<density\> points away.
- These lie all on a circle around \<x\>, \<y\> with radius \<radius\>
- \<startangle\> (optional) defines where to start with the polygon

Please note that these **may** give some nice stars if the values for corners and density are an odd/even combination (if you want to know more read the wikipedia article linked above about those two numbers supposed to be coprimes). Meerk40t will provide some feedback if the chosen parameters are suboptimal:
density

You can combine and experiment with all parameters to get some nice looking polygons, i.e.
image

NB: you may invoke shape with 1 corner and 2 corners as well: these will result in 1 single point, a line with length _radius_ respectively

## polygon and polyline
allow now tuples of valid Lengths as parameter
so 
```
polyline 0 0 1000 1000  1000 0 
```
is equivalent to 
```
polyline 0cm 0cm 1in 1in 1in 0mm
```

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -3318,7 +3318,7 @@ class Elemental(Modifier):
         )
         @context.console_command(
             "radial",
-            help=_("radial <repeats> <radius> <cx> <cy> <startangle> <endangle> <rotate>"),
+            help=_("radial <repeats> <radius> <startangle> <endangle> <rotate>"),
             input_type=(None, "elements"),
             output_type="elements",
         )
@@ -3521,7 +3521,7 @@ class Elemental(Modifier):
         @context.console_argument(
             "cy", type=Length, help=_("Y-Value of polygon's center")
         )
-        @context.console_argument("radius", type=Length, help=_("Radius"))
+        @context.console_argument("radius", type=Length, help=_("Radius (length of side if --side_length is used)"))
         @context.console_option(
             "startangle", "s", type=Angle.parse, help=_("Start-Angle")
         )
@@ -3533,10 +3533,18 @@ class Elemental(Modifier):
             help=_("Shall the polygon touch the inscribing circle?"),
         )
         @context.console_option(
+            "side_length",
+            "l",
+            type=bool,
+            action="store_true",
+            help=_("Do you want to treat the length value for radius as the length of one edge instead?"),
+        )
+
+        @context.console_option(
             "radius_inner",
             "r",
             type=Length,
-            help=_("Alternating radius for every other vertice"),
+            help=_("Alternating radius for every other vertex"),
         )
         @context.console_option(
             "alternate_seq",
@@ -3552,7 +3560,7 @@ class Elemental(Modifier):
         @context.console_command(
             "shape",
             help=_(
-                "shape <corners> <x> <y> <r> <startangle> <inscribed> or polyshape <corners> <r>"
+                "shape <corners> <x> <y> <r> <startangle> <inscribed> or shape <corners> <r>"
             ),
             input_type=("elements", None),
             output_type="elements",
@@ -3567,6 +3575,7 @@ class Elemental(Modifier):
             radius,
             startangle=None,
             inscribed=None,
+            side_length=None,
             radius_inner=None,
             alternate_seq=None,
             density=None,
@@ -3595,6 +3604,7 @@ class Elemental(Modifier):
                 radius = radius.value(
                     ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
                 )
+                # No need to look at side_length parameter as we are considering the radius value as an edge anyway...
                 if startangle is None:
                     startangle = Angle.parse("0deg")
 
@@ -3627,29 +3637,6 @@ class Elemental(Modifier):
                     if not radius.is_valid_length:
                         raise SyntaxError("radius: " + _("This is not a valid length"))
 
-                if startangle is None:
-                    startangle = Angle.parse("0deg")
-                if inscribed is None:  # Use circumscribed circle by default
-                    inscribed = False
-
-                if alternate_seq is None:
-                    if radius_inner is None:
-                        alternate_seq = 0
-                    else:
-                        alternate_seq = 1
-
-                if radius_inner is None:
-                    radius_inner = radius
-                else:
-                    if not radius_inner.is_valid_length:
-                        raise SyntaxError(
-                            "radius_inner: " + _("This is not a valid length")
-                        )
-                if density is None:
-                    density = 1
-                if density<1 or density > corners:
-                    density = 1
-
                 cx = cx.value(
                     ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
                 )
@@ -3659,18 +3646,51 @@ class Elemental(Modifier):
                 radius = radius.value(
                     ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
                 )
+
                 if (
                     isinstance(radius, Length)
                     or isinstance(cx, Length)
                     or isinstance(cy, Length)
                 ):
                     raise SyntaxError
-                if inscribed:
-                    radius = radius / cos(pi / corners)
 
-                radius_inner = radius_inner.value(ppi=1000, relative_length=radius)
-                if isinstance(radius_inner, Length):
+                if startangle is None:
+                    startangle = Angle.parse("0deg")
+
+                if alternate_seq is None:
+                    if radius_inner is None:
+                        alternate_seq = 0
+                    else:
+                        alternate_seq = 1
+
+                if density is None:
+                    density = 1
+                if density<1 or density > corners:
+                    density = 1
+
+                # Do we have to consider the radius value as the length of one corner?
+                if not side_length is None:
+                    # Let's recalculate the radius then...
+                    # d_oc = s * csc( pi / n)
+                    radius = 0.5 * radius / sin(pi / corners)
+
+                if radius_inner is None:
                     radius_inner = radius
+                else:
+                    radius_inner = radius_inner.value(ppi=1000, relative_length=radius)
+                    if not radius_inner.is_valid_length:
+                        raise SyntaxError(
+                            "radius_inner: " + _("This is not a valid length")
+                    )
+                    if isinstance(radius_inner, Length):
+                        radius_inner = radius
+
+                if inscribed:
+                    if side_length is None:
+                        radius = radius / cos(pi / corners)
+                    else:
+                        channel(_("You have as well provided the --side_length parameter, this takes precedence, so --inscribed is ignored"))
+
                 if alternate_seq < 1:
                     radius_inner = radius
 
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -3272,8 +3272,14 @@ class Elemental(Modifier):
             y = y.value(ppi=1000, relative_length=height)
             if isinstance(x, Length) or isinstance(y, Length):
                 raise SyntaxError
+            if origin is None:
+                origin = (1,1)
             cx, cy = origin
             data_out = list(data)
+            if cx is None:
+                cx = 1
+            if cy is None:
+                cy = 1
             # Tell whether original is at the left / middle / or right
             start_x = -1 * x * (cx - 1)
             start_y = -1 * y * (cy - 1)
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -3224,7 +3224,7 @@ class Elemental(Modifier):
             "o",
             type=int,
             nargs=2,
-            help=_("Position of copy in matrix (e.g '2,2' or '4,3')"),
+            help=_("Position of original in matrix (e.g '2,2' or '4,3')"),
         )
         @context.console_command(
             "grid",
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -3575,139 +3575,164 @@ class Elemental(Modifier):
         ):
             if corners is None:
                 raise SyntaxError
-            if corners < 3:
-                raise SyntaxError(_("A polyshape needs to have at least three corners"))
-
-            if cx is None:
-                raise SyntaxError(
-                    _(
-                        "Please provide at least one additional value (which will act as radius then)"
-                    )
-                )
-            else:
-                if not cx.is_valid_length:
+            if corners <= 2:
+                if cx is None:
+                    cx = Length(0)
+                elif not cx.is_valid_length:
                     raise SyntaxError("cx: " + _("This is not a valid length"))
-
-            if cy is None:
-                cy = Length(0)
-            else:
-                if not cy.is_valid_length:
+                if cy is None:
+                    cy = Length(0)
+                elif not cy.is_valid_length:
                     raise SyntaxError("cy: " + _("This is not a valid length"))
-            # do we have something like 'polyshape 3 4cm' ? If yes, reassign the parameters
-            if radius is None:
-                radius = cx
-                cx = Length(0)
-                cy = Length(0)
+                cx = cx.value(
+                    ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
+                )
+                cy = cy.value(
+                    ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
+                )
+                if radius is None:
+                    radius = Length(0)
+                radius = radius.value(
+                    ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
+                )
+                if startangle is None:
+                    startangle = Angle.parse("0deg")
+
+                starpts = [(cx, cy)]
+                if corners==2:
+                    starpts += [(cx + cos(startangle.as_radians) * radius, cy + sin(startangle.as_radians) * radius)]
+
             else:
-                if not radius.is_valid_length:
-                    raise SyntaxError("radius: " + _("This is not a valid length"))
-
-            if startangle is None:
-                startangle = Angle.parse("0deg")
-            if inscribed is None:  # Use circumscribed circle by default
-                inscribed = False
-
-            if alternate_seq is None:
-                if radius_inner is None:
-                    alternate_seq = 0
-                else:
-                    alternate_seq = 1
-
-            if radius_inner is None:
-                radius_inner = radius
-            else:
-                if not radius_inner.is_valid_length:
+                if cx is None:
                     raise SyntaxError(
-                        "radius_inner: " + _("This is not a valid length")
+                        _(
+                            "Please provide at least one additional value (which will act as radius then)"
+                        )
                     )
-            if density is None:
-                density = 1
-            if density<1 or density > corners:
-                density = 1
-
-            cx = cx.value(
-                ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
-            )
-            cy = cy.value(
-                ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
-            )
-            radius = radius.value(
-                ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
-            )
-            if (
-                isinstance(radius, Length)
-                or isinstance(cx, Length)
-                or isinstance(cy, Length)
-            ):
-                raise SyntaxError
-            if inscribed:
-                radius = radius / cos(pi / corners)
-
-            radius_inner = radius_inner.value(ppi=1000, relative_length=radius)
-            if isinstance(radius_inner, Length):
-                radius_inner = radius
-            if alternate_seq < 1:
-                radius_inner = radius
-
-            # print("These are your parameters:")
-            # print("Vertices: %d, Center: X=%.2f Y=%.2f" % (corners, cx, cy))
-            # print("Radius: Outer=%.2f Inner=%.2f" % (radius, radius_inner))
-            # print("Inscribe: %s" % inscribed)
-            # print(
-            #    "Startangle: %.2f, Alternate-Seq: %d"
-            #    % (startangle.as_degrees, alternate_seq)
-            # )
-
-            pts = []
-            myangle = startangle.as_radians
-            deltaangle = tau / corners
-            ct = 0
-            for j in range(corners):
-                if ct < alternate_seq:
-                    # print("Outer: Ct=%d, Radius=%.2f, Angle=%.2f" % (ct, radius, 180 * myangle / pi) )
-                    thisx = cx + radius * cos(myangle)
-                    thisy = cy + radius * sin(myangle)
                 else:
-                    # print("Inner: Ct=%d, Radius=%.2f, Angle=%.2f" % (ct, radius_inner, 180 * myangle / pi) )
-                    thisx = cx + radius_inner * cos(myangle)
-                    thisy = cy + radius_inner * sin(myangle)
-                ct += 1
-                if ct >= 2 * alternate_seq:
-                    ct = 0
-                if j == 0:
-                    firstx = thisx
-                    firsty = thisy
-                myangle += deltaangle
-                pts += [(thisx, thisy)]
-            # Close the path
-            pts += [(firstx, firsty)]
+                    if not cx.is_valid_length:
+                        raise SyntaxError("cx: " + _("This is not a valid length"))
 
-            starpts = [(pts[0][0], pts[0][1])]
-            idx = density
-            while idx != 0:
-                starpts += [(pts[idx][0], pts[idx][1])]
-                idx += density
-                if idx >= corners:
-                    idx -= corners
-            if len(starpts) < corners:
+                if cy is None:
+                    cy = Length(0)
+                else:
+                    if not cy.is_valid_length:
+                        raise SyntaxError("cy: " + _("This is not a valid length"))
+                # do we have something like 'polyshape 3 4cm' ? If yes, reassign the parameters
+                if radius is None:
+                    radius = cx
+                    cx = Length(0)
+                    cy = Length(0)
+                else:
+                    if not radius.is_valid_length:
+                        raise SyntaxError("radius: " + _("This is not a valid length"))
+
+                if startangle is None:
+                    startangle = Angle.parse("0deg")
+                if inscribed is None:  # Use circumscribed circle by default
+                    inscribed = False
+
+                if alternate_seq is None:
+                    if radius_inner is None:
+                        alternate_seq = 0
+                    else:
+                        alternate_seq = 1
+
+                if radius_inner is None:
+                    radius_inner = radius
+                else:
+                    if not radius_inner.is_valid_length:
+                        raise SyntaxError(
+                            "radius_inner: " + _("This is not a valid length")
+                        )
+                if density is None:
+                    density = 1
+                if density<1 or density > corners:
+                    density = 1
+
+                cx = cx.value(
+                    ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
+                )
+                cy = cy.value(
+                    ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
+                )
+                radius = radius.value(
+                    ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
+                )
+                if (
+                    isinstance(radius, Length)
+                    or isinstance(cx, Length)
+                    or isinstance(cy, Length)
+                ):
+                    raise SyntaxError
+                if inscribed:
+                    radius = radius / cos(pi / corners)
+
+                radius_inner = radius_inner.value(ppi=1000, relative_length=radius)
+                if isinstance(radius_inner, Length):
+                    radius_inner = radius
+                if alternate_seq < 1:
+                    radius_inner = radius
+
+                # print("These are your parameters:")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Sophist-UK** [comment]
Looks interesting. This is IMO one of the great powers of MK - the idea that if you want a way to do some clever stuff it is really easy to add a command to achieve it.

**Comment by tiger12506** [comment]
I love this!
I like that you comment your code as well, very nice work!
Typo where you have "eerything", just fyi,  if you want to fix it.

Looking forward to those extended graphical tools!

**Comment by tiger12506** [comment]
The additional items are also really cool. A few thoughts:
1) I think the existing polygon command should be renamed to polyshape, and yours should be polygon. (This may be controversial, don't do without the other meerkittens weighing in)
2) Your start/end angles are strictly in screen coords without correcting for y-down (i.e. increasing angle is clockwise). I'm not saying this is wrong, but it's something we should consider keeping consistent across all tools and I haven't checked to see if there are any other examples that do this correction, but personally I'd like the correction to be made so that it is consistent with general trig (increasing angle is anti-clockwise)
3) The radius specified to star and polyshape make it easy to create an inscribed version, but it might be nice to specify both ways (have parameter to allow the polygon to sit outside the circle of radius r)

**Comment by Sophist-UK** [comment]
> The additional items are also really cool.

👍 

> 1. I think the existing polygon command should be renamed to polyshape, and yours should be polygon. (This may be controversial, don't do without the other meerkittens weighing in)

There are probably few users using this, so it seems like a good suggestion. I just wish that I had improved the help command already so it would be easier to spot.

> 2. Your start/end angles are strictly in screen coords without correcting for y-down (i.e. increasing angle is clockwise). I'm not saying this is wrong, but it's something we should consider keeping consistent across all tools and I haven't checked to see if there are any other examples that do this correction, but personally I'd like the correction to be made so that it is consistent with general trig (increasing angle is anti-clockwise)



> 3. The radius specified to star and polyshape make it easy to create an inscribed version, but it might be nice to specify both ways (have parameter to allow the polygon to sit outside the circle of radius r)
> 4. ~This might be a terrible idea. Does it make sense to add (inscribed angle)/2 to polyshape start angle? I don't know if I really like that in the argument of cleanliness/technically correctness, but might be more what users expect to see...
>    `polyshape 5 5in 5in 1in 0` would then mean `polyshape 5 5in 5in 1in 54`
>    Maybe it doesn't matter if it's going to be a GUI tool later, in which case this might be harmful instead.~ I decided that this is really a bad idea after all.

**Comment by tiger12506** [comment]
> Not sure that the right click rotate function does that.

Right-click rotate assumes y+ is down, so +10 degrees is a rotation in clockwise direction (opposite of what I was hoping for). So, for this pull request probably should leave it as is to match meerk40t.

Yeah, the Transform pane has the rotations labeled this way as well. So I guess the convention in meerk40t is to assume +y down = +rotation in clockwise direction.

Kindly disregard my item 2 in list above.

**Comment by tatarize** [comment]
Stars can have a few different other parameters.

You can have the stars cross themselves as you do these are the typical polygram type stars, your regular pentagrams for example.

You can also do regular isotoxal stars which is to say you alternate between major radius and minor radius this can be set with a phase. Your typical minor radius would be at 50% between the last two major radius parts. However this could be anywhere along the values from 0-1. So you could basically make a gear or a saw-tooth like structures fairly easily.

Furthermore along this same path you can have your spirograms™ which are your epicycloid values that you sample several times each time. Here also have major and minor radii but this is due to spirograms being made by marking an offset position from a circle that is travelling outside, or inside another circle.

There are also in addition to arguments `command_option` these are part of the commands but are not specifically required So rather than needing to set an inscribe value, you can set one with an option and not require it to exist.

I also notice that many of your examples have you rotating the polyshapes you make slightly to make really cool patterns. This is basically a circle-copy `circ_copy <copies> <radius> <startangle> <endangle> <rotate>` with a radius of 0 right? 

I have no trouble with any of these commands but they seem a bit like they should fit in a bit better. There's a lot of ways of doing this. Star can have a few different things, and if you don't specify an internal angle you basically just have a ngon anyway.

The regular polygon command makes a polygon, this however, makes a regular polygon.

**Comment by tatarize** [comment]
`circ_copy <copies> <radius> <startangle> <endangle> <rotate>`

I would expect this to be `array`, `circle-array` or `step-repeat` which is used in some code things like guilloches building apps. Which are about as advanced as these sorts of things can get.  I would expect `startangle` and `endangle` to be options rather than arguments. See `rect` for the `-rx` `-ry` values. And I would expect the rotate to be a boolean which is *usually* an option which calls `action='store_true'`  with a `type=bool` and this means it takes no values. But a flag -z or `--no_rotate`  anywhere in the command would set the bool to True.

Then you could do something like `circle 1in 1in array 5 3in` which would do a 1 inch circle and then make 5 of them around another 3 inch circle. It would know that since you didn't specify a start or end angle or a sweep amount that it's doing 1 circle worth.

If you specify a sweep of 1 turn rather than end angle you can do some interesting stuff since you could do say 200 turns.  I'm not opposed here I'm just saying there's a lot of things this could be and if it turns popular or gets used a bunch it would get harder to tweak if better API was established.

circ-array

For examples of this Lightburn has this sort of feature and does it based on the step-angle and count. So if you're stepping by 90 degrees or 0.25turn and you have 4 copies in your array. Then clearly you get flipped corners or whatever. If you step by 3.29 degrees and you are making 200 copies you will get 658 degrees of total without any of them ever overlapping since these do not evenly divide.

**Review by tatarize — DISMISSED** [review]
Generally this stuff is kind of hidden. And while I think with command chaining and command_options this could be simplified. Especially the 1 = Yes should be done as a `store_true` boolean option. But, nothing outwardly changes for the bulk of users so if you want this merged... it's fine. Deprecating commands isn't too hard especially in the switch to 0.8.x where some of these would need to be done as visual tools controlled with the mouse.

**Comment by tatarize** [comment]
Also, and while not very important I think maybe if you're going to have position of the original grid value be a thing it should maybe accept a rowXcolumn value. I get this isn't a unit anything uses but if you accept a string for that option (shouldn't really be an argument either), you'd be able to say 3,3 which would get sent to you as `"3,3"` a string and then you could parse it by splitting at the comma to get 3 and 3. Which would the position of the first copy. And create your grid around that.

**Comment by jpirnay** [comment]
@tatarize 
> Stars can have a few different other parameters.
> 
> You can have the stars cross themselves as you do these are the typical polygram type stars, your regular pentagrams for example.
> 
> You can also do regular isotoxal stars which is to say you alternate between major radius and minor radius this can be set with a phase. Your typical minor radius would be at 50% between the last two major radius parts. However this could be anywhere along the values from 0-1. So you could basically make a gear or a saw-tooth like structures fairly easily.

That's actually quite straightforward:
grafik
grafik

**Comment by jpirnay** [comment]
> Also, and while not very important I think maybe if you're going to have position of the original grid value be a thing it should maybe accept a rowXcolumn value. I get this isn't a unit anything uses but if you accept a string for that option (shouldn't really be an argument either), you'd be able to say 3,3 which would get sent to you as `"3,3"` a string and then you could parse it by splitting at the comma to get 3 and 3. Which would the position of the first copy. And create your grid around that.

grafik

**Comment by jpirnay** [comment]
@tatarize 
> For examples of this Lightburn has this sort of feature and does it based on the step-angle and count. So if you're stepping by 90 degrees or 0.25turn and you have 4 copies in your array. Then clearly you get flipped corners or whatever. If you step by 3.29 degrees and you are making 200 copies you will get 658 degrees of total without any of them ever overlapping since these do not evenly divide.
No problem :-)
grafik
You can have that one as well...

**Comment by jpirnay** [comment]
@tatarize 
> Generally this stuff is kind of hidden. And while I think with command chaining and command_options this could be simplified. Especially the 1 = Yes should be done as a `store_true` boolean option.

grafik
That's done as well..
So I will commit these changes and ask for a re-review.

**Comment by tatarize** [comment]
`star 5 2 1in` crashes.

**Comment by tatarize** [comment]
image

I'm not sure if 'hops` makes sense. This is the skip value for the connecting the points of the stars yes? star 6 2 makes a triangle since it never hit the other values. This might be correct but it's a little bit strange.

**Comment by tatarize** [comment]
Inscribed and startangle options for star do not set a second value this should be like "s" or "i" so that `-ia 20grad` properly means inscribed=True and startangle gets set at 20grads.

**Comment by tatarize** [comment]
* command `polyshape 5 2in 2in 1in fill blue` crashes and fails as `polyshape` returns a list of list-tuples rather than a polygon or path.

**Comment by jpirnay** [comment]
> 

Any combination of two even numbers will not connect all vertices. That's the reason I mentioned the odd/even or even the coprime requirements in my lead comment for this PR.
When this is integrated I will work on some GUI to help with all of these options/complexities /requirements.
(My priorities might change slightly, as I tend to work on those things that I miss/ that annoy me most first :-))

**Comment by tatarize** [comment]
* Degenerate stars with `hops` set to 1 are actually regular polygons. These are not degenerate and would perhaps tread on your `polyshape` command (or allow you to create subsets) but `hops=1` shouldn't be an error.
* The term `hops` here is non-typical within the nomenclature of stars. Apparently density is the preferred term. ( https://en.wikipedia.org/wiki/Density_(polytope) )

**Comment by tatarize** [comment]
Your code detecting degenerate polygrams is incomplete. Actually any values for `p` and `q` which are not coprimes. For example, 12 and 9 are not both even but because 3 is a common factor of both 12 and 9 we get a triple wound rectangle. Other than comparing all the common factors of each I don't know any way to detect the coprimes and while degenerate this is what the user asked.

**Comment by tatarize** [comment]
If you wanted to merge the star and regular_polygon type commands you could allow polyshape to accept the Schläfli value. So `polyshape 3 1in` is a triangle with sides of 1in, whereas `polyshape 5/2 1in` is a 5/2-star with sides of 1in. You could also, if the designs were made with a turtle like graphics system permit spirolaterals by allowing the radius to be given as a list or a regular reducing integer count like most such shapes.

I'm not proposing or demanding such things but I find it's often helpful to overbuild thing so you can simplify things. `polyshape 14/6 3in,2in,1in` or some overbuilt nomenclature as such could build the varieties of shapes that you get if each line-segment has a pattern of re-occurring lengths ( the first length is 3in then 2in then 1in then 3in again, etc). You can make some really great patterns this way, but you can also make all the regular polygons and regular polygrams, excluding the isotoxal stars which are simple polygons (non-self crossing). These are usually defined by a major and minor radius, point-count and phase among other things. 5 pointed stars would have 10 points usually with like a half-sized minor radius about equidistant between points.

**Comment by tatarize** [comment]
grid-origin2,2

Trying `grid 3 3 -o 2,2` caused the grid to become degenerate. I expected my 12/9-star to either be at the center of the grid or in the lower-right hand corner.

**Comment by tatarize** [comment]
`Position of copy in matrix (e.g '2,2' or '4,3')` -- might better read `position of original within grid`. I'm usually the first person to shrug at 'might confuse user' with regard to console commands but the command is grid and calling the grid a grid is fine even while matrix isn't wrong it's a bit confusing since it's a new term that is just elsewhere used to describe affine transformation matrices.

**Review by tatarize — DISMISSED** [review]
While I did spend a lot of time punching holes in this nothing new gets broken. The `--origin` value for grid doesn't quite work, but it's optional. And most of the rest of the stuff somewhat works. I'd prefer a simpler syntax and maybe some more geometrically consistent terms but such things can always be deferred to later so long as nothing absolutely builds on this in the gui, little background console commands aren't a problem. Though Sophist would certainly be less laissez-faire.

**Comment by jpirnay** [comment]
> * Degenerate stars with `hops` set to 1 are actually regular polygons. These are not degenerate and would perhaps tread on your `polyshape` command (or allow you to create subsets) but `hops=1` shouldn't be an error.
> 
>     * The term `hops` here is non-typical within the nomenclature of stars. Apparently density is the preferred term. ( https://en.wikipedia.org/wiki/Density_(polytope) )

grafik

**Comment by jpirnay** [comment]
@tatarize 
> Trying grid 3 3 -o 2,2 caused the grid to become degenerate. I expected my 12/9-star to either be at the center of the grid or in the lower-right hand corner.

Well it should be called with grid 3 3 -o "2,2" , as otherwise the last ,2 will be recognized as the value for x by the superordinated routines. I am not sure whether this intentional or not... You can easily spot this behaviour if you state 
```
grid 3 3 100% 100% -o 2,2 
```

**Comment by jpirnay** [comment]
@tatarize 
> And most of the rest of the stuff somewhat works. I'd prefer a simpler syntax and maybe some more geometrically consistent terms but such things can always be deferred to later so long as nothing absolutely builds on this in the gui, little background console commands aren't a problem. Though Sophist would certainly be less laissez-faire.

Well my intention is never to provide that nearly works ;-) And I am very much aware that I am a newbie to MK and have still to understand the finer details of the logic inside it. So I take any constructive criticism gracefully and gratefully :-)

**Comment by tatarize** [comment]
> Well it should be called with grid 3 3 -o "2,2" , as otherwise the last ,2 will be recognized as the value for x by the superordinated routines. I am not sure whether this intentional or not... You can easily spot this behaviour if you state

Hm. That makes sense. The parsing routine must be interpreting the commas. I am pretty sure I didn't use the commas in that fashion or really expect them to be done like that. I could see needing them for the polyline or polygon commands. But, clearly LONG, PARAM, and SKIP in the parsing routine consider `,` as a delimit character. Given that it might be preferable to use a different delimit character eg. `2x2`. I suppose properly I consider altering the parsing to permit setting a number of tokens taken by a particular value-set. Usually in CLI stuff there's a NARG call which sets the number of arguments and then passes that along.

Okay, turns out I already did add in `nargs`. You can specify that origin is `type=float` with `nargs=2` and it will quite clearly give you a list of size 2 for the given arguments from --origin. Hm. Nifty feature... don't think I ever used it.

**Comment by tatarize** [comment]
And oddly it *only* works for LONG so it weirdly works for --origin but not for -o. I think that's in error, but using nargs is probably the way to go there. I'll fix the help and the short values using nargs.

**Comment by tatarize** [comment]
I fixed up nargs and submitted it corrected in your grid modification. The help is also corrected to display the options help a bit better for that.

**Comment by jpirnay** [comment]
Added circ_array for completeness and amended the initial comment accordingly.
@Sophist-UK 
On a sidenote: I would add these explanations to the wiki - but where?

**Comment by Sophist-UK** [comment]
> @Sophist-UK On a sidenote: I would add these explanations to the wiki - but where?

@jpirnay Jen's - At this stage create a Tech page and add it to the sidebar. The difficulty bit is getting the page written - it is ready to move it elsewhere later.

**Comment by tatarize** [comment]
Writing documentation at this stage is premature. That's why there's no documentation for anything because so much of it is likely to change. Though some things are certainly starting to get down to reasonable.

File: meerk40t/core/elements.py [hunk 1]
```diff
@@ -3258,21 +3272,562 @@ def element_grid(
...
+            startangle=None,
+            inscribed=None,
+            data=None,
+            **kwargs,
+        ):
+
+            if corners is None:
+                raise SyntaxError
+            if corners < 5:
+                raise SyntaxError(_("A star needs to have at least 5 corners"))
```
tatarize, line 3732: Consider permitting `star 3 1`  shapes. While these are generally just going to be polyshapes. I see a lot of overlap between these commands.  It might be correct to combine these into just a `shape` command to make a wide variety of regular shapes. Where your needed values are `x,y` location and `points` and `radius` maybe permitting a wider sets of values since many of the other properties would be optional. Currently the `star` command cannot create isotoxal stars. The `shape`  is about as generic as things can get. And you could probably let people specify weird properties that tend to permit a lot of different geometry.

meerk40t/core/elements.py
```diff
@@ -3511,13 +4078,24 @@ def element_polygon(args=tuple(), **kwargs):
         )
         def element_polyline(command, channel, _, args=tuple(), **kwargs):
```
tatarize, line 4079: The help for polyline still reads `(float, float)*` and likely is (x y)* for the more correct value there.

meerk40t/core/elements.py
tatarize: If you decided to go with `shape` as a command. You might actually want to reduce corners to permit omission and default to 1. While this is actually just a dot. It's actually the case that making dots is functionally useful and we sort of need a method to do that since we actually have code to go to the location of a dot and fire the laser for a set period of time.

meerk40t/core/elements.py
```diff
@@ -3258,21 +3272,562 @@ def element_grid(
...
+            if (
+                isinstance(radius, Length)
+                or isinstance(x_pos, Length)
+                or isinstance(y_pos, Length)
+            ):
+                raise SyntaxError
+            if inscribed:
+                radius = radius / cos(pi / corners)
+
+            radius_inner = radius_inner.value(ppi=1000, relative_length=radius)
```
tatarize, line 3637: Clever. Shows a firm understanding of the use of relative_length there for % values.

meerk40t/core/elements.py
jpirnay: Have changed that - on another note:
how does command trailing work, I though one command would look at all the parameters it would require and the remainder of the line would then take the result of that command use its as the base for further commands.
If you do have a command like polygon or polyline with an arbitrary amount of parameters this approach fails.
Any idea how to overcome that? Is there a pipe / end of command parameter in place that forces an end of parameters for a command?
(And still the way of generating the polyline / polygon elements needs to be amended  - will do this in this PR)

meerk40t/core/elements.py
```diff
@@ -3258,21 +3272,562 @@ def element_grid(
...
+                ppi=1000, relative_length=bed_dim.bed_width * MILS_IN_MM
+            )
+            if (
+                isinstance(radius, Length)
+                or isinstance(x_pos, Length)
+                or isinstance(y_pos, Length)
+            ):
+                raise SyntaxError
+            if inscribed:
+                radius = radius / cos(pi / corners)
```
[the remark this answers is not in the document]
jpirnay: The diameter of the circumscribed circle for a given segment length s, and corners n is:
```` d_OC= s * csc ( pi / n )  ````
for the inscribed it is :
``` d_IC= s * cot ( pi / n ) ````
so the ratio of 
``` 
d_IC/d_OC = (s * cot (pi /n ) ) / (s * csc(pi/n) ) 
= cot(pi/n) / csc (pi/n) 
= (cos(pi/n) / sin(pi/n)) / (1 / sin(pi/n)) 
= cos(pi/n)
```

File: meerk40t/core/elements.py [hunk 5]
```diff
@@ -3258,21 +3272,562 @@ def element_grid(
...
+            center_x = (bounds[2] + bounds[0]) / 2.0 - radius
+            center_y = (bounds[3] + bounds[1]) / 2.0
+
+            # print ("Copies: %d, Radius: %.1f" % (copies, radius))
+            # print ("Center: %.1f, %.1f" % (center_x, center_y))
+            # print ("Startangle, Endangle, segment_len: %.1f, %.1f, %.1f" % (180 * startangle.as_radians / pi, 180 * endangle.as_radians / pi, 180 * segment_len / pi))
+
+            for cc in range(copies):
+                # print ("Angle: %f rad = %f deg" % (currentangle, currentangle/pi * 180))
+                if cc>0:
```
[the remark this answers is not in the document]
jpirnay: Done

meerk40t/core/elements.py
```diff
@@ -3258,21 +3272,562 @@ def element_grid(
...
+
+            self.context.signal("refresh_scene")
             return "elements", data_out
 
+        @context.console_argument("copies", type=int, help=_("Number of copies"))
+        @context.console_argument("radius", type=Length, help=_("Radius"))
+        @context.console_argument("cx", type=Length, help=_("X of center to circle around"))
+        @context.console_argument("cy", type=Length, help=_("Y of center to circle around"))
+        @context.console_argument("startangle", type=Angle.parse, help=_("Start-Angle"))
+        @context.console_argument("endangle", type=Angle.parse, help=_("End-Angle"))
```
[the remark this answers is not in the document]
jpirnay: a) Agree, cy and cy stem from an earlier iteration and make no more sense - removed
b) Not sure I understand what you are missing, I think you can do all with the different parametersets. Especially the deltaangle parameter will just regard the startangle and iterate around the center how long it takes:
image

jpirnay:
Well just a quick interim release to help you play with shape, but yes it provides some new interesting results:
image

jpirnay:
Combined polyshape and star into shape. Allow 1 corner shape and 2 corner shape for completeness although these collapse into a single point and a line.
image

jpirnay:
Updated the main description for this PR to make it consistent with the current implementation

meerk40t/core/elements.py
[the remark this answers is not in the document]
tatarize: Not missing something. I'm pointing out there are other parameters that achieve the same thing. You can parameterize code in a *lot* of different ways especially with regard to circles. `Sweep`, `start`, and `total_copies` gives you the needed parameters. So does `delta` `start` and `total_copies` since delta is just `sweep / total_copies` and you might be able to solve some-things with `start` `end` and `total_copies` but in this case there's an infinite number of different sweep values that could suffice, and 2 of them are going to be less than 1 turn.

tatarize:
The alternative parameterizations are a bit weird but you could want a shape based on enough information but not necessarily have the information your code is asking for, but that code could be derived. This matters a bit if we're using the sweep angle rather than delta_angle and number of vertexes. But, also, imagine I don't want to use radius. I want a hexagon with side lengths equal to 1in, how do I do that? I mean clearly we can determine that `r = side_length / (2 * sin(tau / (2*points)))`.

The question here is how does the command `shape` determine this stuff for alternate parameterizations. If I give you sweep rather than delta_angle this is still solvable. If I give you side_length rather than radius, this is solvable. But, how do we specify which bits of information you are given and which you are not? --- I think the answer is that you should accept `x, y, points` and the rest of the properties likely as options rather than arguments. Maybe `x, y, points, distance` where distance is assumed to be radius unless you set an option flag to say it's side_length just as you assume it's circumscribed unless flagged as inscribed.

meerk40t/core/elements.py
tatarize: The `@console_command` will uses a dictionary for everything that wasn't matched and one of the parameters it gives you is `args` this is all the arguments for the command without parsing. So if a command is not intended to be chained like `polygon` and `polyline` they can accept `args` which is just the tuple of split values in the `remainder` without context. The `remainder` is somewhat similar in that it's the string of `remaining` unparsed values after or within the current command.

The chaining of the commands is predicated on consuming all the commands successfully. So if we have a command that has 4 arguments then we know that `command a0 a1 a2 a3 <new_command>` is valid we could do chain our `shape` command into a `grid` command or add a `fill` or `stroke` as part of the current command. `circle 3in 3in 1in fill red` will chain the circle command which returns an `elements` out_type with `fill` which accepts an `elements` in_type. So we make a circle (adding it to the scene) and then give that circle a `red` fill value.

For some commands we don't know how many arguments we will have and in that case, chaining like this becomes impossible. If we use `polyline` we can't set a fill value at the end.

tatarize: It might be possible to run some validation on arguments and if the casting of the value causes a ValueError stop and call that arguments and could be added in as nargs="*" or so, assuming the type isn't `str` or something where all the values would be valid there.

meerk40t/core/elements.py
```diff
@@ -3258,21 +3272,477 @@ def element_grid(
...
+            "i",
+            type=bool,
+            action="store_true",
+            help=_("Shall the polygon touch the inscribing circle?"),
+        )
+        @context.console_option(
+            "radius_inner",
+            "r",
+            type=Length,
+            help=_("Alternating radius for every other vertice"),
```
tatarize, line 3539: `vertex` is more common for this but this is valid.

meerk40t/core/elements.py
```diff
@@ -3258,21 +3272,477 @@ def element_grid(
...
+                    radius = cx
+                    cx = Length(0)
+                    cy = Length(0)
+                else:
+                    if not radius.is_valid_length:
+                        raise SyntaxError("radius: " + _("This is not a valid length"))
+
+                if startangle is None:
+                    startangle = Angle.parse("0deg")
+                if inscribed is None:  # Use circumscribed circle by default
```
tatarize, line 3632: All this does is scale the radius (and maybe inner_radius?) so you don't need an explicit value for it, since the truthiness of None is False.

tatarize (dismissed):

jpirnay:
> The alternative parameterizations are a bit weird but you could want a shape based on enough information but not necessarily have the information your code is asking for, but that code could be derived. This matters a bit if we're using the sweep angle rather than delta_angle and number of vertexes. But, also, imagine I don't want to use radius. I want a hexagon with side lengths equal to 1in, how do I do that? I mean clearly we can determine that `r = side_length / (2 * sin(tau / (2*points)))`.
> 
> The question here is how does the command `shape` determine this stuff for alternate parameterizations. If I give you sweep rather than delta_angle this is still solvable. If I give you side_length rather than radius, this is solvable. But, how do we specify which bits of information you are given and which you are not? --- I think the answer is that you should accept `x, y, points` and the rest of the properties likely as options rather than arguments. Maybe `x, y, points, distance` where distance is assumed to be radius unless you set an option flag to say it's side_length just as you assume it's circumscribed unless flagged as inscribed.

@tatarize  
That's easily done - but can we please reach consensus about the names and sequence of the parameters?
I would still assume that the majority of cases would use the 'radius' interpretation, so that should take precedence for the naming convention.
I follow you guidance for 'shape corners cx cy' vs 'shape cx cy corners' i have no emotional attachment to any variant ;-)
image

jpirnay:
@tatarize 
I think I still have bug in my code where I could use some help:  if I am using an option in it's abbreviated form, it's getting recognised but creates an error message...
image

jpirnay:
@tatarize 
On the short param issue: I believe in kernel.py
```
 def _cmd_parser(text: str) -> Generator[Tuple[str, str, int, int], None, None]:
        pos = 0
        limit = len(text)
        while pos < limit:
            match = _CMD_RE.match(text, pos)
            if match is None:
                break  # No more matches.
            kind = match.lastgroup
            start = pos
            pos = match.end()
            if kind == "SKIP":
                continue
            elif kind == "PARAM":
                value = match.group()
                yield kind, value, start, pos
            elif kind == "QPARAM":
                value = match.group()
                yield "PARAM", value[1:-1], start, pos
            elif kind == "LONG":
                value = match.group()
                yield kind, value[2:], start, pos
            elif kind == "OPT":
                value = match.group()
                for letter in value[1:]:
                    yield kind, letter, start, start + 1
                    start += 1
```
the last lines need to be amended to:
```
            elif kind == "OPT":
                value = match.group()
                for letter in value[1:]:
                    yield kind, letter, start, pos
                    start += 1
```

tatarize (approved):

PR #869, opened 2022-03-01
Add raster optimisation on/off switch (no gui)

Sophist-UK:
I haven't added a GUI because I am hopeful that this will be robust and just work correctly - but just in case, here is a setting that you can use (via console) to turn it off and go back to a single raster image per Raster op.


tiger12506 (approved):
Seems pretty straightforward.

Sophist-UK:
Bump

tatarize (approved):

PR #872, opened 2022-03-02
Fix ansi check on windows 8.1, add dark theme support

Sophist-UK:
1. Fixes windows version detection and avoids crash on Windows 8.1 and handles Windows 11.
2. Handles darkmode colours (as reported by @joerlane .


tiger12506:
I'm fine with this, it makes sense. Has anyone checked if the WinDLL change still works from within a build? Or is that what I'm supposed to be checking?

Sophist-UK:
Sorry Tiger. In the issue user is still diagnosing the solution. My fault for not marking this draft.

Sophist-UK:
Bump

Sophist-UK:
This is good to go. The actual change to how we load the dll needs to be in a different PR.

tatarize (approved):

Sophist-UK:
Should be good for review again now. Foreground colours are now correct on Linux (and should be OK on Mac too). Also done something with scroll bars but I could not recreate the scrolling issue that @tiger12506 reported so we will need to see if this fixes that.

Sophist-UK:
If this doesn't completely fix @tiger12506 's scroll issue, I suggest that it is merged anyway as it is much closer than before.

tiger12506 (dismissed):
This seems to have fixed the issues I experienced. Nice work.

tatarize (approved):

PR #874, opened 2022-03-04
Draft: Airassist

jpirnay:
### This a draft PR for discussion mostly

I've added (mostly for my own benefit) some basic air-assist support.
It adds additional special operations (Air Assist On, Air-Assist off) 
grafik
as well as a console command
grafik.

There is a minimal GUI apart from issuing the command in the console or via the Job Execution inside the laser-panel:
grafik

There are currently three (the RPI one is commented out) methods to switch AirAssist on / off:
- Issuing a shell command, so you could call whatever you like
- A mqtt connection to a broker to publish values to turn e.g. a Tasmota Switch on (my use-case to turn on/off the air-pump)
- Switching a relay via one of the Raspberry PIs GPIO pins (commented out as I need to build a relay board and a raspberry installation first)
What is missing:
a) Any GUI to make the parameters customisable - currently they are hardcoded in core/airassist.py - so if you want to test it in your environment you have to amend them in the code.
b) Better recovery from errors (my MQTT server is rather sluggish and needs sometimes several attempts to acknowledge the command)
c) Reading the current status of the Air-Assist
d) some signals to snychronize the button in laserpanel.py to the commands in the console

This is more a hack intended to work for my environment than anything else (and it does work here - so happy to be able to programmatically switch my airpump on and off) . So some feedback on the structure how to tie the events to the drivers etc would be welcome. I will stop to invest more effort into this until there is a consensus whether this approach would make any sense or not (you could argue that's what MK plugins are for and I don't disagree, but you would still need the integration part as (inelegantly) done here)

### DIFF
--- a/meerk40t/core/airassist.py
+++ b/meerk40t/core/airassist.py
@@ -4,85 +4,41 @@ AirAssist contains basic code to externally turn on / off air assist .
 """
 from subprocess import run
 from threading import Timer
-
+from time import sleep
+from io import open
+import paho.mqtt.client as mqtt
 
 AA_METHOD_NONE = 0
 AA_METHOD_MQTT = 1
 AA_METHOD_SHELL = 2
 AA_METHOD_RPI_GPIO = 3
 
-
-class AirAssist:
-
+# Overarching class with logic and timer control
+class onoff:
     myTimer = None
     overshoot = 5.0
-    shell_command_on = "turn_aa_on.cmd"
-    shell_command_off = "turn_aa_off.cmd"
 
     def __init__(self):
-        self._method = AA_METHOD_NONE
         self._status = False
 
-    # function to get value of _method
-    @property
-    def method(self):
-        return self._method
+    def cmd_on(self):
+        return
 
-    # function to set value of _method
-    @method.setter
-    def method(self, a):
-        if (a < AA_METHOD_NONE) or (a > AA_METHOD_RPI_GPIO):
-            raise ValueError("Sorry this AirAssist-method isn't supported")
-
-        if a == AA_METHOD_NONE:
-            pass
-        elif a == AA_METHOD_MQTT:
-            print("MQTT-Method called, need to establish connectivity")
-        elif a == AA_METHOD_SHELL:
-            print("Shell-Method called")
-        elif a == AA_METHOD_RPI_GPIO:
-            print("RPI-Method called")
-
-        self._method = a
+    def cmd_off(self):
+        return
 
     @property
     def status(self):
         return self._status
 
-    @method.setter
+    @status.setter
     def status(self, a):
         if a:
             self._status = True
+            self.cmd_on()
         else:
             self._status = False
-        if self._method == AA_METHOD_NONE:
-            print(
-                "AirAssist ignored, would have been set to %s"
-                % ("ON" if self._status else "OFF")
-            )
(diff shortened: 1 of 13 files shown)


tatarize:
* Turning things on and off isn't usually done with integers. Usually you'd toggle things or set a parameter. Eg. `air` gives the known air status. `air on` turns the air on. `air off` turns the air off. `air toggle` toggles the air. Not that there's anything wrong with your command there.
* Adding this to the laser-panel may be a bit strange since very few people are going to have access to controlling their air assist.
* As a general rule and for security reasons I don't actually let somebody using meerk40t issue shell commands or execute arbitrary code. While this is doable this seems a bit strange.
* There is within the Ruida Controller the ability to turn the air on and off. This is actually done as part of the operations typically and so you can raster without air, and cut with air if that's something you want to do. This also tends to allow you to set it as part of specific settings for example you might not want air on for acrylic.

I believe the best way to do this would be a plugin in 0.8.x. Since you can dynamically create a pane, and because the methods you have for turning on the air assist are not quite exhaustive. You have your stuff setup in a special and thus have some notable personal abilities to hack this in there but generally wouldn't be that useful for everybody since most of us do not have air that can be toggled. 

The way I would set this in 0.8.0.
1) Big button that says air as a pane that is created dynamically. I wouldn't hack the laser panel. 
2) Add a console command for `air on` `air off` and `air toggle`, have this command send a signal called `air` with the 0 or 1 value of the new state.
3) Have you big button listen to the air signal to reflect the current state.
4) Have a lifecycle-delegate of the laser device class you are using.
5) Add in your very specific method of turning your air on and off calling the MQTT server. Have this toggled by the signal value.

Sophist-UK:
Is there a standard way that everyone connects their sir assist to the M2 nano so that it can be programmatically controlled. Because if not, there is a problem.

Solution: create a mk wiki page describing how to set the hardware up to do this so that the command work h i.e. create a de facto standard.

jpirnay:
My solution is deceptively easy, I have a small aquarium pump that I plug into a smart socket - most of them can be reprogrammed with [Tasmota](https://github.com/arendst/Tasmota)  (e.g. the SOnOff sockets).
These sockets can be switched on / off via mqtt.
Alas there is no standard, and afaik the M2Nano boards don't have an additional programmable IO-Port (and that might need some soldering too).
There's the very same issue for 3d-Printets how to turn them on (or better off after a lengthy job). [Octoprint](https://octoprint.org) provides a PSU plugin for that with additional plugins to use different switching mechanisms.

Sophist-UK:
I have a feeling that the stock PSU on some machines may have some extra connector that you can tie to a relay and turn left on and off. (I have a feeling I read about this a year or so ago when I first got my K40.

tatarize:
(reopened, draft + discussion, though yeah, not going much of anywhere as a PR but maybe a plugin).

This is basically the quintessential example of what a plugin should be, it includes a bunch of fairly simple but needed requirements which aren't going to be needed outside the very limited set of circumstances. I need to update https://github.com/meerk40t/meerk40t-example-plugin at some point soon for 0.8.0 and this seems like a pretty solid example. Also I might need to slightly tweak the 0.8.0 protocol to let you create multiple plugins out of a single plugin.

jpirnay:
I will take the different feedbacks into consideration and incorporate them over the next weeks (somewhat busy at the moment).

Issue #873, opened 2022-03-03
[Bug]: Not responding on close


dorsi90:
### Summary Description

The application not responding on close when it tries to connect over usb.
If i click some button like home the application tries to connect to the k40 but if it's unplugged or for other reasons the connection failed then the app tries to connect in loop. At this point if i close the app then not responding maybe because waiting for the loop connection terminate your job but if before close the app i click stop button i can close the app as expected.
So a solution can be just terminate the usb connection on close event like call the code of stop button.
Note: clicking reset or release usb before close the app wont solve the problem, only stop button works

### MeerK40t Version

0.7.5001

### MeerK40t Type

Source (zip/tar file from Github)

### Your Operating System

Windows

tatarize:
Hm.  I seem to recall having this happen. The check for shutdown=True occurs during the sending process. So if you aren't connected or sending it doesn't detect that or die gracefully. The connect cycles are a different loop than the send cycles and the send cycles are the only ones that check for the shutdown check.

dorsi90:
Well, then clicking stop button before close abort the send cycle and there isn't the problem on close the app.
The solution can be just call the code of stop button on close event. I havent read the code but i can suppose the send loop is running from another thread and the application waiting for no threads running.
If you allow me i would also like to suggest of adding a dialog on close that if there are some process runnings notify the user that some process still running and if he's sure to close the app.

tatarize:
I believe this may be fixed in 0.8.0 but wasn't fixed in the 0.7.x version. The reconnect loop doesn't detect the attempt to shutdown the program whereas the regular send loops (different loops for different devices) does detect that. It also could be moshi related if that's the routine since the reconnect loops are handled in the devices themselves.

dorsi90:
I see that clicking stop also interrupt the reconnect loop. In this case calling automatically the stop routine on shutdown stops the reconnect loop and solves the issue related at least for moshi.

tatarize:
This should be fixed in #878.

Issue #862, opened 2022-02-27
[Moshi]: Wrong raster position


dorsi90:
### Summary Description

To replicate the problem:
In inkscape I made a rectangle with a red stroke (cut) and a black text, the latter converted into "Object to path" because otherwise on meerk40t the text is in the wrong position and results with larger font size. After importing the file on meerk40t the rectangle is correctly recognized as cut and the text as raster but running the job the raster is done in the wrong position. However, if I click on make raster image before doing the job, it works correctly.

I would like to point out that my machine has a moshiboard MS10105 v4.85 and despite some other small bugs I found  you have done a really good job.

In the second pictures below i show the test without clicking "make raster image" and after do that.

![Cattura di schermata (93)](https://user-images.githubusercontent.com/44925234/155901793-5869fd7e-b7a2-41ae-b01b-7ad82e585ab5.png)
![IMG_20220227_223313](https://user-images.githubusercontent.com/44925234/155901801-87b411cc-d5c0-46d8-80ac-617eee22edcc.jpg)

### MeerK40t Version

0.7.5000

### MeerK40t Type

Executable

### Your Operating System

Windows

Sophist-UK:
@dorsi90 Guiseppe: Thanks for such a clear description of the issue. Can you please open the Device, Config and Execute Windows (and position them so they don't overlap) and attach a further screenshot?

Thanks.

dorsi90:
@Sophist-UK thank you for responding so quickly.
In the first picture the execute is before clicking on start, update or simulate buttons and the second is after.

![Cattura di schermata (94)](https://user-images.githubusercontent.com/44925234/155905493-cfcb8fbc-bb19-48ab-89ba-5a800b0084ac.png)
![Cattura di schermata (95)](https://user-images.githubusercontent.com/44925234/155905503-212bbed7-82a5-4e7b-815f-965db71251c3.png)

Sophist-UK:
Thx. @tatarize  may be the only Moshi expert who can look at this.

tatarize:
I think it's an iteration of the problems with text objects and has nothing to do with Moshi. In Inkscape, convert text to paths before exported.

dorsi90:
Thanks @Sophist-UK.
I would also like to point out that by clicking "make raster image" and then after the raster image is created as an element I do not move it to the raster or image I simply start the job and it works perfectly but if I delete the element just created and I start the job again the laser head moves on the x-axis out of the working area and bumps against the rail.
[...]

PR #878, opened 2022-03-12
Moshi Debug Mode, check #862

tatarize:
Check the Moshi bug #862 with debug mode.
Correct Moshi #873 bug of failing to close on reconnect cycle. Confirmed that it gets stuck in _thread_data_send function.


tatarize:
Closing this for now, it still needs some work but it means getting my moshiboard out and safeguarding it from blowing up again.

PR #882, opened 2022-03-15
Extended item-manipulation (0.7)

jpirnay:
## Motivation
I have amended the look & feel as well as the behaviour of the item manipulation (i.e selected) rectangle to make it's behaviour more "standard". It has now manipulation handles and extended operations (i.e. rotation and skew).
I've tagged this pr as draft to allow discussion on the placement / behaviour of the newly introduced elements. All relevant standard manipulations like moving and resizing work as well as the rotation variants (rotate around center, rotate around opposing corner) and the skew variant work ~~is still work in progress (some hints why it behaves as it does would be appreciated ...)~~

## Details
Manipulation rectangle

### Manipulation
details

- You have 4 corner handles to allow simultaneous resizing in X- and Y-direction (marked red in the picture above)
- You have 4 side handles to allow sizing in either X- or Y-direction (marked blue)
- You have 1 center handle to allow moving  (marked green)
- You have two handles on the bottom / on the right that allow skewing in X- or Y (marked orange): if you click on one of the skew-handles, then you can skew the element(s) between -90° and +90° in X- or Y-direction respectively

image

### Rotation
Rotation
In addition there is a rotation indicator at every corner. if you click on the rotation symbol then you can rotate the element(s) around their center.
corner
There is a secondary rotation option if you click just inside the corner then a rotation will take place around the opposing corner.
corner_color
So: 

- Click on the green area: rotation around center
- Click on the blue area: rotation around opposing corner
- Click on the red area, the handle itself: sizing as described above.

### Options
image
If you don't need skew or rotate or get confused by the handles you may turn these on/off via some check-boxes in the statusbar. (I did not want to use more valuable screen estate)

NB: The code has not undergone extensive blacking/codacing or whatever... I wanted first to get some feedback on this.


tatarize:
Hm. This is some impressive work. I had this planned for deeper in the future. 

It might be a bit hard without the underlying understanding of the API, but, you may want to consider using the widget more directly. I can tell from the code that it's sort of God-Object using the existing code to create and handle the handles, as one single complex widget. While it might be harder, I'd be a lot less likely to replace some of these in the future if you replaced the handles with widget handles.

The idea would be if you selected an item you would actually add widgets to the selection widgets that would each perform their individual functions, and be highly encapsulated code. So the widget would draw itself, and it would have a set size and respond to interactions within the code of just that widget. So grabbing a handle would, in theory, tend to be much more simplified. You can see an example of this in the widgets in Camera where there are these 4 different colored handle widgets each of which is just a widget itself but can be grabbed and moved accordingly with their logic just being the internal logic.

Also, I'm planning on making 0.8.x primary fairly soon, even if it makes things a lot buggier in the short term since the code is getting fairly hard to maintain with other changes. Though widget code is not been changed very much except that 0.8.x has a bunch of tools, which are themselves a type of widget.

jpirnay:
I will have a look at it, initially I did not want to refactor the code, but "just" wanted to add some handles, it got quite a bit more since that...
I believe the functionality is feature-complete now after I figured out how to directly manipulate the transformation matrix to allow for fluent skewing.
@tatarize I am not sure how to proceed now: whether to amend the code to include it in 0.8RC1 or to continue in the legacy branch

tatarize:
Well, changes to legacy should probably be tapered down. If it works well enough then legacy is fine since it was started before the switch but if you're figuring you'll flush this out some more you could switch to just the 0.8.x branch. Either one shouldn't cause an issue since the widgets code isn't changed extensively between the different versions.

jpirnay:
@tatarize I will definitely rework the code according to your suggestions for 0.8 (there's a nice use case for this approach I have in mind). But I hate to see good food or code spoil ;-) I believe it would provide some benefit even with this more old-fashioned approach to the legacy branch.

tatarize:
```
MeerK40t crash log. Version: 0.7.5001 git on Windows:AMD64 - 4.1.1 msw (phoenix) wxWidgets 3.1.5
  File "C:\Users\Tat\PycharmProjects\meerk40t\meerk40t\gui\scene\scene.py", line 316, in on_mouse_move
    self.scene.event(event.GetPosition(), "move")
  File "C:\Users\Tat\PycharmProjects\meerk40t\meerk40t\gui\scene\scene.py", line 857, in event
    response = current_widget.event(window_pos, space_pos, event_type)
  File "C:\Users\Tat\PycharmProjects\meerk40t\meerk40t\gui\scene\scenewidgets.py", line 505, in event
    self.tool(space_pos, dx, dy, 0)
  File "C:\Users\Tat\PycharmProjects\meerk40t\meerk40t\gui\scene\scenewidgets.py", line 650, in tool_translate
    obj.node.modified()
AttributeError: 'NoneType' object has no attribute 'node'
```

Tested this for legacy7 including. Crashed nearly instantly.

jpirnay:
Strange, will look at it tonight

jpirnay:
@tatarize 
Should be fine now, I hope ;-)

tatarize:
Since 0.7.x is a legacy and a bit feature frozen at the moment the code needs to be much closer to perfect to get included.  Whether that's worth it depends on bit on a few factors. It's already going in 0.8.x as a better code base, and good bones. Though also more complex, some bits of complexity shouldn't be ported over. Since simple and just works is strongly preferred for this more stable branch.

* Drawn widgets and selection boxes remain even while the tool is being used. I'd expect these to go away.
* Skew is weird.
* The checkboxes in the status bar are too scrunched.
* Handle doesn't make sense as a task.
* Rotation is sluggish. It's not rotating that corner to the current mouse cursor but going through all the different steps.
* Move appears to be sluggish too. My computer is a decade old so it might be more obvious here but some of the stuff like redrawing things and recalculating the bounding box (perhaps from scratch or something) is going on. Comparing this to legacy7 code is pretty apparent. The translate code is mostly the same so it seems likely that it's to do with redrawing the widget itself.  I told it to not draw while the tool was moving and it went faster but still seemed a little slower than typical legacy7.

I tried a couple bits of caching ( https://github.com/meerk40t/meerk40t/tree/cache-elements-with-images ) and it seemed to go a bit faster to not draw things while it's moving, it wasn't great for the speed. It might be the rotation corners. They seem a bit hefty of code. When really they should be done with a GraphicsPath ( https://docs.wxpython.org/wx.GraphicsPath.html ) object draw the path and store it then just send it to the screen with gc.StrokePath()  ( https://docs.wxpython.org/wx.GraphicsContext.html ) -- In fact, it might speed it up pretty considerably if you did that for most of the things you're drawing if they are complex and not changing.

tatarize:
I tried the speedup in that branch (which is quite buggy and *very* rough) a proper implementation would modify the widget/scene routines to do correct invalidation of particular parts of the scene and just redraw things in a more limited fashion. There's some graphics systems like in Android that do that to great effect, and it still ran pretty slow on some test svgs that are extremely slow. So I think the issue is that translation is actually doing translation, more directly than the matrix. Though I'm not sure, maybe I'm just looking for performance issues that are just my decade old computer being older.

The bulk of this code is fine. It should probably just require fixing the scrunched up checkboxes, the brokenish skew, and the angle by angle ticking up in the rotate. And maybe use a path for the arrow part. You can also use an image but the this tends to need to be converted to a py file (usually base64 version of the file in python text placed in icons), and drawn on screen. ( https://github.com/meerk40t/meerk40t/wiki/Tech:-Adding-Icons-to-a-MeerK40t-Module )

tatarize (changes requested):
* The Handle, Rotate, Skew in the status bar is too close together. In a typical launch I can't read Handle.
* Handle isn't Move, and seems confusing.
* Turning off Handle causes the corner arrows to become dotted lines. (you didn't set the pen style without handle selected).
* The handles blend into the background color of the scene. (we might could change the scene color, though it goes red during laser firing).
* Rotate moves too quickly. Moving the mouse in any direction causes it to spin quite rapidly.
* Touching Skew causes it to instantly turn into a flat/angled result.

tatarize:
I expected the rotate tool section to be akin to:
```python
        elif event == 0:
            self.draw_border = False
            if self.rotate_cx is None:
                self.rotate_cx = (self.right + self.left) / 2
            if self.rotate_cy is None:
                self.rotate_cy = (self.top + self.bottom) / 2
            current_angle = Point.angle(position[:2], (self.rotate_cx, self.rotate_cy))
            if self.last_angle is None:
                self.last_angle = current_angle
            delta_angle = current_angle - self.last_angle
            self.last_angle = current_angle
            self.rotated_angle += delta_angle
            for e in elements.flat(types=("elem",), emphasized=True):
                obj = e.object
                obj.transform.post_rotate(delta_angle, self.rotate_cx, self.rotate_cy)
                try:
                    obj.node._bounds_dirty = True
                except AttributeError:
                    pass
```

The angle function there in the svgelements.point class is: `return Angle.radians(atan2(p2[1] - p1[1], p2[0] - p1[0]))` which is basically just calling `math.atan2(y,x)` (yes, y goes first).

https://user-images.githubusercontent.com/3302478/160251109-c9e0c34f-ad59-4ec9-b55f-706be8d4bbee.mp4

tatarize:
https://github.com/meerk40t/meerk40t/commit/1a02f16a0e73599915af730558e7fad22f944f10
https://github.com/meerk40t/meerk40t/commit/914c051f4167df59bd28258e1f0c264e58375498

Will fix the math heavy parts of weird rotate and weird skew. The 0.8.x version with the statusbar is already more nicely spaced.

tatarize (approved):
The purple is decided rather ugly. But, I'm not a designer and I'm certainly not going to demand people make it prettier, though I have had design minded people come through and give a bunch of advice on making stuff a lot prettier. It decided works and is a very marked improvement and settled every objection I had.

PR #883, opened 2022-03-15
Fix resize of lines

Sophist-UK:
@Lexan reported a bug in #support on discord. I thought this would be an easy fix but it turned out more complicated than expected.

This PR does the following:

1. Adds support for mil units to SVGElements. I will submit this as a PR to the SVGelements repo in a few minutes. It also fixes a bug where "inchx" would be recognised as "inch" by tweaking the regex.
2. Simplifies repeated code in PositionPanel.
3. Fixes "resize" command so that it doesn't treat horiz/vertical lines as dots.

### DIFF
--- a/meerk40t/svgelements.py
+++ b/meerk40t/svgelements.py
@@ -43,7 +43,7 @@ Though not required the Image class acquires new functionality if provided with 
 and the Arc can do exact arc calculations if scipy is installed.
 """
 
-SVGELEMENTS_VERSION = "1.6.10"
+SVGELEMENTS_VERSION = "1.6.11"
 
 MIN_DEPTH = 5
 ERROR = 1e-12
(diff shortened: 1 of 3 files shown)


Sophist-UK:
Good to go

meerk40t/svgelements.py
```diff
@@ -157,12 +157,23 @@
 
 SVG_VALUE_NON_SCALING_STROKE = "non-scaling-stroke"
 
-PATTERN_WS = r"[\s\t\n]*"
+PATTERN_WS = r"[\s\t\r\n]"
```
tatarize, line 160: \r is defined within the CSS xbnf. as a white space character.

meerk40t/svgelements.py
```diff
@@ -190,22 +201,56 @@
...
+    "cm": "cm",
+    "cms": "cm",
+    "mm": "mm",
+    "mms": "mm",
+    "in": "in",
+    "ins": "in",
+    "inch": "in",
+    "inches": "in",
+    "mil": "mil",
+    "mils": "mil",
```
tatarize, line 219: Mil is not a unit within the CSS length routine. If you wish additional parsing be done here it should be done within the meerk40t project and not the svgelements project.

meerk40t/svgelements.py
```diff
@@ -190,22 +201,56 @@
...
+    "grads": "grad",
+    "gradian": "grad",
+    "gradians": "grad",
+    "rad": "rad",
+    "rads": "rad",
+    "radian": "rad",
+    "radians": "rad",
+    "turn": "turn",
+    "turns": "turn",
+}
```
tatarize, line 234: Many of these are not valid CSS length or angle units.

meerk40t/svgelements.py
```diff
@@ -190,22 +201,56 @@
...
 
 REGEX_IRI = re.compile(r"url\(#?(.*)\)")
 REGEX_DATA_URL = re.compile(r"^data:([^,]*),(.*)")
-REGEX_FLOAT = re.compile(PATTERN_FLOAT)
+REGEX_FLOAT = re.compile(r"(%s)" % PATTERN_FLOAT)
```
tatarize, line 238: REGEX_FLOAT is not purely about units it's just a regex of the floats.

tatarize (changes requested):
Nearly anything that asks for changes to svgelements will be rejected. This stuff should be parsed locally if such parsing is needed in a more local fashion. See the `is_length_valid()` routine for examples of this.  If these are not valid CSS Length components and parsing they won't used.

meerk40t/svgelements.py
Sophist-UK: Not sure what this comment means - the purpose of this change was to capture the group.

meerk40t/svgelements.py
Sophist-UK: Funnily enough I was thinking about what units of measure are defined in SVG when I was showering this morning. I was going to check.

So yes - you are right that most of this SVGelement change is incorrect, but there are few things that need to be retained - disallowing e.g. "turns" rather than "turn" for example, because "turns" is invalid SVG.

meerk40t/svgelements.py
Sophist-UK: As an aside, MK uses "turns" instead of "turn" and "mils" instead of "mil". I will address these.

Sophist-UK:
Summary - yes, I went down a rabbit hole and lost sight of the need to be precise about supporting the SVG spec and no more in SVGelements. I will rework both this and the SVGelements PR.

PR #885, opened 2022-03-16
RC-1 Version Prep

tatarize:


tatarize:
If it's the thing that is causing the X fails. Though I gotta get that finished.

Sophist-UK:
No - its use of SyntaxError exception - this has to be changed to a bespoke ConsoleSyntaxError exception as otherwise a Python syntax error in an import gets hidden.

tatarize:
Meh. Isn't going to make the initial version completely fail. And sounds more like Codecy being a pain.

PR #887, opened 2022-03-16
Attempt to fix windows builds

Sophist-UK:
I suspect that the failure was due to a "windows-latest" switch from win server 2019 to 2022 or possibly due to using powershell instead of bash. Fixing to windows-2019 and defaulting to bash (even on Windows).

Also some minor renaming to improve file naming consistency and prepare for future x64 builds.

If this works it should be merged back into `legacy7` branch.


Sophist-UK:
I guess this will not run until A) it is merged and B) we do another RC release.

tiger12506:
This doesn't fix the problem, and also makes it worse. (No testing I see...)
I can fix it, please wait patiently.

Sophist-UK:
Not sure how I could test it without a release. But I didn't think that the changes I made should make it worse either. ☹️

tiger12506:
> Not sure how I could test it without a release. But I didn't think that the changes I made should make it worse either. ☹️

Well the original issue was that it was trying to "move" a file that doesn't exist. Forcing windows to use bash shell where "move" is not a command wasn't helpful. Not a big deal in the grand scheme of things.

You can test it --- in your own fork of meerk40t, which is what I am doing currently to fix the issue. No worries.

Sophist-UK:
Of course. It's mv in bash. But using nah consistently in workflow is probably a decent idea, as you can use same or similar workflows across win, Linux and Mac.

tiger12506:
> Of course. It's mv in bash. But using bash consistently in workflow is probably a decent idea, as you can use same or similar workflows across win, Linux and Mac.

Sure, if nothing else is different between the three platforms. You should test that out in your own fork to see. Pretty sure you were the one advocating splitting up the yaml into different platforms just a little while ago so that we could run them individually. Now that I've done so, you want to try to increase their consistency with each other so you can combine them again? Exploration is good, but there is a place for that...

tiger12506:
Fixed the windows build issues in #894 

The rest of this, the renaming and the bash I will ok if it is made consistent across all platforms and is tested in someone's fork, might as well make a fresh pull request if you wish to do that.

PR #890, opened 2022-03-16
Fix errormessage when using abbreviated options

jpirnay:
This fixing the same issue in 0.8 that we already covered in PR#877 for the legacy7 branch


jpirnay:
Duplication - sorry

PR #892, opened 2022-03-16
Use bespoke exception for CommandSyntaxErrors

Sophist-UK:
Use of the Python syntax error results in syntax errors in panes being ignored and main window not opening!!!

### DIFF
--- a/meerk40t/camera/plugin.py
+++ b/meerk40t/camera/plugin.py
@@ -1,3 +1,5 @@
+from meerk40t.kernel import CommandSyntaxError
+
 def plugin(kernel, lifecycle=None):
     if lifecycle == "plugins":
         from .gui import gui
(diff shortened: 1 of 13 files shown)


tatarize (approved):

Sophist-UK:
The issue with the new CommandSyntaxError was certainly something I should have caught, however IMO the correct solution was not to make it a subclass of SyntaxError - which remixes the handling of Python and MK console syntax errors when the purpose of this PR was to separate them - but instead to change the code that picks up the exceptions (in kernel/kernel.py and gui/wxmmain.py) which use e.msg to get at the string (which appears to be undocumented as can I find no reference to it) rather than str(e) which is the documented ,method.

I will raise a PR to change this. Also I note that CommandSyntaxError is used in ruida/device.py when it is not referring to a Console command - this should probably be changed to a e.g. RuidaCommandError instead.

Also, I see that CommandSyntaxError is also used in the load routine to recognise a malformed file during load. I will need to track this down, but it doesn't seem like this should be a CommandSyntaxError either.

PR #900, opened 2022-03-17
Bypass ctrl, control confusion

jpirnay:
Keybind gets a "ctrl+letter" delivered while the majority of the predefined binds expect a "control+letter". This code translates these two into each other (well only ctrl to control, not vice versa).
Can be disregarded if theres consensus about the new terminology - I found it very annoying myself ...

### DIFF
--- a/meerk40t/core/bindalias.py
+++ b/meerk40t/core/bindalias.py
@@ -288,9 +288,25 @@ class Bind(Service):
         if not len(self.keymap):
             self.default_keymap()
 
+    # help transition from old definitions of control-key-combinations
+    def is_found(self, keyvalue, target):
+        valu = False
+        if not keyvalue is None:
+            s = keyvalue
+            if s in target:
+                valu = True
+            else:
+                s = keyvalue.replace("ctrl", "control")
+                if s in target:
+                    keyvalue = s
+                    valu = True
+        return valu, keyvalue
+
     def trigger(self, keyvalue):
-        if keyvalue in self.keymap:
-            if keyvalue not in self.triggered:
+        fnd, keyvalue = self.is_found(keyvalue, self.keymap)
+        if fnd:
+            fnd, keyvalue = self.is_found(keyvalue, self.triggered)
+            if not fnd:
                 self.triggered[keyvalue] = 1
                 action = self.keymap[keyvalue]
                 cmds = (action,) if action[0] in "+-" else action.split(";")
@@ -301,8 +317,10 @@ class Bind(Service):
 
     def untrigger(self, keyvalue):
         keymap = self.keymap
-        if keyvalue in keymap:
-            if keyvalue in self.triggered:
+        fnd, keyvalue = self.is_found(keyvalue, self.keymap)
+        if fnd:
+            fnd, keyvalue = self.is_found(keyvalue, self.triggered)
+            if fnd:
                 del self.triggered[keyvalue]
             action = keymap[keyvalue]
             if action.startswith("+"):


Sophist-UK:
wx uses Ctrl - so I changed control+ to ctrl+ for compatibility.

Did my 0.,7 PR make it into 0.8 or do I need to revisit it?

Sophist-UK:
I need to review both my keybind PR and the raster optimization PR and check that they made it into the 0.8 code correctly and fix it up if it didn't.

tatarize (approved):

tatarize:
> I need to review both my keybind PR and the raster optimization PR and check that they made it into the 0.8 code correctly and fix it up if it didn't.

The keybind stuff should have mostly ended up in the code. The optimize raster did not for a couple reasons. It muddied too many different files and it had some highly problematic code bits that violated the coupling between device type and information that couldn't be ported. The time taken for a raster differs from device to device and cannot tap directly into the lhystudios speed-code since that's device specific. It needs to calculate the raster time for a default device. And some things like moshiboard might need to home between rasters which changes the calculus pretty drastically. Also, this code tends to just get fixes and might be preferred after 0.8.x gets more fully corrected, especially if there's some slapdash methods in the original. They were fine since 0.7.x was going towards legacy anyway but the changes would need to be more consistent overall with the loose coupling between devices in 0.8.x

Sophist-UK:
I am sorry that I didn't get to review this PR before it was merged as I fear that it is off the mark.

The text "control+" is not used anywhere in the current code. Any previous keybindings do not carry across from 0.7 to 0.8 anyway, so they will not contain "control+" either as far as I can tell. @tatarize rightly removed the code I put into 0.7 to deal with the conversion which was needed there. As a consequence, I think that this PR #900 can be reverted as it doesn't do anything.

However in my original 0.7 code, I included functionality which checked whether keybinds or aliases had changed in the new version and which updated these automatically providing that the user had not changed the keybindings from a previous default. My rationale for this was that it was better to update any bindings still at default settings to the new defaults than to leave the user with out of date default keybindings. (Obviously, user defined keybindings would not be changed.) This code is no longer in the 0.8 codebase - and at this stage I am assuming that this was a deliberate decision by @tatarize to omit it and so I am not going to attempt to reverse this without his say so (though I still feel this code should be included). I will point out, however, that if we do not include this code, then we don't need to store the older kinybinds/aliases in a tuple and we should revert to using simple strings.

jpirnay:
I have no objections to revert this PR, I can only state that quite a lot of non-working control+ got carried over. If the code in your former PR would replace e.g.  Control-c with ctrl+c, then this would work as well. A migration routine like yours specifically addressing individual keybinds or a general one updating all corresponding control+ keys,  or whatever, as long as we don't compromise user experience / expectations

Sophist-UK:
@jpirnay Jens - I apologise if I didn't explain myself clearly. The code I wrote to convert control+ to ctrl+ was **only needed** in 0.7.x because the settings in 0.7 contained keybinds with control+. As @tatarize explained, all the settings in 0.8 are fresh and will never have had a control+ in them - so the conversion is unnecessary in 0.8.

I am more concerned about what happens to existing default keybinds and aliases as we change the defaults to something else. Should users who have never changed keybinds and who may not even know about them be stuck with the old defaults, or should they be automatically updated if the user hasn't changed them? I wrote code to change them at startup because that is what I believe is best, @taterize dropped that code, and I presume he had his reasons for that. As the owner and leader of this project that is his prerogative, though I am happy to resurrect the update code if he changes his mind.

PR #901, opened 2022-03-17
2nd attempt at getting CommandSyntaxError correct

Sophist-UK:
1. Separate out Console and Python Syntax Errors.
2. Provide a msg property for backwards compatibility for any non-MK plugins that still use a msg property.
3. Replace CommandSyntaxError in ruida device with a ruida specific exception because this is not a Console Command Syntax error.
4. Switch CommandSyntaxError handling to use `str(e)` rather than e.msg.

Unit tests pass, so this is not repeating the problems with #892.

### DIFF
--- a/meerk40t/gui/wxmmain.py
+++ b/meerk40t/gui/wxmmain.py
@@ -7,7 +7,7 @@ import wx
 from PIL import Image
 from wx import aui
 
-from meerk40t.kernel import CommandSyntaxError
+from meerk40t.core.exceptions import BadFileError
 from meerk40t.kernel import lookup_listener, signal_listener
 
 from ..core.cutcode import CutCode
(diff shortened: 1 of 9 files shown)


meerk40t/gui/wxmmain.py
```diff
@@ -1921,10 +1921,14 @@ def load(self, pathname):
                     channel=self.context.channel("load"),
                     svg_ppi=self.context.elements.svg_ppi,
                 )
+            # Note: I cannot see how loading a file can run commands that raise CommandSyntaxError
+            # and this exception is not raised outside commands and
+            # is not raised in svg or dxf file loading code,
+            # however this code is left in just in case.
```
tatarize, line 1927: I believe this is a different bit of code intended to raise SyntaxError which should not be CommandSyntaxError but is actually using SyntaxError as a standard candle that implies malformed file. Since syntax is the thing being malformed. I'm unsure whether this is older code or the malformed parsing error is a subtype of syntax error. This is not called with the CommandSyntaxError and should not be a part of that.

tatarize: In order to correctly test this create a svg file and remove the closing brackets to cause the file to be malformed.

meerk40t/ruida/device.py
```diff
@@ -520,7 +525,7 @@ def write(self, data):
         for array in self.parse_commands(data):
             try:
                 self.process(array)
-            except CommandSyntaxError:
+            except RuidaCommandError:
```
tatarize, line 528: I believe I caught this somewhere and wanted the type to be SyntaxError for that error flow. It works fine in RuidaCommandError but must be caught in the "write" section of the code.

meerk40t/gui/wxmmain.py
Sophist-UK: I cannot see how a python syntax error could indicate a malformed data file either.

I will test what happens with a malformed file and see if there is something I need to do.

meerk40t/ruida/device.py
Sophist-UK: For the same reasons as changing console command syntax exceptions, we should not use the Python SyntaxError exception for this either. I am unclear what (if anything) you are wanting me to do here?

meerk40t/gui/wxmmain.py
tatarize: I think ParseError might be a type of syntax error. In any event that is *not* something for the error routine within commands.

tatarize (changes requested):

meerk40t/gui/wxmmain.py
Sophist-UK: Ok - I did as you suggested and made a malformed SVG file and got an xml.tree exception and I trapped that exception and got a decent error message.

All I need to do now is to generalise this to SVG, DXF etc. file loads (e.g. by excepting this error in svgio and raising a meerk40t bespoke BadFileError exception, and doing the same in other formats). I also need to clear the "File loading..." box before I display the error message.

tatarize (approved):
Looks properly corrected.

Sophist-UK:
> Looks properly corrected.

😄

PR #902, opened 2022-03-18
Alter keybind defaults for rect / circle

jpirnay:
The former rect / circle defaults have become rather useless due to the change of internal units. This is reestablishing the former behaviour.

### DIFF
--- a/meerk40t/core/bindalias.py
+++ b/meerk40t/core/bindalias.py
@@ -124,7 +124,7 @@ DEFAULT_KEYMAP = {
     ),
     "ctrl+a": ("element* select",),
     "ctrl+c": ("clipboard copy",),
-    "ctrl+e": ("circle 1in 1in 1in stroke red classify",),
+    "ctrl+e": ("circle 0.5in 0.5in 0.5in stroke red classify",),
     "ctrl+f": (
         "",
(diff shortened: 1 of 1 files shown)


meerk40t/core/bindalias.py
```diff
@@ -124,7 +124,7 @@
     ),
     "ctrl+a": ("element* select",),
     "ctrl+c": ("clipboard copy",),
-    "ctrl+e": ("circle 500 500 500",),
+    "ctrl+e": ("circle 0.5in 0.5in 0.5in stroke red classify",),
```
Sophist-UK, line 127: Please see the comments earlier in the file - you need to insert a new entry before the previous one so this should read:
```
    "ctrl+e": ("circle 0.5in 0.5in 0.5in stroke red classify","circle 500 500 500",),
```
The old versions are used to provide automated update if the user has not changed the definition themselves.

Sophist-UK (changes requested):

meerk40t/core/bindalias.py
jpirnay: Done

Sophist-UK (approved):
LGTM

PR #905, opened 2022-03-19
RC3-Property Primacy

tatarize:
Makes Properties select properties sheets based on the node type in the activated_instances, reuses the same window.

This is a bit larger than RC3 typical fixes. The idea is to allow highlighting of a node in the tree and display the properties for that node. This should occur regardless of the node type. This means all properties windows need to be merged into one single over-arching property window type. Selecting a node should permit editing of that node. All node types provide `property/__class__/<subvalue>` sheets which all take in a `node` value.


tatarize:
* needs to correctly show the ImageElement properties.
* needs to correctly show the TextElement properties.
* Error removing value on window close with op selected.

PR #908, opened 2022-03-20
Fix ansi console legacy7 branch

Sophist-UK:
I have tested this on Windows and Linux and it works. Cannot test on mac,


tatarize (approved):
Mac is only partially supported sometimes. Though it may be the case that the background colors shouldn't be set at all, since that should inherit from the underlying stuff.

Sophist-UK:
The reason we need to set background colours explicitly is because there is an ansi code to change it and to switch back to "normal".

PR #912, opened 2022-03-21
Smooth raster intergrations for ruidacontrol+

tatarize:
The idea is to fix the clickity clack on ruidacontrol rasters. We do this by refusing direct moves along certain axis, but only as smooth moves at an angle when we travel along the other axis.


tatarize:
This may actually work. Requires testing on physical machine.

PR #913, opened 2022-03-21
Extended item manipulation

jpirnay:
## Motivation
I have amended the look & feel as well as the behaviour of the item manipulation (i.e selected) rectangle to make it's behaviour more "standard". It has now manipulation handles and extended operations (i.e. rotation and skew).

## Details
Manipulation rectangle

### Manipulation
details

- You have 4 corner handles to allow simultaneous resizing in X- and Y-direction (marked red in the picture above)
- You have 4 side handles to allow sizing in either X- or Y-direction (marked blue)
- You have 1 center handle to allow moving  (marked green)
- You have two handles on the bottom / on the right that allow skewing in X- or Y (marked orange): if you click on one of the skew-handles, then you can skew the element(s) between -90° and +90° in X- or Y-direction respectively

image

You can immediately create a copy of your selected element if you press the alt key while starting the move:
copy

### Rotation
Rotation
In addition there is a rotation indicator at every corner. if you click on the rotation symbol then you can rotate the element(s) around their center.
corner
There is a secondary rotation option if you click just inside the corner then a rotation will take place around the opposing corner.
corner_color
So: 

- Click on the green area: rotation around center
- Click on the blue area: rotation around opposing corner
- Click on the red area, the handle itself: sizing as described above.

While the default rotation is around the center of an object you can define the rotation center (which is by default to the center position with any new selection):
image

You an drag this indicator around to allow for custom rotations which would otherwise very difficult to achieve:
rotcenter

### Reference Object
In addition you do have an indicator whether this element works as Reference-Object 
imageimage

reference1

With such a reference object established you can move and scale elements relative to it via the context menu (right mouse-click):
image

image

 reference2

Usecase: Use a rectangle, circle or whatever shape as template for multiple burn tasks.

### Options
If you don't need skew or rotate or get confused by the handles you may turn these on/off via some check-boxes in the statusbar. (I did not want to use more valuable screen estate)

image

This is especially useful if you want to show and manipulate the rotation center: turn off the move handle to see the rotation center indicator:
image


tatarize (approved):
Wow, that's a lot of really things.

tatarize:
Skew handle seems a bit picky. Like turns it flat really quick or whatever.

jpirnay:
> Skew handle seems a bit picky. Like turns it flat really quick or whatever.

That's due to the algorithm that increases the angle and not the tan by a fixed value. It's more helpful at small angles but not as good above 70deg and more. Will see whether a different interpretation makes sense, but first some improvements around reference object and regmarks

PR #919, opened 2022-03-23
Fix mil conversion

jpirnay:


jpirnay:
Has become fruitless after the complete revision in PR918

PR #928, opened 2022-03-27
RotCenter fix, fix skew/improve rotation behaviour

jpirnay:
- Apply same logic as in 0.7 branch for skew and rotation
- Make rotcenter behaviour consistent


tatarize (approved):
Works great.

PR #935, opened 2022-03-28
Remove scipy from win build for legacy7

tiger12506:


tatarize (changes requested):
Also get rid of pyserial

tatarize (approved):

PR #939, opened 2022-03-29
Fix startangle in shape

jpirnay:
Wasn't recognized as parameter names and variable names did not match (has this always been the case?)


tiger12506 (approved):
Tested. It works now.

PR #942, opened 2022-03-30
0.7.6

tatarize:
* Version bump to 0.7.6
* Ruidacontrol upgrades
     * Raster without clickity-clack
     * Better support for angle rasters
     * Correct too fast speed vector issues
     * Vector between running without jogs fixed
* Console ScrolledWindow
* Correct Delete functionality for delete key
* Correct right click keymap crash.
* Control Handles for objects in scene @jpirnay
* Increase time for popup tool tips
* Corrections to dwell dots functionality

### DIFF
--- a/meerk40t/core/cutcode.py
+++ b/meerk40t/core/cutcode.py
@@ -617,6 +617,9 @@ class LineCut(CutObject):
         )
         settings.raster_step = 0
 
+    def __repr__(self):
+        return f'LineCut({repr(self.start())}, {repr(self.end())}, settings="{self.settings}", passes={self.passes})'
+
     def generator(self):
         # pylint: disable=unsubscriptable-object
         start = self.start()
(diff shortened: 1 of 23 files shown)


meerk40t/core/cutcode.py
```diff
@@ -931,17 +934,17 @@ def check_if_rasterable(self):
...
             if 0 < self.max_dx <= 15:
                 self.vertical_raster = True
                 self.settings.constant_move_y = True
             if 0 < self.max_dy <= 15:
                 self.horizontal_raster = True
                 self.settings.constant_move_x = True
+            if self.vertical_raster or self.horizontal_raster:
+                self.settings.raster_step = min(self.max_dx, self.max_dy)
+                self.settings.raster_alt = True
             return True
```
tatarize, line 948: Might want to return False if not vertical or horizontal raster

PR #950, opened 2022-04-01
Magnets: Fix for GRBL + extensions

jpirnay:
- Fix for inverted X- and Y-axis (e.g. for GRBL)
- Extended functionality within context-menu:
grafik


tatarize (approved):
half-lines can overlap with text

* Half lines can overlap with text.
* While this fixes GRBL it does not fix balor. (Balor's 0,0 point is the center of the bed).

PR #951, opened 2022-04-02
Lock selection to laser movement

jpirnay:
### Lock the selection to the laser position:
If you right click on one of the alignment buttons then the laser will move to the indicated position of the selected elements (regular click behaviour so far), but from then on the selection will remain at its relative position to the laser head, so if you move the laser then the selection will follow the movement. This behaviour ends when you either right click the same button again (laser head remains at its current position) or do a normal click of an alignment button (laser head will move).
Locked button
The current lock state will be indicated by highlighting the button (visually not extremely appealing but a clear indicator).
locked_laser

NB there might be some limits to this when it comes to manual physical movement of the laserhead.
### Laserpath appearance
Additionally the laserpath has become less obtrusive and therefore easier to distinguish from regular elements.

### DIFF
--- a/meerk40t/gui/navigationpanels.py
+++ b/meerk40t/gui/navigationpanels.py
@@ -376,34 +376,24 @@ class Drag(wx.Panel):
 
     @lockmode.setter
     def lockmode(self, value):
-        if value == 1:
-            self.on_button_align_corner_tl(event=None)
-        elif value == 2:
-            self.on_button_align_corner_tr(event=None)
-        elif value == 3:
-            self.on_button_align_corner_bl(event=None)
-        elif value == 4:
-            self.on_button_align_corner_br(event=None)
-        elif value == 5:
-            self.on_button_align_center(event=None)
-
-        self._current_lockmode = value
-
-    def lock_mode_toggle(self, button):
         coldefault = self.bg_color
         collocked = wx.GREEN
         bval = [coldefault, coldefault, coldefault, coldefault, coldefault]
-        if button == self.lockmode:
-            bval[button - 1] = coldefault
-            self.lockmode = 0
-        else:
-            bval[button - 1] = collocked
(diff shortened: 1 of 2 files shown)


meerk40t/gui/navigationpanels.py
```diff
@@ -487,6 +527,59 @@ def on_button_align_trace_quick(
...
 
+    def pane_show(self, *args):
+        self.context.listen("driver;position", self.on_update)
+        self.context.listen("emulator;position", self.on_update)
+
+    # Not sure whether this is the right thing to do, if it's still locked and then
+    # the pane gets hidden?! Let's call it a feature for now...
+    def pane_hide(self, *args):
+        self.context.unlisten("driver;position", self.on_update)
+        self.context.unlisten("emulator;position", self.on_update)
```
tatarize, line 538: Emulator position refers to the `lhyemulator` and I gotta imagine if you don't turn this off before starting a job it will go pretty crazy if the driver position is still locked. The job will be set because it'll already be through the planner but that would look pretty interesting to have it updating a lot based on the laser head position.

meerk40t/gui/navigationpanels.py
```diff
@@ -487,6 +527,59 @@ def on_button_align_trace_quick(
         self.context("trace_quick\n")
         self.drag_ready(True)
 
+    def pane_show(self, *args):
+        self.context.listen("driver;position", self.on_update)
+        self.context.listen("emulator;position", self.on_update)
+
+    # Not sure whether this is the right thing to do, if it's still locked and then
+    # the pane gets hidden?! Let's call it a feature for now...
```
tatarize, line 535: Pane hide is also called at the close of the program to unhook those things. Mostly as a sanity check.

tatarize (approved):
This seems like it has some fantastic potential.

PR #954, opened 2022-04-03
Kbd-Modifiers for rotation to allow 5/15deg steps

jpirnay:


meerk40t/gui/scene/scenewidgets.py
```diff
@@ -1026,12 +1027,37 @@ def tool_rotate(self, position, dx, dy, event=0):
 
             # Improved code by tatarize to establish rotation angle
             current_angle = Point.angle(position[:2], (self.rotate_cx, self.rotate_cy))
+            # boost the angle to allow for simpler bigger movements
+            if self.key_shift_pressed:
+                current_angle *= 2
+            elif self.key_control_pressed:
+                current_angle *= 4
```
tatarize, line 1034: The code is a lot better usage-wise without this angle-boost. The transitions from 1 deg angle to stepped angles doesn't spin weirdly as the previous current_angle is unceremoniously doubled or quadrupled, and it preserves the angle between the center of the object and the mouse is pretty much always constant.

tatarize (changes requested):

meerk40t/gui/scene/scenewidgets.py
```diff
@@ -1026,12 +1027,37 @@ def tool_rotate(self, position, dx, dy, event=0):
...
+                current_angle += 1 * math.tau
+
             if self.last_angle is None:
                 self.last_angle = current_angle
                 self.start_angle = current_angle
+
+            # Update Rotation angle...
+            if self.key_shift_pressed:
+                # Only steps of 5 deg
+                desired_step = math.pi / 36
```
tatarize, line 1048: Sticking with tau here `5 * (tau / 360)` is seemingly quite understandable.

meerk40t/gui/scene/scenewidgets.py
```diff
@@ -1,3 +1,4 @@
+from cmath import pi
```
tatarize, line 1: I don't really know the reason to import this here you import math just below this and use `math.pi` so this isn't used and likely should just use the regular math library.

tatarize (changes requested):

meerk40t/gui/scene/scenewidgets.py
jpirnay: This is the too smart  for its own good Visual Code Studio feature to look for undeclared stuff and add imports on its own...

meerk40t/gui/scene/scenewidgets.py
jpirnay: Done

PR #957, opened 2022-04-04
Better handling of small-sized selections

jpirnay:
Make sure handling handles remain visible and dont overlap for low zoom-factors or small-sized selections


jpirnay:
By default they are on the center:
Normal

If the selection size becomes too small / the zoom factor too big, the handles are pushed out:
Small

tatarize (approved):

PR #958, opened 2022-04-04
Improve display for small selection (0.7)

jpirnay:


tatarize (approved):
They might benefit from being partially transparent. But, certainly this is a bit better.

PR #965, opened 2022-04-07
Snap points

jpirnay:
Ready for testing. Introduce amongst a couple for fixes for position new functionalities to snap the cursor to either grid points or relevant points on elements.
Feedback appreciated on behaviour and look & feel.
image
magnet

magnet2

While you are moving your mouse when having selected a element-creation tool (i.e. intentionally not the 'Regular Scene' tool), you will see neighbouring points of interest being displayed and highlighted if they would be become a snapping target):
image
NB: the snapping behaviour will be ignored if you press the shift-key, in this case the current mouse-position will be considered.


tatarize:
2-lock points

You can sometimes almost show two different locations without clearly indicating which will end up being right.

tatarize:
If you're drawing a polygon and something else is selected the selection widgets takes over for the snap-to-elements and tool widget code. So it will highlight the corner but clicking the corner tries to move the corner rather than snap to it.

tatarize:
Drawing with a polygon tool can cause the mouse cursor to change if you highlight a selected corner of an element. Since you're drawing it will not revert back to normal.

tatarize:
Unknown if related but `Editing -> Element Size` does not resize smaller by pushing button.

tatarize:
Even without drawing, it's fairly easy to get the mouse cursor that got switched stuck in after mousing over anything that changes the mouse cursor.

tatarize:
Sometimes highlighting the an element snap you can see that the element position is there but it's not clear what element it is snapping to. It might be nice if it would maybe highlight the other snap elements in a more transparent way to know what it's snapping to or maybe just write on screen what the point type is or something.

tatarize (approved):

tatarize:
When the field of view for the zoomed in grid was applied I somewhat assumed it would also snap to that grid, in case I needed finer detailed points to snap to.

tatarize:
zoomed out snap

The snapview when zoomed out is also quite strange.

jpirnay:
@tatarize I will merge the PR to get on top of the multiple changes it already has and will provide some PRs to follow up on the things you've identified

PR #974, opened 2022-04-10
Force Legacy Classification Scheme

tatarize:
Turns legacy classification on by default. Overrides the current setting and turns it on the first run.


tatarize:
Due to problems during the raster-size elem-node switch over this was broken in 0.8.x and the older raster classifications were seemingly locked in. This will probably not be fixed in 7, but will be basically required in 8 unless I could somehow figure out all those edge condition and stuff and whether they are proper which they are probably weird.

PR #976, opened 2022-04-10
User defined color (0.7)

jpirnay:
Subset of [PR#977](https://github.com/meerk40t/meerk40t/pull/977) - limited to manipulation and grid color:

Syntax: you can specifiy the colour for an element by providing a color_code for it (hex representation in the format: 0xRRGGBBAA with RR the 2-digit hexvalue for the red component, GG for the green component, BB for the blue component and AA for the transparency (00 invisible to FF fully opaque))

```
set color_xxxx 0xRRGGBBAA
```
The following elements can be user-colored:
- color_manipulation - The manipulation rectangle plus modification elements
- color_grid - Grid-Lines

### DIFF
--- a/meerk40t/gui/scene/scenewidgets.py
+++ b/meerk40t/gui/scene/scenewidgets.py
@@ -52,7 +52,7 @@ def str_to_color(value):
     if len(value)<=8:
         value = value + "FF" # append opacity
     result = value[:2] + value[8:] + value[6:8] + value[4:6]+ value[2:4]
-    return eval(result)
+    return int(result, base=16)
 
 class ElementsWidget(Widget):
     """
--- a/meerk40t/gui/scene/scenewidgets.py
+++ b/meerk40t/gui/scene/scenewidgets.py
@@ -39,20 +39,11 @@ ORIENTATION_GRID = 0b00000100000000
 ORIENTATION_NO_BUFFER = 0b00001000000000
 BUFFER = 10.0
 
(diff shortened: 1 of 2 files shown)


tatarize (changes requested):
There should probably be a command to make things change instantly, rather than delaying until startup.

tatarize:
The issue is using set changes the value, but it doesn't change the derived values which are ultimately what causes the color.

0.7.x is mostly going to be abandoned going forward so hacking a solution in there is fine, but it might be best if it's easier to use and takes affect more directly.

jpirnay:
@tatarize 
It should be nearly immediate:
color1
The grid requires a redraw (i.e. a drag)
color2
Is there a signal to listen to a "set ..."? Then I could force such a redraw...

tatarize:
See #979.

Part of the issue here is that you're automatically resetting the pens and brushes during the draw cycle. While it might not matter with the given graphics interface it's usually a bad idea for most of these things. Generally the way the drawing cycles work it's almost always best to already have the pens in place and the right color and then update that color.

This is why I thought it wouldn't automatically update with set because I figured it would be done like that, but it wasn't. I wrote the PR there with 0.8.x adaptations in mind. Saving colors as valid colors in the settings makes the most sense even then you could do `set color_grid green` though you'd need to restart the program to get that to take effect if you didn't switch the data with the `scene color` command.

jpirnay:
image
Incorporated tatarizes improvements and extended the color-theming for handles

PR #985, opened 2022-04-13
Sandor-Balor Bug List

tatarize:
- [ ] Speed and power for the hatch operation doesn't work.
- [x] Incomplete circles, the last bit of the data sent to cut gets lost.
- [x] Settings for a polygon only kick in after it's completed the first segment. So it uses global settings then changes to the per-element settings.
- [ ] Disable and enabling flip at in opportune times apparently causes out of range errors. 
- [ ] Per operation override values are a bit sucky. I was worried about this since it's not clear when those values get set from default. I might need to give this a rethink.
- [ ] There may be issues with the travel moves not disabling the laser correctly for these moves. And it's firing the laser during what should be a travel.


sandorkonya:
> * [ ] Speed and power for the hatch operation doesn't work. ---> still
> * [ ]  There may be issues with the travel moves not disabling the laser correctly for these moves. And it's firing the laser during what should be a travel. --> still (engraving is independent on changing local travel speed setting
> * [ ]  Disable and enabling flip at in opportune times apparently causes out of range errors. ---> raises value error (y) or parameter overflow (x)
> * [ ]  Per operation override values are a bit sucky. I was worried about this since it's not clear when those values get set from default. I might need to give this a rethink
> * [ ] **NEW** Power for the cut operation doesn't work, but speed works

> * [x]  **SOLVED** -  Incomplete circles, the last bit of the data sent to cut gets lost.
> * [x]   **SOLVED** -  Settings for a polygon only kick in after it's completed the first segment. So it uses global settings then changes to the per-element settings.

sandorkonya:
> * [ ] Speed and power for the hatch operation doesn't work. ---> still
> * [ ]  There may be issues with the travel moves not disabling the laser correctly for these moves. And it's firing the laser during what should be a travel. --> still (engraving is independent on changing local travel speed setting
> * [ ]  Disable and enabling flip at in opportune times apparently causes out of range errors. ---> raises value error (y) or parameter overflow (x)
> * [ ]  Per operation override values are a bit sucky. I was worried about this since it's not clear when those values get set from default. I might need to give this a rethink
> * [ ] **NEW** Power for the cut operation doesn't work, but speed works

> * [x]  **SOLVED** -  Incomplete circles, the last bit of the data sent to cut gets lost.
> * [x]   **SOLVED** -  Settings for a polygon only kick in after it's completed the first segment. So it uses global settings then changes to the per-element settings.

PR #999, opened 2022-04-19
Make sure newly added elements appear in the tree

jpirnay:


tatarize (approved):

tatarize:
This code is mostly called from a signal so it should remove quick duplicates. So it won't get called too many times and should get called on item adds.

PR #1002, opened 2022-04-20
Raster Size Corrections

tatarize:
* Move towards using DPI rather than raster step as primary means of defining density within the program.
* When put through plot planning then the DPI will be converted to native units of a given step size in both x and y direction.


tatarize:
The make_image_for_op in cutplan likely needs some changing and probably the rasterwizard, since the main switch here is that DPI is pre-cutplan size based on actual values size whereas raster_x and raster_y are post cutplan information.

PR #1011, opened 2022-04-26
Additional tool: measure

jpirnay:
Additional tool to measure (a polygon): line lengths and contained area.
measure


tatarize (approved):
Looks solid. I can certainly see use cases for that.

PR #1016, opened 2022-04-28
Improve menu behaviour for reset/reify

jpirnay:


meerk40t/core/elements.py
```diff
@@ -4697,9 +4708,10 @@ def break_subpath_elem(node, **kwargs):
             self("element subpath\n")
 
         @self.tree_conditional(lambda node: not is_regmark(node))
+        @self.tree_conditional(lambda node: has_changes(node))
```
tatarize, line 4711: If you actually specifically have the something calling the conditional specifically with node and aren't inverting it. You don't really need the lambda for that. As `has_changes` would work perfectly well.

tatarize (approved):

PR #1017, opened 2022-04-29
Fix/Extend trace functionality

jpirnay:
All trace commands have been consolidated into a single trace command:
```
trace <method>
```
where method is being one of: hull, quick, complex, circle (and for completeness and backwards compatibility segment, although that is not real useful imho). If omitted method defaults to hull.
- hull is using the point representation of all elements and creating a complex hull around it
- quick is using the bounding box of the elements
- complex is very similar to hull but is dealing better with arcs (i.e. circles, ellipses, bezier shapes etc.)
- circle is using the smallest circle containing all the elements
- segments is just following the segments of all elements - a bit of a preview of the actual burn 
2022-04-29 16_14_57-MeerK40t v0 8 0006 RC5 git      m2nano

There is as well a command to create this hulls:
```
tracegen <method>
```
with the very same parameters as above. This will create the hull-shapes but will not move the laser.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -2,8 +2,10 @@ import functools
 import os.path
 import re
 from copy import copy
+from random import shuffle, randint
 from math import cos, gcd, pi, sin, tau, sqrt
 from os.path import realpath
+from numpy import linspace
 
 from meerk40t.core.exceptions import BadFileError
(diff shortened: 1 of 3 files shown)


tatarize (approved):
Seems like a nice change. Yeah the trace hull for circles is bugged. Generally because the points in the circle define certain arc characteristics and not something for which hull is valid.

Issue #1021, opened 2022-04-30
[Bug] SVG save of base64 non-conformant with RFC4648


Sophist-UK:
v0.7.6

RFC4648 Section 3.1 limits Base64 encoding to lines of no more than 76 characters.

Inkscape saved SVGs with Base64 encoding conform to this, but MK saved SVG files save the data as a single line.

tatarize:
This is accurate I posted a branch to try to conform to RFC2045 though the writing of newlines within the xml is a bit of a pain since it gets idiomized. I opened a SO question since I couldn't google any solid information on the problem and I have limited time just now.

https://stackoverflow.com/questions/72085514/write-a-literal-newline-in-elementtree-attribute

tatarize:
I'm not sure this is possible without writing my own xml writer, and I'm pretty sure it's not worth the implementation time. Though it does look better it has little impact on the current results.

Sophist-UK:
Agreed not worth the effort.

PR #1024, opened 2022-05-02
Conform to RFC2045

tatarize:
* Currently does not work. Returns idiomized version of carriage returns. See #1021.

### DIFF
--- a/meerk40t/core/svg_io.py
+++ b/meerk40t/core/svg_io.py
@@ -1,6 +1,6 @@
 import gzip
 import os
-from base64 import b64encode
+from base64 import encodebytes
 from io import BytesIO
 from xml.etree.ElementTree import Element, ElementTree, ParseError, SubElement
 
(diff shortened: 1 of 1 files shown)


Sophist-UK:
Have you checked whether the idiomising of newlines is done when you use `SubElement.set` or when it is written to file or somewhere in-between?

This is indeed pretty obscure - I have not found anything about this elsewhere.

Sophist-UK:
The code that does the idiomising is [here](https://github.com/python/cpython/blob/main/Lib/xml/etree/ElementTree.py#L1065).

There appears to have been extensive discussion about this [here](https://bugs.python.org/issue39011), [here](https://bugs.python.org/issue17582) and [here](https://bugs.python.org/issue5752), and it seems right that this should be escaped for a general case. 

However the code that calls this [here](https://github.com/python/cpython/blob/main/Lib/xml/etree/ElementTree.py#L903) suggests that using a QName might be able to avoid this (even though this might not be quite what QNames are for). QNames are documented [here](https://docs.python.org/3/library/xml.etree.elementtree.html#qname-objects).

Perhaps something like:
```
data = QName(f"data:image/png;base64,{png}")
 subelement.set("xlink:href", data)
```

tatarize:
Doesn't seem possible. Lateral move.

Sophist-UK:
The other alternative is to export the xml to a string and then insert the missing LFs. But you would have to search for specific strings indicating base64 and I am not sure that this is worth the effort either.

PR #1031, opened 2022-05-07
Circular gridlines

jpirnay:
Addition of a feature to show circular grid lines in addition to or as a replacement for the regular perpendicular grid:
circ_grid


tatarize (dismissed):
I'm not sure what the usecase for this is. It might be useful perhaps for balor where the center bit is radically important. And I did consider making a different grid type for rotary but that's mostly because the topology for a rotary is notably different.

jpirnay:
Will provide an extended version

PR #1033, opened 2022-05-09
Secondary + Circular grid

jpirnay:
## Command 'scene grid'
A new 'grid' command in the scene-context to allow a secondary rectangular grid with optional scaling and shifted origin (main use case rotary) as well as a circular grid (mainly for Galvo).
circ_grid1

circ_grid2

```
scene grid
```
displays the current parameters, e.g.
```
[20:04:09] scene grid
[20:04:09]     Grid-Parameters:
[20:04:09]     Primary: On
[20:04:09]     Secondary: On
[20:04:09]        scale-x: 3.00
[20:04:09]        scale-y: 1.50
[20:04:09]     Circular: On
```

```
scene grid primary
```
toggles the primary grid

```
scene grid secondary <cx> <cy> <scalex> <scaley>
```
toggles the secondary grid and sets the center to \<cx\> \<cy\> and the scale compared to the regular coordinates.
**NB: if you dont provide any parameters then the scale defaults to that in the rotary settings**
image

```
scene grid circular <cx> <cy>
```
toggles the circular grid and sets the center to \<cx\> \<cy\>

A simplified method of accessing the commands is via the contextmenu over the guide area:
image

Noteworthy:
- You may abbreviate primary, secondary and circular with p, s and c respectively
- If you omit the additional parameters for center and scaling, then the center defaults to that of the primary grid and the scale to 1.0

## Color scheme:
There are some more commands to set the colors of the elements:
### Primary grid:
scene color set guide #000000
scene color set grid #000000
### Secondary grid
scene color set grid2 #A0A0A080
scene color set guide2 #000000
### Circular grid
scene color set grid3 #A0A0A080
scene color set guide3 #A0A0A080

### DIFF
--- a/meerk40t/gui/scene/guicolors.py
+++ b/meerk40t/gui/scene/guicolors.py
@@ -27,9 +27,9 @@ class GuiColors:
             "measure_text": "#FF000060",
             "bed": "#FFFFFF",
             "grid2": "#A0A0A080",
-            "guide2": "#0000FF80",
+            "guide2": "#000000",
             "grid3": "#A0A0A080",
-            "guide3": "#00FF0020",
+            "guide3": "#A0A0A080",
         }
         for key in self.default_color:
             self.context.setting(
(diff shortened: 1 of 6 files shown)


tatarize (approved):
Maybe just `grid` rather than `showgrid`. I don't think anything else uses the more specific command stuff.

tatarize (approved):

PR #1036, opened 2022-05-12
Balor Attempted no 8021 commands

tatarize:
* 8021 turns the laser on but doesn't do so in a highly exact fashion.


joerlane:
Might be short command 0x0033 (IPG_OpemMO, IPGCloseMo) with data 0x0001; happens right before sending the 3099 length job packet. There's a corresponding 0x0033 with data 0x0000 right after the job completes.

PR #1043, opened 2022-05-16
Update planner.py

dpticar:
Position lost after "step repeat"


tatarize (approved):
Should fix the error of return to start after step-repeat

PR #1045, opened 2022-05-18
Extended Text-Entry + some fixes

jpirnay:
- Survive a corrupted configuration file (happened during a computer crash)
- Fix one wordlist issue (there are more...)
- Extended the Text-Tool (a simple textentry so far):

image


tatarize (approved):
Hm. Interesting. I mostly added the default {version} bit for examples but I could see a few things like time and date mattering especially if you're going to make a pre-programmed card or something for a particular material and giving the date would be sort of useful for that.

PR #1048, opened 2022-05-21
Wordlist extended

jpirnay:
A little bit rough around the edges (~~no translations yet~~, and some smaller flaws) but mostly working to do some tests.

### wordlist commands
```
[01:37:31]     --- wordlist Commands ---
[01:37:31]     add             add value to wordlist
[01:37:31]     addcounter      add numeric counter to wordlist
[01:37:31]     backup          Saves the current wordlist
[01:37:31]     get             get current value from wordlist
[01:37:31]     index           sets index in wordlist
[01:37:31]     list            list wordlist values
[01:37:31]     load            Attach a csv-file to the wordlist
[01:37:31]     restore         Loads a previously saved wordlist
[01:37:31]     set             set value to wordlist
```
### Switching between variable-view and regular view
image

### User defined formats for date and time
image
image
See [python documentation](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes) for available format codes

### GUI-Editor
image
#### Editing
You may delete or add a new wordlist pattern by using the Add text/AddCounter/Delete Buttons and you can edit individual (content) entries by slow-clicking on the relevant value:
image

#### Import of csv-files and using indexes and offsets
First of all you can import a CSV-file:
image
This will add corresponding wordlist entries with all their content of the file associated to them:
image

The index-value corresponds to the active dataset in the csv-file (highlighted by a red text entry in the right content-grid)
So the usage of {data} in the above example will use the active entry (in our case the "2")
You can add additional data from the next row, the row after that etc. by using the "#+num" modifier:
{data} provides you with the current dataset
{data#+1} with the next entry
{data#+2} with the one after that
and so on.
image
image

You can also use {data#-1} which will give you the previous entry, or if you omit the "+" or "-" in front of the shift-number, then this will be considered the entry from exactly this row: {data#4} will give you always the value from the 5th row regardless where the current index resides (notabene the index is deemed to be zero-based).

If you would change the index for instance to "5" (i.e. the 6th row in the csv-file) then all textboxes will update accordingly:
image
image

#### Saving and loading of a wordlist
You can backup and restore your wordlist:
image


jpirnay:
To do:
- remove pickle import to use mk-internal routines for backup/restore
- Fix bug: combo-box for index displays one option too many

tatarize (approved):

PR #1051, opened 2022-05-27
Increase raster merge margins...

Sophist-UK:
... so as not to split raster ops when areas are only just further apart than acceleration/deceleration gaps.

The extra margin needs to be applied in both directions so I decided to use an absolute of 1/20" rather than e.g. doubling dx/dy.

### DIFF
--- a/meerk40t/core/cutplan.py
+++ b/meerk40t/core/cutplan.py
@@ -476,7 +476,7 @@ class CutPlan:
 
             def adjust_bbox(bbox):
                 x1, y1, x2, y2 = bbox
-                return x1 - dx, y1 - dy, x2 + dx, y2 + dy
+                return x1 - dx - 50, y1 - dy - 50, x2 + dx + 50, y2 + dy + 50
 
             groups = group_overlapped_rasters(
                 [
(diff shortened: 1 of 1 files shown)


meerk40t/core/cutplan.py
```diff
@@ -476,7 +476,7 @@ def make_image_raster_op(self, op: LaserOperation):
 
             def adjust_bbox(bbox):
                 x1, y1, x2, y2 = bbox
-                return x1 - dx, y1 - dy, x2 + dx, y2 + dy
+                return x1 - dx - 50, y1 - dy - 50, x2 + dx + 50, y2 + dy + 50
```
tatarize, line 479: This change needs better justification. This is just hardcoding 50 rather than just adding 50 to dx and dy. Since those just call `get_raster_margin()` code it should be placed into that function with whatever justifications there are to that. Or added to dx and dy when those are declared in L475, perhaps with acceleration margins declared as a comment.

However, since this does do the thing you want and this code will not survive into 0.8.x due to too tight of coupling with the m2nano driver itself (50 is only 1/20th of an inch in M2 units) and that code *is* maintenance-mode only. 

Weak sorts of hardcoded static bug fixes like this are permitted, there's not really a need to preserve deep maintainability and slapdash fixes are fine for that.   I'll give it a day or two for any wouldbe objections and merge it.

tatarize (approved):

meerk40t/core/cutplan.py
Sophist-UK: I'll tweak it later today.

tatarize (approved):

PR #1053, opened 2022-05-27
Some fixes for wordlist

jpirnay:
a) Fix combobox display
b) remove pickle (uses json instead, as this is already used)

### DIFF
--- a/meerk40t/core/wordlist.py
+++ b/meerk40t/core/wordlist.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 import csv
+import json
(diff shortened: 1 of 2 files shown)


tatarize (approved):

tatarize:
This should probably set itself in the same working directory as the config files. Since there's sometimes a quirky thing with like OSX where you run it in a directory and you have no access to the working directory. Since you don't have access trying to save anything throws a permission error. This is why some files have a get_safe_directory() or whatever gets used for Settings there.

Issue #1020, opened 2022-04-30
[Bug] fill-rule:nonzero not respected


Sophist-UK:
V0.7.6

Overlapping text-to-path with explicit `fill-rule:nonzero` is rastered as `fill-rule:evenodd`

tatarize:
This is accurate. I'm not really sure why anybody would want the different fill rule but it should be pretty clear from loading that the non-zero isn't respected and all fill rules are evenodd.

tatarize:
I don't know any way to set the fill rule in wxPython, and it should be fairly obvious from the preview within MK that the rule will not be respected and from the simulation. This isn't really a bug, it's a defect since it's not an error it's just something that wasn't implemented since it seemed like it should be fairly rare to need such a thing. Though I'll apparently be hacking in end-caps and line-joins when I get free time next week.

Sophist-UK:
> I'm not really sure why anybody would want the different fill rule but it should be pretty clear from loading that the non-zero isn't respected and all fill rules are evenodd.

You want this because if you convert text to path and then merge the paths (to reduce the number of objects) then overlapping text has white spots.

tatarize:
Well, I get people might want it but I don't know anyway to make wxPython set a fill rule. If I can't draw it with wxPython then I can't render it with wxPython and fill rules are not able to be done. I could be looking in the wrong spot jpirnay did point out where gradients were hiding so there's certainly some things that aren't yet implemented but totally could be, but I don't know if or where fill rules could be set for a path to be drawn in wxPython.

Sophist-UK:
https://docs.wxpython.org/wx.GraphicsContext.html#wx.GraphicsContext.FillPath

`Fillpath` takes a `fillstyle` parameter wx.ODDEVEN_RULE or wx.WINDING_RULE.

In master:
* https://github.com/meerk40t/meerk40t/blob/main/meerk40t/gui/laserrender.py#L411=
* https://github.com/meerk40t/meerk40t/blob/main/meerk40t/gui/laserrender.py#L461=

In legacy7:
* https://github.com/meerk40t/meerk40t/blob/legacy7/meerk40t/gui/laserrender.py#L307=
* https://github.com/meerk40t/meerk40t/blob/legacy7/meerk40t/gui/laserrender.py#L332=

tatarize:
Yeah, that'll work. I'll hack it in where jpirnay put the end cap stuff. Should be about the same and needs a sort of per-path rendering or whatever.

[...]

PR #1054, opened 2022-05-27
svg fill-rule acknowledgement 

jpirnay:
Acknowledge fill-rule attribute:
image
Should fix [Issue 1020](https://github.com/meerk40t/meerk40t/issues/1020)

### DIFF
--- a/meerk40t/core/node/elem_line.py
+++ b/meerk40t/core/node/elem_line.py
@@ -61,6 +61,9 @@ class LineNode(Node):
             fill=copy(self.fill),
             stroke=copy(self.stroke),
             stroke_width=self.stroke_width,
+            linecap=self.linecap,
+            linejoin=self.linejoin,
+            fillrule=self.fillrule,
             **self.settings,
         )
 
--- a/meerk40t/core/node/elem_line.py
+++ b/meerk40t/core/node/elem_line.py
@@ -1,6 +1,6 @@
 from copy import copy
 
-from meerk40t.core.node.node import Node, Linecap, Linejoin
+from meerk40t.core.node.node import Node, Linecap, Linejoin, Fillrule
 from meerk40t.svgelements import Path
 
 
@@ -18,6 +18,7 @@ class LineNode(Node):
         stroke_width=None,
         linecap = None,
         linejoin = None,
+        fillrule= None,
         **kwargs,
     ):
         super(LineNode, self).__init__(type="elem line", **kwargs)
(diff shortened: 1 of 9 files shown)


meerk40t/core/node/elem_path.py
```diff
@@ -136,4 +141,5 @@ def as_path(self):
         self.path.stroke_width = self.stroke_width
         self.path.linecap = self.linecap
         self.path.linejoin = self.linejoin
+        self.shape.fillrule = self.fillrule
```
tatarize, line 144: This really should be `self.path` elem_path doesn't have a self.shape value.

meerk40t/core/node/elem_line.py
```diff
@@ -136,4 +141,5 @@ def as_path(self):
         self.shape.stroke_width = self.stroke_width
         self.shape.linecap = self.linecap
         self.shape.linejoin = self.linejoin
+        self.shape.fillrule = self.fillrule
```
tatarize, line 144: I guess this is fine, lines are never filled though, but for consistency I guess... but if that's the case, why not rect and circle? (those are non-crossing and can't need a fill rule either).

tatarize (changes requested):
Small bug in path (should crash), and some inconsistency with line but not ellipse/rect.

tatarize:
Fixes #1020

tatarize:
Also it's not really too essential there. The shape linecap and linejoin aren't actually part of `svgelements` so writing them doesn't matter.

In the case with stroke_width and the path it does matter since that code is still used for bounds checking here and there since it can calculate the bounds of a some path. And this can include or exclude the stroke width. This matters for cases when you're rendering them to raster them. So in cases when you want to know this information you set these back on the path so that the calculations work.

The linecap and linejoin and fillrule aren't respected by the `svgelements` since there's nothing about them that matters. In theory, if I get a free month or couple weeks and fully implement my `svgio` project these things would exist as `@property` values, and thus could matter though it would set clear boundaries between rendered paths and `<Path>` node objects, which `svgelements` does not do much to it's determent.

jpirnay:
I've added fillrule to those elementd where they are not actually needed per se. Not sure how the SVG-standard deals with those attributes for rects and ellipses... So this may need to be revisited

tatarize (approved):
The fill rules won't matter for them since they by definition simple. But, writing that stuff doesn't matter that much.

PR #1055, opened 2022-05-28
Fix crash when trying to save penboxes

jpirnay:
MK40 crashed if you never had any penboxes...


tatarize (approved):

tatarize:
Ironically since I fixed this in main already and my fix also fixed penbox saving I went with that for the conflict and this PR ended up doing no action.

PR #1059, opened 2022-05-29
Correct Hatch moves between non-contiguous blocks

tatarize:
* Hatch moves between non-contiguous blocks must be flagged differently to be sent as a movement rather than a cut.
* This code was fixed for the efill routine in vpype-embroidery (MIT license, I wrote it), but wasn't ported back here until now.


tatarize:
Code doesn't fix the problems in fact the efill code is older and includes extra spelling mistakes and a pointless calculation of line slope.

PR #1077, opened 2022-06-01
Explore numpath (numpy path) structure for data storage.

tatarize:
First, this might be too dramatic of a change to implement at this point. It may be better off to skip or ignore push things down the road, at least until 0.8.x is finished up. Or to implement it but then don't switch everything over very quickly or at all if it's not that needed. It may be excessive to recode a lot of the plotplanning, for example, to use the newer structure. Though the two-opt code is already adapted. But, for something like scanline_hatch algorithm, or the eulerian (if we ignored these things) we could do significantly better with a data structure that allows for gaps.

---

The idea behind numpy laser paths is to define a common structure that could replace the whole of cutcode and do so in a way that could be both be faster and more compact for other data structures commonly used throughout Meerk40t.

* Some drivers can use direct circular arcs, lines, and others use bezier curves. We should also avoid prematurely interpolating our data prior to knowing what types of data the driver might want or accept.
* Some drivers accept and use pure step versions of our code. Rather than using lines we would more directly use  individual orthogonal steps.
* Some drivers can use dwell points, in that, they can go to a particular location and turn the laser on for a period of time.
* Rasters are converted into a series of tiny steps that are usually either on or off but these are so plentiful that attempting to render them results in an unusably slow program.
* We should/must optimize our pathing such that we can minimize travel time.
* Some hatch operations result in many disjointed paths. The current objects we have to capture this data requires many node objects and is very inefficient.

For these reasons we have the numpath class which uses numpy arrays to define geometric and highly laser-specific objects, with enough generalization to capture hypothetical additional forms.

Each segment is defined by 5 complex values, these are:

`Start, Control1, Type/Power, Control2, End`

* Line: Start, NOP, Type=0, NOP, End
* Quad: Start, C0, Type=1, C0, End  -- Note C0 is duplicated and identical. But, only one needs to be read.
* Cubic: Start, C0, Type=2, C1, End
* Arc: Start, C0, Type=3, C0, End -- Note C0 is duplicated and identical.
* Dwell: Start, NOP, Type=4/Time, NOP, Start -- Note, Start and End are the same point.
* Wait: NOP,, NOP, Type=5/Time, NOP, NOP, -- Note, Start and End are the same point.
* Ramp: Start, PStart, Type=6, PEnd, End -- Power ramp line.
* Break: NOP, NOP, Type=99, NOP, NOP -- Structural segment.

Note: the Arc is circular only and defined by 3 points. Start, End, and a single control point. It  does not do elliptical arcs since they are weird/complex and no laser appears to use them.

Additional extensions can be added by expanding the type parameters. Since each set of values is a complex, the points provide x and y values (real, imag). The type parameter (index 2) provides the power, except for Dwell and Wait where it provides the time.

At first glance the structure may seem somewhat unusual but the `Position, Position, Type, Position, Position` structure  serves a utilitarian purpose. All paths are reversible. We can flip this data, to reverse these segments. The start and end points can swap as well as the control points for the cubic, and the power-start and power-end for Ramp. Whereas the types of parameter remains in the same position.

The numpy flip operation can be done over the x and y axis. This provides the list of segments in reverse order while also reversing all the segments themselves. This is the key to doing 2-opt travel minimization, which are provided in the code.

Ruida has some segments which start at one power and ramp up to a second amount of power. In this case, the reverse of this needs to attenuate power over that length. In such a case the C0 and C1 places are adapted to set the power.

This class is expected to be able to replace many uses of the Path class as well as all cutcode, permitting all affine transformations, and serve as a laser-first data structure. These should serve as faster laser-centric path-like code.


tatarize:
Drat. If I ever want to do proper closed paths I actually do need close operations. Since there's not really a way to easily do that without them.

tatarize:
Double drat. Proper Close events still need an initial Move to clearly indicate the structural parts of paths. But, doing that is asymmetric. You cannot sort the structural parts within the path, while maintaining the structure. And close paths occur at the end which creates a distinction between the beginning and the end.

tatarize:
Added in breaks that got renamed end because "break" is a reserved word. This appears to work well enough and corrected the display code to show numpaths fairly well.

While the goal here is still to stop, I'm going to go ahead and merge this code in, and simply not use it. You can create a numpath with `element* numpath` and see that there's elements of bounding box that still don't work quite right yet. But, the core of the code works and it's a good enough stopping point. The data structure might be useful for hatches to return heavily disjointed data without creating new nodes for everything.

PR #1082, opened 2022-06-04
Fix grid in simulation scene

jpirnay:


jpirnay:
Conflicts now with new bedwidget - redrawn

PR #1084, opened 2022-06-05
Optimise drawing of circular grid

jpirnay:
- Optimise drawing of circular grid at high zoom levels
- Establish max zoom level


tatarize (approved):
I love the use of tau in the comments

PR #1086, opened 2022-06-06
Identifying cameras

jpirnay:
Try to look for cameras to assign them as default for camera0 to camera4. Needs testing on Darwin.
camera

### DIFF
--- a/meerk40t/camera/gui/camerapanel.py
+++ b/meerk40t/camera/gui/camerapanel.py
@@ -26,24 +26,8 @@ _ = wx.GetTranslation
 
 CORNER_SIZE = 25
 
-available_cameras = []
-
-def get_cameras():
-    # checks the first 5 indexes, fill up with default after that.
-    index = 0
-    i = 5
-    while i > 0:
-        cap = cv2.VideoCapture(index)
-        if cap.read()[0]:
-            available_cameras.append(index)
-            cap.release()
-        index += 1
-        i -= 1
-    print("Get Cameras provided:", available_cameras)
-
 
 def register_panel_camera(window, context):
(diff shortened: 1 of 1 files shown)


meerk40t/camera/gui/camerapanel.py
```diff
@@ -1,5 +1,6 @@
 import platform
 
+import cv2
```
tatarize, line 3: This is actually somewhat problematic. OpenCV is/was not a requirement however, since services exists now this doesn't appear to be a problem anymore. There was a time historically that the gui for the camera was distinct from the background backend bit of the camera as it was intended to (though a bit wrongly) act like a plugin module. If OpenCV is not installed the Camera does not exist and everything just sort of acts normally.

I'll have to check.

Worked fine. Disabled the camera and lost the camera app exactly as it should. Though I do notice that the warning it throws is wrong.
```
OpenCV is not installed. Disabling Camera. Install with:
	pip install opencv-python headless
```
There's actually a hyphen between python and headless there rather than a space. But, that's not on you.

tatarize (approved):

meerk40t/camera/gui/camerapanel.py
jpirnay: I've encapsulated cv2 a bit (even if it's not necessary) and the associated UBS port for cam 0 to 4 are saved beyond a session - so it should be okay in principle to detect cameras only once.
To reset the association to the default just call "Identify cameras" again

tatarize: I checked it was fine.

tatarize (approved):

PR #1091, opened 2022-06-09
Extend event with nearest_snap

jpirnay:
If you have snap enabled then the creation of elements will more WYSIWYG instead of WYGIWIITGY ;-)
creation


tatarize (approved):
surprising number of files changed but understandable given the extra variable

jpirnay:
If you have snap enabled then the creation of elements will be more WYSIWYG

PR #1095, opened 2022-06-10
Some font tweaks

jpirnay:
- Improved fontface selection
- Load and save text-anchor and text x and y attributes, better recognition of "class" provided font-attributes
- Some tweaks around bound-box calculation for a text

There's a fundamental flaw in wxPython to get the right fontsize.
Both GetTextExtent as well as GetFullTextextent provide the fontmetric-size as result for the font-height and dont take the real glyphs into account. That means that ".", "a", "g" and "T" all have the same height...
Consequently the size is always off... This can be somewhat compensated by taking the descent from the font-metric into account.
A 'real' height routine would most probably need to draw the string on an empty canvas and find the first and last dots on a line. I was wondering whether the imagetools provide such a functionality?!

### DIFF
--- a/meerk40t/core/svg_io.py
+++ b/meerk40t/core/svg_io.py
@@ -21,6 +21,7 @@ from ..svgelements import (
     SVG_ATTR_STROKE_OPACITY,
     SVG_ATTR_STROKE_WIDTH,
     SVG_ATTR_TAG,
+    SVG_ATTR_TEXT_ANCHOR,
     SVG_ATTR_VERSION,
     SVG_ATTR_VIEWBOX,
     SVG_ATTR_WIDTH,
@@ -50,6 +51,7 @@ from ..svgelements import (
     Polyline,
     Rect,
     SimpleLine,
+    Length,
     SVGImage,
     SVGText,
 )
@@ -277,9 +279,13 @@ class SVGWriter:
                     ("font_weight", SVG_ATTR_FONT_WEIGHT),
                     ("font_style", "font-style"),  # Not implemented yet afaics
                     ("text_transform", "text-transform"),
+                    ("anchor", SVG_ATTR_TEXT_ANCHOR),
+                    ("x", SVG_ATTR_X),
+                    ("y", SVG_ATTR_Y),
                 ]
                 for attrib in attribs:
                     val = None
+                    # Look for both element and node
                     if hasattr(element, attrib[0]):
                         val = getattr(element, attrib[0])
                     if val is None and hasattr(c, attrib[0]):
@@ -492,6 +498,10 @@ class SVGProcessor:
             node.linejoin = nlj
 
     def parse(self, element, context_node, e_list):
+        # print ("Parse element: %s" % vars(element))
+        # for key in element.values:
+        #     entry = element.values.get(key)
+        #     print ("Key=%s, Entry=%s" % ( key, entry ))
         if element.values.get("visibility") == "hidden":
             context_node = self.regmark
             e_list = self.regmark_list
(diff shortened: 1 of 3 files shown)


tatarize (approved):

tatarize:
Image tools does not but it's a pretty shallow view of Pillow which almost certainly could. If you did `.bbox` on a pillow image of a drawn font there's a somewhat solid chance it would find the right coords for the height there. There's also some chance that a similar thing could be done more directly in Pillow. Even some font rendering might be done in Pillow.

Other alternatives, there's some chance of getting freetype working. Which is a font rendering engine. I believe it requires the freetype python addon and a copy of the free type dll file.

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-
# -----------------------------------------------------------------------------
#
#  FreeType high-level python API - Copyright 2011-2015 Nicolas P. Rougier
#  Distributed under the terms of the new BSD license.
#
# -----------------------------------------------------------------------------
#
# Direct translation of example 1 from the freetype tutorial:
# http://www.freetype.org/freetype2/docs/tutorial/step1.html
#
import math

import matplotlib.pyplot as plt
from freetype.raw import *
from PIL import Image

WIDTH, HEIGHT = 640, 480
image = Image.new('L', (WIDTH,HEIGHT))

def to_c_str(text):
    ''' Convert python strings to null terminated c strings. '''
    cStr = create_string_buffer(text.encode(encoding='UTF-8'))
    return cast(pointer(cStr), POINTER(c_char))

def draw_bitmap( bitmap, x, y):
    global image

    x_max = x + bitmap.width
    y_max = y + bitmap.rows
    p = 0
    for p,i in enumerate(range(x,x_max)):
        for q,j in enumerate(range(y,y_max)):
            if i < 0  or j < 0 or i >= WIDTH or j >= HEIGHT:
                continue;
            pixel = image.getpixel((i,j))
            pixel |= int(bitmap.buffer[q * bitmap.width + p]);
            image.putpixel((i,j), pixel)

def main():

    library = FT_Library()
    matrix  = FT_Matrix()
    face    = FT_Face()
    pen     = FT_Vector()
    filename= 'Vera.ttf'
    text    = 'Testings !'
    num_chars = len(text)
    angle   = ( 20.0 / 360 ) * 3.14159 * 2

    # initialize library, error handling omitted
    error = FT_Init_FreeType( byref(library) )

    # create face object, error handling omitted
    error = FT_New_Face( library, to_c_str(filename), 0, byref(face) )


    # set character size: 50pt at 100dpi, error handling omitted
    error = FT_Set_Char_Size( face, 50 * 64, 0, 100, 0 )
    slot = face.contents.glyph

    # set up matrix
    matrix.xx = (int)( math.cos( angle ) * 0x10000 )
    matrix.xy = (int)(-math.sin( angle ) * 0x10000 )
    matrix.yx = (int)( math.sin( angle ) * 0x10000 )
    matrix.yy = (int)( math.cos( angle ) * 0x10000 )

    # the pen position in 26.6 cartesian space coordinates; */
    # start at (300,200) relative to the upper left corner  */
    pen.x = 200 * 64;
    pen.y = ( HEIGHT - 300 ) * 64

    for n in range(num_chars):
        # set transformation
        FT_Set_Transform( face, byref(matrix), byref(pen) )

        # load glyph image into the slot (erase previous one)
        charcode = ord(text[n])
        index = FT_Get_Char_Index( face, charcode )
        FT_Load_Glyph( face, index, FT_LOAD_RENDER )

        # now, draw to our target surface (convert position)
        draw_bitmap( slot.contents.bitmap,
                     slot.contents.bitmap_left,
                     HEIGHT - slot.contents.bitmap_top )

        # increment pen position
        pen.x += slot.contents.advance.x
        pen.y += slot.contents.advance.y

    FT_Done_Face(face)
    FT_Done_FreeType(library)

    plt.imshow(image, origin='lower',
               interpolation='nearest', cmap=plt.cm.gray)
    plt.show()

if __name__ == '__main__':
    main()
```
image

The advantage here is that it actually drew the glyphs themselves which means we would have access to the actual curves of the glyphs. It's a bunch of dll stuff and other things I don't love but which seems to work. Though there might be different compatibility issues with other OSes.

---

The final option would be to maybe write my own parser. I'm told it's a lot of work but all the tables and how to parse the stuff is already online and it seems like the majority of this is the glyph tables and a few custom kerning tables for between different letters to slightly adjust the differences. The advantage here is that it would be universally compatible and could be easily adapted to Hershey fonts. Which are not TTL, but would need their own font parsing anyway. These are not fonts made out of pure glyphs but rather single line fonts where they only have the width of a single line. The importance here for laser cutting is that implementing the engraving of a hershey font (which is certainly different than normal TTF formats) is that it can engrave very small letters. I believe lightburn has this sort of capability as well as specifying directories directly.

---

A somewhat weirder extra possibility is that inkscape has console accessable font to path capabilities. If we wrote a tiny svg of the correct font size and such to a temporary directory we could ask inkscape if the person has it installed on the computer to convert that to a path and then deal with the font path. This is a lot less helpful but would in theory parse all the bits of the SVG font data correctly. Though it might heavily restrict editing or whatnot.

PR #1109, opened 2022-06-20
Tree stuff

jpirnay:
a) Try to make the tree context-menu somewhat more comprehensible - especially around the deletion from ops / from scene stuff
image
b) Recreate the state of expanded tree-branches after tree manupulation actions that  require a tree_rebuild

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -5213,12 +5213,28 @@ class Elemental(Service):
                 node.insert_sibling(n)
             node.remove_node()  # Removing group/file node.
 
+        @self.tree_conditional(lambda node: len(list(self.elems(emphasized=True))) > 0)
+        @self.tree_operation(_("Elements in scene..."), node_type=elem_nodes, help="", enable=False)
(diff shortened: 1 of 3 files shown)


meerk40t/core/elements.py
```diff
@@ -5406,68 +5422,145 @@ def clear_all_regmarks(node, **kwargs):
         # Calculate the amount of selected nodes in the tree:
         # If there are ops selected then they take precedence
         # and will only be counted
+
         @self.tree_conditional(
             lambda cond: len(
-                list(self.flat(selected=True, cascade=False, types=operate_nodes))
+                list(self.flat(selected=True, cascade=False, types=("reference")))
```
tatarize, line 5428: `("reference")` is still just `"reference"` which is notably different than `("reference",)` though the code there accepts a singe string without issue and does require a tuple for types.

tatarize (approved):
There was some discussion that removing references from elements should maybe be called "remove" and deleting the elements themselves should be called "delete". This might make things a bit more consistent.

PR #1110, opened 2022-06-20
Trafo-Panel: restrict laserhead movement

jpirnay:
a) small cosmetic fix for "Set current" for moshi / m2nano preferences
b) restrict laserhead movement in transformation panel to physical bedsize (default, can be turned off at your own risk)
constraint1
c) Fix for align laser with selection


tatarize (approved):
Probably should be on by default. And maybe not exist as an option to turn it, at least not an obvious option. Or if you want to run out of the correct range you should maybe do so with console commands directly.

PR #1127, opened 2022-06-28
Additional attributes in text-propertypanel

jpirnay:
image

### DIFF
--- a/meerk40t/gui/propertypanels/textproperty.py
+++ b/meerk40t/gui/propertypanels/textproperty.py
@@ -10,38 +10,38 @@ from ..mwindow import MWindow
 _ = wx.GetTranslation
 
 class PromptingComboBox(wx.ComboBox) :
-    def __init__(self, parent, choices=[], style=0, **par):
-        wx.ComboBox.__init__(self, parent, wx.ID_ANY, style=style|wx.CB_DROPDOWN, choices=choices, **par)
+    def __init__(self, parent, choices=[], style=0, **args):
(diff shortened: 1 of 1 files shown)


meerk40t/gui/propertypanels/textproperty.py
```diff
@@ -9,6 +9,40 @@
...
+        event.Skip()
+
+    def OnText(self, event):
+        currentText = event.GetString()
+        if self.ignoreEvtText:
+            self.ignoreEvtText = False
+            return
+        if self.deleteKey:
+            self.deleteKey = False
+            if self.preFound:
```
tatarize, line 33: This looks like it should crash but does not appear to crash. Should have been set in Init. But, this code doesn't seem to invoke. Since self.preFound (other than camel case issues) should hit an error since it's not set already.

tatarize (dismissed):
Attempting to use this, it didn't seem to update the text object in the scene. I switched my font to Dancing Font but that wasn't updated within the scene until I used the actual font dialog itself which did manage to update the font correctly.

meerk40t/gui/propertypanels/textproperty.py
jpirnay: Will look into it

jpirnay: Normalized variable names and predeclared a variable. In my environment stuff is update properly. A different font needs to be confirmed a) by pressing enter b) selecting one via the combobox c) leaving the field

tatarize (approved):

PR #1135, opened 2022-07-03
GUI amendments

jpirnay:
a) Follow tool icon when "tool" command issued
b) Reset tool to standard after ribbonPage change and new file load
c) Improve/Fix creation of reference icons in tree
d) ESC key stops tool if tool active / resets to 'tool none'  if inactive


tatarize (dismissed):

tatarize:
I merged in some changes that also affected buttons in wxribbon. I'll fix the conflict

tatarize:
Might be harder we touched a lot of the same code...

PR #1141, opened 2022-07-04
Two more geometry tools + individual enable-rules

jpirnay:
image

### DIFF
--- a/meerk40t/gui/wxmribbon.py
+++ b/meerk40t/gui/wxmribbon.py
@@ -565,8 +565,6 @@ class RibbonPanel(wx.Panel):
             button_bar.Bind(wx.EVT_RIGHT_UP, self.button_click_right)
 
         self.ensure_realize()
-        # Disable buttons by default
-        self.apply_enable_rules()
 
     def apply_enable_rules(self):
         for k in self.button_lookup:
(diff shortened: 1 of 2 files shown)


tatarize:
Heh, saw you click and try to select the circle with circle tool selected.

tatarize (dismissed):

meerk40t/gui/wxmribbon.py
```diff
@@ -470,6 +470,11 @@ def sort_priority(elem):
             b.identifier = button.get("identifier")
             b.action = button.get("action")
             b.action_right = button.get("right")
+            if "rule_enabled" in button:
+                b.enable_rule = button.get("rule_enabled")
+            else:
+                b.enable_rule = lambda cond: True
+
```
tatarize, line 477: Dictionaries `.get()` takes `key, default` so you can just do `b.enable_rule = button.get("rule_enabled", lambda cond: True)`

Though the preference for PEP8 etc is to avoid lambda if you can (for actual functions, this is just a quicky fake function and clearly great for that) but `None` is probably preferable and just check if `b.enable_rule` is not None. This is how it's working for "action" and "right".

Also if the rule_enable could vary between toggle/untoggle state or multi button (drop downs like camera) then it would need to save in the aspect and get a bit more complex and would need to check on state change.

tatarize (dismissed):

tatarize (approved):

PR #1159, opened 2022-07-11
Use and acknowledge lock status

jpirnay:
Consistently use and acknowledge the 'lock' status of an element.
a) new commands:
```
element lock
element unlock
```
b) Visual aknowledgement by 

1.  removing all corresponding modification actions from the context menus

image
vs
image

2.  displaying an appropriate icon in the element-tree
image

3. removing all modification elements from the scene-item-manipulation
image
vs
image

lock


jpirnay:
All now in 0.8b6

PR #1161, opened 2022-07-12
Some fixes / extending the simulation window

jpirnay:
Based on comments provided in discord plus some more:
a) Fix enable/disable of special ops: just fixing the error but not adding the output property or save it
b) Remove debug message from deregister listener in Kernel
c) Extend simulation window so that you can see the buildup of travel/time more clearly
simulate

d) Add scrollbars to property windows:
image


jpirnay:
All now in 0.8b6

PR #1170, opened 2022-07-16
Classification Logic, Lock-Flag, Minor tweaks

tatarize:
There are some changes incorporated:
a) Locking (ie preventing manipulations) of elements - via the context menu or the two newly established  commands "element lock", "element unlock" you can effectively protect an element from inadvertent deletion or manipulation
lock

b) Fixed / extended auto-classification:
Every operation type (engrave, cut, raster, hatch, dots - with the exception of image) is now color-aware - when classifying an element, every operation will check whether it's own color fits to one of the element stroke/fill property and will adopt the element if matched. The classification of this element  will continue even if a match was found, unless the stop-Flag of this op has been set. In This case it will stop classifying.
grafik

The color comparison can be made tolerant by checking and changing  the fuzzy color-logic flag in preferences.

grafik

c) Improved manual classifcation (extending tartarizes work):
If you right click on an element in the scene then you can assign the element to an operation - there are two options that if checked will not only assign the selected element but all elements with the same stroke or fill color
grafik

assign

d) added labels to all node types - will be honored when reading / saving files to svg (inkscape compatible)
grafik

e) Fixed the infamous "del" deletes all behaviour for good (I hope) - important: please verify in keybinds, that the del key is bound 'tree selected delete':
grafik

f) Quite a lot of visual improvements, fixing of visual inconsistencies (general + ribbonbar)

g) Still somehwat rough around the edges: icons-sizes within the GUI can be chosen: 

image


meerk40t/balormk/device.py
```diff
@@ -450,29 +450,36 @@ def __init__(self, kernel, path, *args, **kwargs):
...
+            "{danger}{defop}{enabled}{penvalue}{pass}{element_type}{direction}{speed}mm/s @{power} {frequency}kHz {colcode} {opstop}",
         )
         self.register(
             "format/op dots",
-            "{enabled}{pass}{element_type} {dwell_time}ms dwell {frequency}kHz",
+            "{danger}{defop}{enabled}{pass}{element_type} {dwell_time}ms dwell {frequency}kHz {colcode} {opstop}",
         )
         self.register("format/util console", "{enabled}{command}")
+        # Define maxspeed min, need to be adjusted for balor
+        kernel.register("dangerlevel/op cut", (50, 100))
```
tatarize, line 477: For balor, this isn't dangerous at all, at 5000mm/s without issue.

meerk40t/core/cutcode.py
```diff
@@ -402,43 +402,59 @@ def generate(self):
...
+            stop_at = len(cutcode)
+        if stop_at>len(cutcode):
+            stop_at = len(cutcode)
+        for i in range(0, stop_at):
             curr = cutcode[i]
             extra += curr.extra()
         return extra
 
-    def duration_cut(self):
+    def duration_cut(self, stop_at = -1):
```
tatarize, line 450: This sort of semaphore should probably be done with None. Then if stop_at is None, replace with len(cutcode).

meerk40t/core/elements.py
```diff
@@ -207,6 +285,10 @@ def __init__(self, kernel, index=None, *args, **kwargs):
 
         self.setting(bool, "classify_reverse", False)
         self.setting(bool, "legacy_classification", False)
+        self.setting(bool, "classify_fuzzy", False)
+        self.setting(float, "classify_fuzzydistance", 100.0)
+        self.setting(bool, "classify_autogenerate", True)
+        self.setting(bool, "classify_default", True)
```
tatarize, line 291: These are actually redundant. If the choice above is registered against a context it'll automatically set those values.

meerk40t/core/elements.py
```diff
@@ -1617,14 +1707,52 @@ def op_disable(command, channel, _, data=None, **kwrgs):
...
         # ==========
         # ELEMENT/OPERATION SUBCOMMANDS
         # ==========
+        @self.console_command(
+            "lock",
+            help=_("Lock element (protect from manipulation)"),
+            input_type=("elements"),
+            output_type=("elements"),
+        )
+        def e_lock(data=None, **kwargs):
```
tatarize, line 1732: This is probably a good idea and should be implemented reasonably. While regmark has it's place that place is mostly in actually providing marks for position or a jig or whatnot. There's some need to keep elements ready to burn but prevent them from being selected or messed with.

meerk40t/core/elements.py
```diff
@@ -5737,11 +5897,26 @@ def remove_type_op(node, **kwargs):
...
+        @self.tree_conditional(lambda cond: contains_no_locked_items())
 
+        @self.tree_conditional(
+            lambda cond: len(
+                list(self.flat(selected=True, cascade=False, types=("file", "group")))
+            )
+            == 1
+        )
+        @self.tree_operation(
+            _("Delete file '%s' and all its child-elements fully") % "{name}",
```
tatarize, line 5915: This may imply it deletes the file from disk. When it does not do that.

meerk40t/core/elements.py
```diff
@@ -6506,6 +6681,18 @@ def reset_user_changes(node, copies=1, **kwargs):
...
 
+        @self.tree_conditional(lambda node: node.lock)
+        @self.tree_separator_before()
+        @self.tree_operation(_("Unlock Element"), node_type=elem_nodes, help="")
+        def element_unlock_manipulations(node, **kwargs):
+            self("element unlock\n")
+
+        @self.tree_conditional(lambda node: not node.lock)
+        @self.tree_separator_before()
+        @self.tree_operation(_("Lock manipulations"), node_type=elem_nodes, help="")
```
tatarize, line 6692: Non-parallel wording. Unlock/Lock (Element/manipulations) -- One way or the other on the noun.

meerk40t/core/elements.py
```diff
@@ -7391,36 +7610,73 @@ def classify(self, elements, operations=None, add_op_function=None):
...
                     op = DotsOpNode(output=False)
-                elif hasattr(node, "stroke") and node.stroke is not None:
-                    if (
-                        node.stroke.red == 0xFF
-                        and node.stroke.blue == 0
-                        and node.stroke.green == 0
-                    ):
-                        op = CutOpNode(color=node.stroke, speed=5.0)
+                elif hasattr(node, "stroke") and node.stroke is not None and node.stroke.argb is not None:
+                    is_cut = Color.distance("red", node.stroke) <= 100
```
tatarize, line 7632: I thought this would use `self.classify_fuzzydistance` rather than hardcode the distance.

meerk40t/core/node/node.py
```diff
@@ -168,6 +168,8 @@ def create_label(self, text=None):
             if hasattr(self, key) and mymap[key]=="None":
                 if getattr(self, key) is None:
                     mymap[key] = "-"
+            if key.startswith("st"):
+                print ("Map: key[%s] = %s (%s)" %(key, mymap[key], type(mymap[key]).__name__))
```
tatarize, line 172: starts with st is a bit vague. stroke?

meerk40t/core/parameters.py
```diff
@@ -106,6 +109,14 @@ def validate(self):
         for v in COLOR_PARAMETERS:
             if v in settings:
                 settings[v] = Color(settings[v])
+        for v in LIST_PARAMETERS:
+            if v in settings:
+                if isinstance(settings[v], str):
+                    try:
+                        settings[v] = eval(settings[v])
```
tatarize, line 116: This tends to be a security issue. I can't validate that your loaded settings are from a local or safe source. Calling eval on something that may come from nefarious sources isn't great.

tatarize: Also, if it's a list you can get a string and usually use a character as a delimiter.

meerk40t/core/svg_io.py
```diff
@@ -401,7 +398,7 @@ def _write_operation(xml_tree, node):
         @param node:
         @return:
         """
-        subelement = SubElement(xml_tree, "operation")
+        subelement = SubElement(xml_tree, MEERK40T_XMLS_ID + ":operation")
```
tatarize, line 401: I'm not sure on the backwards compatibility here. While strictly speaking this better adheres to the various RFCs it might be somewhat breaking.

meerk40t/core/svg_io.py
```diff
@@ -517,9 +514,22 @@ def parse(self, element, context_node, e_list):
             context_node = self.regmark
             e_list = self.regmark_list
         ident = element.id
+        # Let's see whether we can get the label from an inkscape save
+        # But don't get too excited yet, it needs a place to sit -> WIP
+        my_label = ""
+        try:
+            my_label = element.values.get("{http://www.inkscape.org/namespaces/inkscape}label")
```
tatarize, line 521: This value would still be in the element.values dictionary even outside of loading code. Though it would be best to check the cascade of different values for this label (EG. we wouldn't be saving an inkscape label element but some saved elements might have a label).

meerk40t/core/svg_io.py
```diff
@@ -646,40 +670,53 @@ def parse(self, element, context_node, e_list):
...
+            # We need to reverse the meerk40t:... replacement that the routine had already applied:
             if tag is not None:
                 tag = tag.lower()
+                torepl = "{" + MEERK40T_NAMESPACE.lower() + "}"
+                if torepl in tag:
+                    oldval = element.values.get(tag)
+                    replacer = MEERK40T_XMLS_ID.lower() + ":"
+                    tag = tag.replace(torepl, replacer, )
+                    element.values[tag] = oldval
+                    element.values[SVG_ATTR_TAG] = tag
```
tatarize, line 682: It would probably be best to include the namespace and maybe define "note" and "operation" as constants at the top, and cover the older files that should still be *mostly* valid and wouldn't include that namespace.

meerk40t/core/svg_io.py
```diff
@@ -646,40 +670,53 @@ def parse(self, element, context_node, e_list):
...
+                        return
                 node_type = f"op {op_type.lower()}"
 
             if node_type is not None:
                 node_id = element.values.get("id")
                 # Check if SVGElement: operation
-                if tag == "operation":
+                op_s1 = MEERK40T_XMLS_ID + ":operation"
+                op_s2 = "operation"
+                if tag == op_s1.lower() or tag == op_s2.lower():
```
tatarize, line 702: op_s1 and op_s2 are defined by us and don't need the lower()
`if tag.endswith("operation")`? `if tag is in ("operation", MEER40T_XMLS_ID:operation)`...

meerk40t/gui/propertypanels/imageproperty.py
```diff
@@ -131,10 +131,10 @@ def set_widgets(self, node=None):
...
         try:
             bounds = node.bounds
-            self.text_x.SetValue(str(bounds[0]))
-            self.text_y.SetValue(str(bounds[1]))
-            self.text_width.SetValue(str((bounds[2] - bounds[0])))
-            self.text_height.SetValue(str((bounds[3] - bounds[1])))
+            self.text_x.SetValue("%.2fmm" % Length(amount=bounds[0], unitless=1).mm)
+            self.text_y.SetValue("%.2fmm" % Length(amount=bounds[1], unitless=1).mm)
+            self.text_width.SetValue("%.2fmm" % Length(amount=bounds[2]-bounds[0], unitless=1).mm)
+            self.text_height.SetValue("%.2fmm" % Length(amount=bounds[3]-bounds[1], unitless=1).mm)
```
tatarize, line 137: There's a `.length_mm` attribute that'll just show the length in right value. You can also set a limit to the number of decimals to consider significant in the Length() call there.

meerk40t/gui/scenewidgets/selectionwidget.py
```diff
@@ -1461,6 +1461,108 @@ def event(
         )
         return response
 
+class LockWidget(Widget):
```
tatarize, line 1464: This is some nice attention to detail.

meerk40t/gui/simulation.py
```diff
@@ -346,13 +386,39 @@ def on_plan_change(self, origin, plan_name, status):
...
             self.request_refresh()
 
-    def pane_show(self):
-        self.context.setting(str, "units_name", "mm")
+    def update_fields(self):
+        step = self.progress
+        travel = self.cutcode.length_travel(stop_at=step)
+        cuts = self.cutcode.length_cut(stop_at=step)
+        travel /= MILS_IN_MM
+        cuts /= MILS_IN_MM
```
tatarize, line 394: This code is probably not device sensitive. Since the native units for a balor device isn't going to be a mil.

meerk40t/core/cutcode.py
jpirnay: Done

meerk40t/core/elements.py
jpirnay: Amendend

meerk40t/core/elements.py
jpirnay: Amended

meerk40t/gui/propertypanels/imageproperty.py
jpirnay: Amended

meerk40t/core/elements.py
jpirnay: Amended

meerk40t/core/svg_io.py
jpirnay: I couldn't figure out how to reverse the xmls Namespace subject, any hints are welcome (e.g. for the inkscape-label subject as well as for others)

meerk40t/core/svg_io.py
jpirnay: The load routine, should have it covered

meerk40t/core/node/node.py
jpirnay: Was a leftover debug statement, removed

meerk40t/balormk/device.py
jpirnay: Will look into it

meerk40t/core/svg_io.py
jpirnay: Done

meerk40t/core/svg_io.py
tatarize: I never figured it out either, the "helpful" nature of that was just weirdly confusing and kind of annoying. And really the whole thing might be better off without respecting the namespace there since we're not going to run into other weird svgs with a "note" or "operation" node. Which is generally what a namespace is good for, making sure you don't clash with the same nodes for other variant files. I'm good just biting the bullet there, and calling things whatever I want without conforming to the whole namespace thing. SVG is supposed to just ignore `foreignObject` or treat them like they are whatever nodes. So long as we don't call something the same thing as a thing that's in real svg this would probably be fine.

Adhering to the the namespace stuff gains us nothing, but costs us writing backwards compatible files.

meerk40t/core/node/op_image.py
```diff
@@ -37,10 +40,52 @@ def __init__(self, *args, **kwargs):
         self.allowed_elements_dnd = ("elem image",)
         # Which elements do we consider for automatic classification?
         self.allowed_elements = ("elem image",)
+        self.stopop = True
+        self.allowed_attributes = []
 
     def __repr__(self):
         return "ImageOpNode()"
 
+    def __str__(self):
```
tatarize, line 49: The str stuff is added in this PR because it's not actually in main anymore. It defaults to the `node.__str__()` call which uses the default map and `self._formatter` (which gets set prior to checking labels) so the nodes str() and label remain the same. And will have the formatter overwritten on device switch.

meerk40t/gui/simulation.py
tatarize: The error here is, I think, mine. I converted everything over to native units during the cut planning process except speed. 

During cut planning, we scaled everything so that it would be in the native units of the device. This is an essential step since we are device agnostic prior to this and we sort of need things to be passed to the back-end in a 1:1 ratio. However, speeds are fairly universally given in `mm/s`, however that's a problem because native distances  are unrelated to physical units at that point. Speeds, consequently, violate this compact because they are still scaled given (partially) in physical rather than native units. -- What I actually need to do is scale the speed too so that you actively have access to speed in `native units / second` so that an M2Nano would get mil/second (or a different unit if you swapped your stepper drivers (for example)) and balor would get `galvos / second` which interesting is totally what the device accepts as real speeds.

meerk40t/core/node/op_image.py
jpirnay: Done

meerk40t/core/svg_io.py
jpirnay: I get that, still I think it makes sense to clearly tag mk stuff to not get into the way of any other svg writing programs.

meerk40t/core/elements.py
jpirnay: To be honest I wasn't aware of this, a code tinkerer like myself might look for clearer evidence of a new context variable being established. So as it is not hurting either I vote for maintaining this bit of code, but as in everything else: your project, your call

meerk40t/gui/choicepropertypanel.py
```diff
@@ -38,10 +48,69 @@ def __init__(self, *args, context: Context = None, choices=None, scrolling = Tru
...
+                                this_page = c["page"].lower()
+                            except KeyError:
+                                this_page = ""
+                            if len(negative)>0 and len(positive)>0:
+                                # Negative takes precedence:
+                                if not this_page in negative and this_page in positive:
+                                    self.choices.append(c)
+                            elif len(negative)>0:
+                                # only negative....
+                                if not this_page in negative:
```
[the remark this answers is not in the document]
tatarize: As minor pedant-stuff it'd get ignored and maybe fixed later on a review.

meerk40t/gui/icons.py
```diff
@@ -1,17 +1,32 @@
 # ----------------------------------------------------------------------
+from glob import glob
```
tatarize, line 2: What's with this import. Are you listing the directories with wildcards and I don't see it? I'll assume you wanted `global` and your IDE was too aggressively helpful.

meerk40t/gui/propertypanels/textproperty.py
```diff
@@ -196,6 +203,16 @@ def pane_hide(self):
     def set_widgets(self, node):
         if node is not None:
             self.node = node
+        try:
+            if node.id is not None:
+                self.text_id.SetValue(str(node.id))
+        except AttributeError:
```
tatarize, line 209: Doesn't seem like it should need AttributeError catch here. If for some reason node.id, or node.label is missing it would be better to crash since we have bigger problems. We put those into the base node class. They should keep existing.

meerk40t/gui/mkdebug.py
```diff
@@ -0,0 +1,88 @@
+import wx
```
tatarize, line 1: How does this class get invoked?

meerk40t/gui/position.py
```diff
@@ -200,6 +211,10 @@ def update_position(self, reset):
             self.position_y = y0 / conversion
             self.position_w = (x1 - x0) / conversion
             self.position_h = (y1 - y0) / conversion
+            self.org_x = self.position_x
```
tatarize, line 214: The org_(x,y,w,h) parameters aren't defined in init. They are always set since it calls the code that sets them from init. But, some of the nanny code-checkers whine a little bit without their set to none originally code.

meerk40t/gui/wxmribbon.py
```diff
@@ -103,14 +110,113 @@ def GetBestSize(self):
...
                     if self.GetLabel() != "":
                         ht += 23
-                    # print ("Found: wd=%s, ht=%s" % (wd, ht))
-                    size = wx.Size(wd, ht)
+                    else:
+                        ht += 10
                     break
+        else:
+            wd = size[0]
+            ht = size[1]
```
tatarize, line 118: If bar size[0]<50 and bar has no children `wd` and `ht` never get set and crashes.

meerk40t/gui/wxmribbon.py
```diff
@@ -103,14 +110,113 @@ def GetBestSize(self):
...
+         is termed a dummy as it sits in the ribbon bar doing nothing, while the expanded
+         panel holds the panel children.
+
+        :see: :meth:`~RibbonPanel.HideExpanded`, :meth:`~RibbonPanel.GetExpandedPanel`
+        """
+
+        if not self.IsMinimised():
+            return False
+
+        if self._expanded_dummy is not None or self._expanded_panel != None:
```
tatarize, line 168: Where do these come from and do they always exist? Codacy insists the property isn't set yet, and I can't find anything to contradict that.

meerk40t/kernel/kernel.py
```diff
@@ -1865,16 +1865,20 @@ def process_queue(self, *args) -> None:
...
                         listen_funct, listen_lso = listen
                         if (listen_funct == remove_funct or remove_funct is None) and (
                             listen_lso is remove_lso or remove_lso is None
                         ):
                             del listeners[i]
                             removed = True
                             break
-                    if not removed:
-                        print("Value error removing: %s  %s" % (str(listeners), signal))
+                    if not removed and ct>0:
```
tatarize, line 1878: This is just notifying of the error it's not the error itself. --- But, I haven't really located the problem. This is just saying that it tried to remove that twice and it did but that's an issue with the lifecycle not the notice here.

tatarize: I'll hunt down the real error later.

meerk40t/gui/wxmribbon.py
jpirnay: It's inheriting the agw class and uses those properties. Codacy isn't aware of that source I think

meerk40t/core/parameters.py
```diff
@@ -106,6 +109,23 @@ def validate(self):
...
             if v in settings:
                 settings[v] = Color(settings[v])
+        for v in LIST_PARAMETERS:
+            if v in settings:
+                if isinstance(settings[v], str):
+                    myarr = []
+                    sett = settings[v]
+                    if sett != "":
+                        # First of all is it in he old format where we used eval?
+                        if sett.startswith("["):
```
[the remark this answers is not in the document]
jpirnay: There might be more elegant ways to do it, for sure.

meerk40t/gui/icons.py
jpirnay: Argghh, I hate it when vscode tries to outsmart me, will fix it

meerk40t/gui/wxmribbon.py
jpirnay: Okay will fixit

PR #1176, opened 2022-07-19
SVGElements 1.7 - Correct Font Parsing and fix reDoS issue.

tatarize:
* Upgrades to SVGElements 1.7
* Use the corrected SVGElements Parsing.

### DIFF
--- a/meerk40t/core/svg_io.py
+++ b/meerk40t/core/svg_io.py
@@ -270,7 +270,6 @@ class SVGWriter:
                 for key, val in element.values.items():
                     if key in (
                         "font-family",
-                        "font-face",
                         "font-size",
                         "font-weight",
                         "anchor",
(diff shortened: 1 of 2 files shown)


tatarize:
@jpirnay I'm not sure what this will break. I fixed the Text.font parsing of objects and added in `.font_list` and properly extracted the `font_family` and fixed the ReDoS issue (https://github.com/meerk40t/meerk40t/security/code-scanning/4)

You seemed to have spackled over some of my errors in parsing the Text objects initially. So I think you'd have a better handle for what should and shouldn't work with the corrections.

The font_family, font_stretch, font_variant, and properties are correctly implemented and parse if the `font` attribute is set. These can either be set directly or can be set through a `font` attribute shorthard. That stuff should work correctly now. `font_weight` will now give a string value usually and `text.weight` would give the int value for that if it was needed. `font_face` will no longer be misused and is deprecated (which was true in the Spec), and was partially an error with regards to family and generic family etc. That stuff will now correctly parse, and `text.font_list` will give the hierarchy of preferred fonts, fully parsed out. Whereas `font_family` will give the valid `font-family` set value either directly set or set in the font attribute shorthand.

jpirnay:
Will look into it  (tomorrow)

PR #1182, opened 2022-07-21
Fix reDOS issue

jpirnay:


tatarize:
This is apparently not enough to fix the issue.

`/(?:[^\s,']|'(?:\\.|[^'\.])*')+/gm`

To match:
`'\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&\&`

Still results in catastrophic backtracking.

`/(?:[^\s,']|'(?:\\.|[^'\\.])*')+/gm`

Solves in 533 steps.

https://regex101.com/
With the given attack vector will tell you whether this matches and whether suffers reDoS.

tatarize:
I'm not sure if the sub example fits the given string very well.

Also, you can fall `list(text.font_list)` and it would give you the list of fonts within that font family. Apparently CSS allows it to be any number of fonts followed by fallbacks if that isn't in the system.

tatarize (changes requested):
Still suffers reDoS.

jpirnay:
I might be at loss here what to do. Any hints ( not an expert in regex ) ?

tatarize:
https://github.com/meerk40t/meerk40t/pull/1184

Check through this. Basically I had already written a parser for the font_list. The last thing is potentially found is generic family. I think that's right.

In theory `font_family="New Times Roman", "new century schoolhouse", serif` is a valid font family which isn't two but rather 3 each of which gets checked in order.

jpirnay:
Thx, will look at it

tatarize:
This regex is better removed than fixed.

PR #1183, opened 2022-07-21
Easy manual operation assignment

jpirnay:
Extension of the concept to manually assign elements to operations:
There is a new pane available: Operations (happy to take some naming-advice into account)

image
This pane will activate when elements are emphasized:

image 
image

You can click on one of the icons in this panel which are representations of the operations defined:

image

This will assign all emphasized objects in the scene to the corresponding operation.

There are some options to influence this behaviour:
image

- Similar: if activated will look for other elements in the scene (all, even those not emphasized) and identifies those with the same color-characteristics (stroke if you click on the icon with the left mouse button, fill if you use the right button) and assigns them as well to the target operation.
- Exclusive: if activated will remove all assignments of the affected elements from all other operations and will only keep those in the new target-operation

In addition you can synchronise the colors of the operation and the elements to be assigned:
- None : does not touch the colors
-  -> Op:  will use the color of first element (stroke if you click on the icon with the left mouse button, fill if you use the right button) and assigns it to the operation
- -> Elem: will take the color of the operation and assign it to the elements (stroke if you click on the icon with the left mouse button, fill if you use the right button) 

simulate


tatarize (approved):
Hm. I sort of wanted to steal the real-estate from the nice status bar for this but maybe if this is more popular that could be revisited. Could also go into the ribbon-bar in theory too. Looks pretty solid either way.

PR #1184, opened 2022-07-21
Eliminated reDoS progress through font-list

tatarize:
* Fix redos, got rid of the entire section there.
* Uses built in `.font_list` routine instead of rolling up it's own.
* Accounts for three or more potential fonts in the font-family.


tatarize:
This works for the thing I needed it for, and better references the other things. Corrects a couple errors in svgelements, so I'll need to update that but, I think this mostly is positive overall and corrects some harder to read stuff in fonts.py and gets rid of a wx4.0.7 incompatibility, which is still nominally supported.

PR #1191, opened 2022-07-24
Prefs reorg / simpler end of tool

jpirnay:
a) Fix missing icon for op dots in tree
b) Reorganise Preferences (and add snap-options to it, which were only accessible via Pane)
prefs

c) Allow quick tool reset: just press right-click on scene
rclick

d) Make opassign panel less annoying by remaining visible and become aware of additions / removal of operations

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -6193,6 +6193,7 @@ class Elemental(Service):
 
             node.remove_node()
             self.set_emphasis(None)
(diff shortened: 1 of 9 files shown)


jpirnay (commented):
Reorganise Preferences (and add snap-options to it, which were only accessible via Pane)
prefs

tatarize (approved):

PR #1192, opened 2022-07-24
Classification for new elements+ default colors

jpirnay:
a) Make classification of new elements user-configurable
b) Use default colors / fills as set in the status-colorbar for newly created elements


jpirnay:
As the default color for most new elements is no longer black (which classifies as raster) but blue (which classifies as engrave), some unit tests fail. Will amend them tomorrow

tatarize (approved):

PR #1197, opened 2022-07-27
Improve scene modifier keyboard controls

tatarize:
* Changes the scene modifier methodologies.
* Rather than hooks on the keyboard down and up that the individual widgets have to catch and track, all widgets are given modifiers for all events which are tracked by the original panel.

### DIFF
--- a/meerk40t/gui/scene/scenepanel.py
+++ b/meerk40t/gui/scene/scenepanel.py
@@ -118,7 +118,7 @@ class ScenePanel(wx.Panel):
             if self.isAltPressed:  # ignore multiple calls
                 self.isAltPressed = False
                 self.scene.event(self.scene.last_position, "kb_alt_release", None)
-        if key == 27: # ESC
+        if key == wx.WXK_ESCAPE:  # ESC
             # Is a tool active? If yes -> stop it, if no -> reset tool to none
             if self.scene.tool_active:
                 self.scene.event(self.scene.last_position, "lost", None)
--- a/meerk40t/gui/scene/scenepanel.py
+++ b/meerk40t/gui/scene/scenepanel.py
@@ -24,6 +24,9 @@ class ScenePanel(wx.Panel):
         self.scene_panel.Bind(wx.EVT_PAINT, self.on_paint)
         self.scene_panel.Bind(wx.EVT_ERASE_BACKGROUND, self.on_erase)
 
+        self.scene_panel.Bind(wx.EVT_KEY_DOWN, self.on_key_down)
+        self.scene_panel.Bind(wx.EVT_KEY_UP, self.on_key_up)
+
         self.scene_panel.Bind(wx.EVT_MOTION, self.on_mouse_move)
 
         self.scene_panel.Bind(wx.EVT_MOUSEWHEEL, self.on_mousewheel)
@@ -41,6 +44,8 @@ class ScenePanel(wx.Panel):
         self.scene_panel.Bind(wx.EVT_MOUSE_CAPTURE_LOST, self.on_mouse_capture_lost)
 
         self.scene_panel.Bind(wx.EVT_SIZE, self.on_size)
+
+        self.last_key = None
 
         try:
             self.scene_panel.Bind(wx.EVT_MAGNIFY, self.on_magnify_mouse)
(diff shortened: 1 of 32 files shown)


meerk40t/gui/scene/scenepanel.py
```diff
@@ -70,61 +72,30 @@ def signal(self, *args, **kwargs):
...
-                self.scene.event(self.scene.last_position, "kb_shift_release", None)
-        if "ctrl+" in literal:
-            if self.isCtrlPressed:  # ignore multiple calls
-                self.isCtrlPressed = False
-                self.scene.event(self.scene.last_position, "kb_ctrl_release", None)
-        if "alt+" in literal:
-            if self.isAltPressed:  # ignore multiple calls
-                self.isAltPressed = False
-                self.scene.event(self.scene.last_position, "kb_alt_release", None)
-        if key == 27: # ESC
```
tatarize, line 121: This functionality is restored in the tools themselves. See modifier key_up/escape for the tools

meerk40t/gui/toolwidgets/tooldraw.py
```diff
@@ -49,9 +49,14 @@ def event(
...
-        elif event_type == "lost":
+        elif event_type == "lost" or (event_type == "key_up" and modifiers == "escape"):
             self.series = None
-            return RESPONSE_DROP
+            if self.scene.tool_active:
+                self.scene.tool_active = False
+                self.scene.request_refresh()
+                response = RESPONSE_CONSUME
+            else:
+                response = RESPONSE_CHAIN
```
tatarize, line 59: All the tools check for lost or if escape is hit directly. If they tool already isn't active, then it chains to the tool container widget.

tatarize:
@jpirnay This code touches a lot of things you modified, especially the modifier keys. Most of the things I know existed still work (or were made to work again).

Alt + Rightclick = Special Zoom
Ctrl + Rightclick = Special Pan
Escape key, tools active = Cancel
Escape key, tool-non active = Default Tool

Some others got modified, I tried to preserve the logic rigorously.
* keep_old = Unknown
* select smallest = Unknown
* selection-widgets, different_event = Unknown

jpirnay:
Need to go through it from a source-review-perspective, just quickly looking at the functionalities, there seem to be some things no longer to work as before:
- alt+dragmove element = create + move copy missing
- select smallest on scene (including ctrl) - seem to be inverted?!
- rotate modifiers (to go in higher steps) missing  
-

tatarize:
I reinverted the select smallest, though I missed two copies of that setting in selectionwidget which still had some keyboard hooks and thus wouldn't have registered the keyboard so it would have done the same thing in both cases.

`smallest = bool(self.scene.context.select_smallest) != bool("ctrl" in modifiers)`

I'll check the rotation ticks...

tatarize:
* Corrected rotation ticks.
* Alt doesn't deselect so it makes copies that get stuck.

tatarize:
The issue with alt+relocate is true in `main` branch too. So that's not an effect of this code.

meerk40t/gui/scene/scenepanel.py
```diff
@@ -141,30 +112,24 @@ def on_mousewheel(self, event):
...
-        event occurs the up and down rotation is treated as left and right.
+        Triggers scene events for up, down, and left and right which exist on some mice and trackpads. If shift is held
+        down while the wheel event occurs the up and down rotation is treated as left and right.
 
         """
         if self.scene_panel.HasCapture():
             return
         rotation = event.GetWheelRotation()
+        modifiers = self.event_modifiers(event)
         if event.GetWheelAxis() == wx.MOUSE_WHEEL_VERTICAL and not event.ShiftDown():
```
jpirnay, line 123: Not sure while you are not using ```"shift" in modifiers``` - would seem more consistent to me. But maybe just a matter of style

jpirnay (approved):
Looks good to me

meerk40t/gui/scene/scenepanel.py
tatarize: Sort of. The axis there is more fundamental than the modifiers. If you hold shift your axis is inverted, that's not really part of the bag of modifiers. Also, I think event.ShiftDown() is just a check of a bool rather than `"shift" in modifiers` which I guess is a hash-table lookup and also constant time (actually a list, I guess it's n-time). But, here we're dealing with the panel itself and the fundamentals of how the panel work, and the modifiers for passing to the scene.

I had originally considered it, but I figured this was easier to read. And even if modifiers got changed or whatever, this isn't part of that code that governs how scenes interact but rather part of the scenepanel which is tasked with taking the interactions and handing them to the scene to process. `Modifiers` is part of the scene processing and really part of the hook into those bindings.

PR #1201, opened 2022-07-28
Codacity, black, isort

jpirnay:


tatarize (approved):
black + isort changes a bunch of stuff. Sometimes it's best to do it as a single thing without changing much of anything else.

PR #1204, opened 2022-07-29
Some work around the keybind/device panel

jpirnay:
a) Allow multi-deletion
b) Get back to start with abolute defaults if needed
image

c) Some optical plus functional extensions to device panel
devices


tatarize (approved):
The label for services should be somewhat unneeded. The path is required to be unique or it references the same data there. And -l as an option is likely easily confused with -I or -1 and somewhat like -O as an option or whatever. But, that's a pretty minor criticism.

PR #1208, opened 2022-08-02
Convert op keeps children

jpirnay:


tatarize (approved):
Covers all the bases. Since these were already children they should be registered already. Swapping the parent is fine when the parent is set there.

PR #1210, opened 2022-08-02
Avoid unnecessary info calculations in statusbar / suppress statusbar if desired

jpirnay:


tatarize (dismissed):
Yeah, tracks whether the calculation is needed. Though the variable could be `self._needs_generation`. Which is a bit more like making it private since it's only for internal use.

tatarize (approved):

PR #1220, opened 2022-08-03
Improve/harmonize textentry behaviour across the board

jpirnay:
Make textinput behaviour consistent across MK:
- changes during textentry are no longer immediately assigned to underlying properties per keystroke. Some of wrong input was caught, some not, but mostly you propagated unwanted changes during textentry (ie change bedwidth from "310mm" to "240mm" created confusing UI behaviour and quite a lot of unnecessary updates in-between before you had your desired value)
- changes are acknowledged after leaving the field (EVT_KILL_FOCUS) or after pressing the enter-key

- Plus visual feedback if invalid data has been entered in more fields

### DIFF
--- a/meerk40t/grbl/gui/tcpcontroller.py
+++ b/meerk40t/grbl/gui/tcpcontroller.py
@@ -13,8 +13,8 @@ class TCPController(MWindow):
         self.button_device_connect = wx.Button(self, wx.ID_ANY, _("Connection"))
         self.service = self.context.device
         self.text_status = wx.TextCtrl(self, wx.ID_ANY, "")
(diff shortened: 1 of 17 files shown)


tatarize (approved):
Looks pretty solid, Though non-standard I guess you could toss the listener in there. Since you know you'll be binding kill_focus and text_enter there each time. You should be able to bind that in the init, and if you passed that function to the constructor call it each time. There's probably some other benefits to normalizing the controls like this.

jpirnay:
I was tempted to, but it's so easy to go back to a standard wx.TextCtrl if needed. So I refrained from it

PR #1223, opened 2022-08-04
Statusbar info for burn progress / Burn history

jpirnay:
a) Provide infos about burn progress in statusbar
burnstatus

b) Display some history about finshed jobs in spoolerpanel

grafik

### DIFF
--- a/meerk40t/balormk/driver.py
+++ b/meerk40t/balormk/driver.py
@@ -111,6 +111,34 @@ class BalorDriver:
 
         @return:
         """
+        # preprocess queue to establish steps
+        dummy_planner = PlotPlanner(self.settings)
+        self.current_steps = 0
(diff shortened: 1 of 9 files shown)


tatarize (dismissed):

tatarize (approved):
It's not bad. I think there might be some better ways but there's certainly some worse ways too. It may need to be tested against large sets of rasters data since that could take up a lot of time. And it might be better in that case to just decide the pixels get a set amount of steps rather than push that on the dummy_planner. But, nothing is fiddling with the back end and the only thing that could go wrong are performance issues which would, at worst be 2x the time to process the count.

jpirnay:
Agree, 'there needs to be a better way' - but at least the frontend is there and the part in the backend can be improved surely

PR #1232, opened 2022-08-08
Make operation warning-level device dependent and user-configurable

jpirnay:
Allow individual settings per device
grafik

### DIFF
--- a/meerk40t/kernel/settings.py
+++ b/meerk40t/kernel/settings.py
@@ -75,12 +75,13 @@ class Settings:
         except PermissionError:
             return
 
+
     def read_persistent(
         self,
         t: type,
         section: str,
         key: str,
-        default: Union[str, int, float, bool] = None,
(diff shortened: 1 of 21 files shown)


meerk40t/kernel/settings.py
```diff
@@ -91,10 +92,37 @@ def read_persistent(
         @param default: default value if item does not exist.
         @return: value
         """
+        def listed(str_value):
+            dummy = str_value.split(";")
+            result = []
+            for item in dummy:
```
tatarize, line 98: Rather than `dummy` you could just do `for item in str_value.split(";")` and it would be more readable.

tatarize (approved):
The only maybe issue is if there's a list of things that can now save and previously was discarded that might persist and cause a behavior change. If such a thing is found we should probably just rename it to start with `_`.

PR #1244, opened 2022-08-09
More consistent time estimates in spooler

jpirnay:
a) Calculate in spooler the time of the cutplan execution (just like in simulator)
b) Update the current time / estimates in spooler panel

### DIFF
--- a/meerk40t/core/spoolers.py
+++ b/meerk40t/core/spoolers.py
@@ -3,7 +3,7 @@ from threading import Lock
 
 from meerk40t.core.units import Length
 from meerk40t.kernel import CommandSyntaxError
-
+from meerk40t.core.cutcode import CutCode
 
 def plugin(kernel, lifecycle):
     if lifecycle == "register":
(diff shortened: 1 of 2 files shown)


meerk40t/core/spoolers.py
```diff
@@ -329,6 +329,25 @@ def __init__(self, label, items, driver=None, priority=0, loops=1):
...
+
+        for item in self.items:
+            time_cuts = 0
+            time_travel = 0
+            time_extra = 0
+            if isinstance(item, CutCode):
+                travel = item.length_travel()
+                cuts = item.length_cut()
+                travel /= MILS_IN_MM
+                cuts /= MILS_IN_MM
```
tatarize, line 343: Eventually this needs to do better jobs based on the driver more. Since there's different native units between different values. This will work for M2Nano, Moshi, and GRBL (the grbl boards use MM and IN but really is written to use Mils since those other units are fractional)

tatarize (approved):

meerk40t/core/spoolers.py
jpirnay: Acknowledged, the overall estimate algorithms need to be improved

PR #1252, opened 2022-08-11
Show/Hide regmarks

jpirnay:


tatarize (approved):
That's really professionally done, way to combine different aspects very coherently.

PR #1253, opened 2022-08-13
Update mac11_pyinst.yml

joerlane:
Mac runner doesn't compile for wxPython4.2. Duh!


tiger12506 (approved):
Harmless. Does what it says it does.

joerlane:
Thanks. I love you all.

PR #1255, opened 2022-08-14
Render Text Cache

tatarize:
* Adds draw_text to use image-rasters as an intermediate step and avoid recreating the required text and correctly sizing the text.


tatarize:
Got this to somewhat work, it was sort of crap.

PR #1273, opened 2022-08-20
Lifecycle Delegation Loop

tatarize:
Modules typically have a sort of weird hamstring structure to provide delegates. This modifies that structure to permit anything within the module to yield `delegates()` these in turn can also yield their own `delegate()` values providing a tree of delegated objects more easily than being forced to call `add_module_delegate()` at the delegate level itself.

### DIFF
--- a/meerk40t/balormk/gui/balorconfig.py
+++ b/meerk40t/balormk/gui/balorconfig.py
@@ -4,6 +4,7 @@ from meerk40t.device.gui.warningpanel import WarningPanel
 from meerk40t.gui.choicepropertypanel import ChoicePropertyPanel
 from meerk40t.gui.icons import icons8_administrative_tools_50
 from meerk40t.gui.mwindow import MWindow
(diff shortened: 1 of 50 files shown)


tatarize:
@jpirnay if you "buggy balorconfig" is good you might want to port it out of here. I'm not going to merge this.

The pane_show/pane_hide is too fundamentally different than module_open/module_close. In theory in the AUI you can show and hide panels all the time and they should get and process hooks only while visible. Which this code isn't terrible and may actually work. I can't throw a wrench this large at the 11th hour. Not without redoing all the work and focusing on the fundamentals of what should happen while the pane is visible vs. what should happen when the wxmain window shuts down (which would actually be the same lifecycle as the kernel itself since it only shutdown at program end.

It's good, but I don't have the time to prove this for every window currently. And with a better understanding of the difference or maybe a revision of AUI's pane system such things could be brought into compliance later.

PR #1281, opened 2022-08-25
Crash Log + Small Fixes + Alignment Panel + Ruida/GRBL update

tatarize:
Checking the crash log for issues.

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -1790,13 +1790,13 @@ class Elemental(Service):
         )
         def makeop(
                 command,
-                remainder,
+                remainder=None,
                 **kwargs,
         ):
-            op = ConsoleOperation(command=remainder)
-
-            self.add_op(op)
-            return "ops", [op]
+            if remainder is not None:
+                op = ConsoleOperation(command=remainder)
+                self.add_op(op)
+                return "ops", [op]
 
         @self.console_argument("dpi", type=int, help=_("raster dpi"))
         @self.console_command("dpi", help=_("dpi <raster-dpi>"), input_type="ops")
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -372,6 +372,7 @@ class Elemental(Service):
         self._first_emphasized = None
         self._align_mode = "default"
         self._align_boundaries = None
+        self._align_group = False
         self._align_stack = []
 
     @property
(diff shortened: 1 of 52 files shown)


meerk40t/core/elements.py
```diff
@@ -470,10 +473,10 @@ def length(self, v):
         return float(Length(v))
 
     def length_x(self, v):
-        return float(Length(v, relative_length=self.device.width))
+        return float(Length(v, relative_length=self.device.width, unitless=1))
```
tatarize, line 476: What's the justification for this? I daresay there's a lot of cases when this function would absolutely expect that unitless values are pixels, since that's fairly consistent.

meerk40t/gui/wxmmain.py
```diff
@@ -1006,14 +1009,13 @@ def part_of_group():
...
+                "label": _("Left"),
                 "icon": icons8_align_left_50,
                 "tip": _(
                     "Align selected elements at the leftmost position (right click: of the bed)"
                 ),
-                "action": lambda v: kernel.elements("align left\n"),
-                "right": lambda v: kernel.elements("align bedleft\n"),
-                "size": buttonsize,
+                "action": lambda v: kernel.elements("alignmode push\nalignmode first\nalign left\nalignmode pop\n"),
+                "right": lambda v: kernel.elements("alignmode push\nalignmode bed\nalign left 1\nalignmode pop\n"),                "size": buttonsize,
```
tatarize, line 1018: An align mode language? I might need to check into this in more detail. Why isn't this something more like: `align bed left 1` or `align --bed left 1` or something like that? What about this requires a stack?

meerk40t/ruida/device.py
```diff
@@ -610,7 +611,7 @@ def process(self, array):
 
             self.x = self.abscoord(array[1:6])
             self.y = self.abscoord(array[6:11])
-            self.plotcut.plot_append(self.x / UNITS_PER_uM, self.y / UNITS_PER_uM, 0)
+            self.plotcut.plot_append(int(self.x / UNITS_PER_uM), int(self.y / UNITS_PER_uM), 0)
```
tatarize, line 614: This code is wrong. Since it feeds that data straight into the spooler it expects it to be in device relevant units and if in actual units it needs to still multiply by the relevant device matrix. I'll probably need to do some work to fix this still. This part of the review is a note to remember to do that.

meerk40t/core/elements.py
```diff
@@ -2546,160 +2730,222 @@ def subtype_align_elements(
...
 
+        @self.console_argument(
+            "asgroup",
+            type=int,
+            help=_("Treat selection as group (1) or as single elements (0)"),
+        )
         @self.console_command(
             "centerv",
-            help=_("align elements at center vertically"),
+            help=_("align elements at center verically"),
```
tatarize, line 2941: Why was this specifically changed to misspell vertically?

meerk40t/core/elements.py
tatarize: Confirmed this would require that all things that use these functions will have a unitless value of 1 native scene unit. That's rect, line, circle, etc.

meerk40t/core/node/op_cut.py
```diff
@@ -244,6 +244,12 @@ def time_estimate(self):
                 estimate = float("inf")
         if self.passes_custom and self.passes != 1:
             estimate *= max(self.passes, 1)
+
+        def isNaN(num):
```
tatarize, line 248: Should probably us `isnan` import from `math` rather than rolling your own.

PR #1287, opened 2022-08-30
Several bug fixes. Linting fixes. Black. OpInfo Panel.

tatarize:


meerk40t/gui/alignment.py
```diff
@@ -70,8 +71,9 @@ def __init__(self, *args, context=None, scene=None, **kwds):
         )
         self.rbox_align_y.SetSelection(0)
         self.rbox_align_y.SetToolTip(
-            _("Align object to the top, centered or to") + "\n" + \
-            _("the bottom in relation to the target point")
+            _("Align object to the top, centered or to")
+            + "\n"
+            + _("the bottom in relation to the target point")
```
tatarize, line 76: The translation on this tooltip might be weird. It requires two lines and the new line there is forced.

PR #1295, opened 2022-09-02
Imageactions

jpirnay:
Some new actions based on selected elements:
a) Render the selected elements and cut the image into multiple pieces to allow for separate burning (or whatever) of oversized stuff
imageprop

b) Create keyhole images by selecting elements plus an additional element that will act as mask or keyhole for the other elements

### DIFF
--- a/meerk40t/extra/imageactions.py
+++ b/meerk40t/extra/imageactions.py
@@ -1,0 +1,275 @@
+from copy import copy
+from math import isinf
+
+import PIL
+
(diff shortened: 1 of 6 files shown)


tatarize:
Hm. At least we can hem and haw over adding new features also "alignment" this is part of alignment tools?

tatarize (approved):
So much for hemming or hawing, this is perfectly self contained and thus pretty harmless across the board. The only bugs it could have are bugs within it.

PR #1302, opened 2022-09-04
Balor Raw Hard Mode

tatarize:
Adds in -h flag and STATE_RAW_MODE to Galvo. We will make absolutely no attempt to check the state or verify that things are sent correctly or the laser is ready or to execute the list. Lists will not be padded on an end of list call or execution. 

Raw hard mode sends only exactly what it's told to. Lists must be completely filled to be sent.


joerlane (approved):
This appears to draw 1:1 command inputs to the controller hardware. Absolutely essential for extreme testing. A+

PR #1309, opened 2022-09-08
Expand/Experiment laserpanel expansion

tatarize:
* Add in laserpanel raw-save function

### DIFF
--- a/meerk40t/gui/laserpanel.py
+++ b/meerk40t/gui/laserpanel.py
@@ -10,6 +10,7 @@ from meerk40t.gui.icons import (
     icons8_laser_beam_hazard2_50,
     icons8_pause_50,
     icons8_pentagon_50,
+    icons8_save_50,
 )
 from meerk40t.gui.wxutils import disable_window
 from meerk40t.kernel import lookup_listener, signal_listener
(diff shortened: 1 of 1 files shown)


tatarize:
[myfile.zip](https://github.com/meerk40t/meerk40t/files/9528080/myfile.zip)

Draw circle. Click Save in LaserPanel. Get file.

```
Document type : LHYMICRO-GL file
File version: 1.0.01
Copyright: Unknown
Creator-Software: MeerK40t v0.8.0023 Beta15 git

%0%0%0%0%
IBzzzzzzzzzzzzzzzzzzzzz157Rzzzzzzzzzz206S1P
ICV2551031128000000CNTRS1E|mMaRwDaUdMaRsMaRoMaRnMaReDaUeMaRjMaRjMaRiMaRhMaRhMaRfDaUaMaRgMaRfMaRgMaRfMaRfMaRfMaReMaRfMaRbDaUbMaReMaReMaReMaRdMaReMaRdMaReMaRdMaRdMaRdMaRdDMaURdMaRdMaRdMaRcMaRdMaRcMaRdMaRcMaRdMaRcMaRcMaRcMaRdMaRbDaUMaRcMaRcMaRcMaRcMaRbMaRcMaRcMaRcMaRcMaRbMaRcMaRbMaRc...R|cMaRdDaU|hFNSE-
```

This file is fairly close to being compatible with Whisperer and the various chinese software. It however has some altered/upgraded code structures. For example the `\n` commands means basically go ahead and send that. And the `-` command triggers the wait-finished (without that it will automatically try waiting finish without ever having sent an `F` command before the NSE. So you could do simple rapid commands without requiring the code be entirely within programmed mode.

* The raw-save only works for M2 Nano. 
* I've hardcoded the file type into the panel rather than defer to the device driver.

PR #1317, opened 2022-09-10
Beta 16: Bug Fixes. Feature Freeze.

tatarize:
Is it a feature? Create a new PR. We need to just proof this stuff and get it out the door.

* Note the update-camera-background thing was before freeze.

You can still implement features. They just need to be in a different PR and won't merge together until 0.8.x full is released. Unless they replace a feature lost between 0.8.x and 0.7.x. You could even merge some of those in a pre-0.8.2 branch or whatever, but unless it fixes something and fixes something *important* like crashing go ahead and put it in a different PR.

* I disabled `grblcontrol` and `ruidacontrol` for now since they aren't really worth fully proofing out.
* Hybrid font drawing and measuring methodology, works like 95% of the time.


meerk40t/core/webhelp.py
```diff
@@ -7,6 +7,7 @@
 DISCORD_MEERK40T = "https://discord.gg/6wxrJ2hn"
 MAKERS_FORUM_MEERK40T = "https://forum.makerforums.info/c/k40/meerk40t/120"
 IRC_CLIENT = "http://kiwiirc.com/client/irc.libera.chat/meerk40t"
+MEERK40T_FEATURE = "https://github.com/meerk40t/meerk40t/discussions/1318"
```
tatarize, line 10: This discussion is tiger asking about version or whatever, and seems like feature requests would be more just raise an issue. Also, I thought I got rid of the irc thing, since that's not a great way to contact us.

meerk40t/core/node/op_raster.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, *args, **kwargs):
             "elem rect",
             "elem line",
             "elem text",
-            "elem image",
+#            "elem image",
```
tatarize, line 57: Hm. You can raster an image though. I see the logic in excluding it but if you wanted an image and then you wanted to obscure that image with a big white heart or something to put something else in there it would need to include images.

Issue #1320, opened 2022-09-10
[0.7.7] Burns partially inside a closed cut should be burned first


Sophist-UK:
At present if you have `Burn Inner First` set, burns that are partially inside a closed cut are **not** burned first (even though the inside may drop out before it is burned).

1. An inside burn with the outside end actually on the boundary of the closed cut are not considered inside. If point 2, isn't addressed, we should add a small (0.5mm?) margin to the outside of the closed cut to check whether a burn path is inside.
2. A non-closed cut or any other type of burn which **crosses** the closed cut should be considered inside so that it gets burned before the piece drops out. To avoid difficulties with two closed path cuts which cross, where deciding which is inside which is ambiguous, you should probably only check for closed cuts being entirely inside other closed cuts.

[Cuts not treated inside.svg](https://user-images.githubusercontent.com/3001893/189484010-f8eb7932-6f32-4feb-ab9c-3d4be3719e61.svg)

jpirnay:
![grafik](https://user-images.githubusercontent.com/2670784/189488865-ef34d2fe-00b0-43a7-a2cb-f7832a2c8a37.png)
I fail to see the issue in your file. Which part of the burn are you referring to?

Sophist-UK:
It may depend on Execute settings - I use Group Inner Burns and the short red open line inside the outside closed path are cut at the very end with Group Inner Burns set rather than being cut before each of the outer curves.

tatarize:
> A non-closed cut or any other type of burn which crosses the closed cut should be considered inside so that it gets burned before the piece drops out. To avoid difficulties with two closed path cuts which cross, where deciding which is inside which is ambiguous, you should probably only check for closed cuts being entirely inside other closed cuts.


This actually gets notably harder to detect because it would also hold true for any line segments which do not have a point within the shape. Picture a circle with two line segments going entirely across the circle with no vertexes inside that. 

![example](https://user-images.githubusercontent.com/3302478/189514902-a040aa36-bb80-49bc-9dec-932927a4db18.png)

You actually would need to check a variety of line segments whether they were ever inside another line segment. It wouldn't be enough to merely detect whether or not a bit of path ever had a point inside another closed path. And really this ambiguity applies to closed paths too.

![example2](https://user-images.githubusercontent.com/3302478/189514982-94570ee7-18c4-47b6-8f9b-500d7ce22b1d.png)

[...]

Issue #1293, opened 2022-08-31
[0.7.7] Semi-transparent white PNG pixels being burned but with laser off


Sophist-UK:
PNG raster lines containing all fully-transparent pixels or partially-transparent white pixels, are being burned with laser off. Resulting image is correct, but takes longer to burn.

I am not expecting this to be fixed in 0.7 - I am reporting it so that this can be corrected in 0.8 if it still exists. (My limited experience with 0.8 is that it didn't happen.)

Also, some but not all raster wizards also treat these same pixels as black.

tatarize:
I need more details and maybe an example of this. Fixes in 0.7.x are still permitted and will probably be done. They just have to be done with an extremely narrow scope that modifies as little code as possible to fix a very specific error. We actually have a few things in 0.7.x that need fixing. If the fix isn't difficult and will not create any additional bugs that's absolutely a thing that goes into maintenance mode.

Secondly, please actually recheck that the laser was *entirely* off. The only reason it would go to a place is to laser it in raster mode. Get the example bit you checked and see that it didn't actually fire a single laser pulse or two in that area. You might need to look at an angle or whatnot. But, the code doesn't really allow the behavior you're describing. If it's going to that location in a raster, it's going there to fire at least one pulse.

---

* Ruidacontrol RunStatus reply is now known.
* Corrections for wx4.1 / py3.10. (There's crashing if they run this combo from source)*
* Correctiions for wx4.2 / py 3.10. (This actually mostly works and a few windows need fixing)*

`*` - python 3.10 changed the binding rules to no longer silently permit floats to be cast to ints when going to C++. These need to be overt, wxPython is entirely made of those things so we have a special test case of `wx.Size(0.5)` and if that crashes we know the python and wxPython aren't compatible, the exception being wxPython 4.2 which mostly fixed everything internal but we'll need to cast a few values to int in MK source.

Sophist-UK:
I checked the png - they were semi transparent white pixels.

I have no idea whether the laser was firing because it was burning faintly om paper i.e. 100ppi and low power so beam was not visible.

The PNG file is the one on the 0.8beta13 bugs issue.

You can see the artefacts if you use the raster wizard on this PNG - you get black dashes around the edge.

Issue #1165, opened 2022-07-12
[0.7.7] Time estimates way, way off


Sophist-UK:
Time estimates for rastering a square are way, way out.

25mm x 25mm 30mm/s step 4 estimated at 1m22s and actually takes about 4m-5m.

See #847 for suggestion how to do this easily. See also #342.

P.S. Although I am reporting this on 0.7.7, I think fix should only be in 0.8.

jpirnay:
I assume the values are still to be amended, but in 0.8b6 the estimates provided by the ops themselves are exposed now:
<img width="392" alt="image" src="https://user-images.githubusercontent.com/2670784/178964105-fa6272f7-710f-4f1a-8569-955839c47b0e.png">
Maybe this helps for further adjustments

tatarize:
The operations should take different amounts of time for the different backend drivers we may need. The math though is probably due to bad research on the amount of time it takes to do a swing. Since each swing takes a certain amount of time and I think it's mostly doing the math based on the difference, but the turn-around is the bulk of that and something like 0.7 seconds. When you compare that to the time for a single scanline of an inch you'd be going 30mm/s which should take less. In fact we could probably calculate it from the square raster. If we say 4.5minutes and we're doing a 1 inch square at step 4 that's 250 scan lines. If we went at full 30mm/s we'd be doing 250 inches of distance which would be like 211 seconds so 3.51 minutes. And we're losing about 1 minute to turn around time which is probably equivalent to traveling about 71 inches / 250 is about 284 mils of distance added to the end of the scanlines. Though that's a bit predicated on it having given 3.5 minutes there rather than 1m22s.

Sophist-UK:
No - this is NOT the case. I was rastering (NWT) at 30mm/s - each swing was taking 4 seconds. So the turn around time of a fraction of a second was very small by comparison. Yet the estimate was out by a factor of 3x-4x.

Issue #1162, opened 2022-07-12
[0.7.7] Delete key on selected operation nodes deletes the element nodes too!!!!!


Sophist-UK:
This is just plain wrong.

The delete functionality is easy - it deletes **selected** nodes and any children and does not delete **emphasised** nodes.

I fixed this in #793 so no idea why it is now broken again. But without Undo it is **bloody annoying** when you try to delete opnodes and it deletes the elements too.

jpirnay:
Fixed in 0.8

tiger12506:
Well the conflicting demands here seem to be conflicting definitions of what emphasized means. It was my understanding that emphasized was a workaround to support broken tree selection code on other platforms (mac). The idea being that emphasized and selected would be kept in sync. So it then came to be that emphasized meant 'selected in the scene' and selection meant 'clicked on in the tree'.
However, the code for that has been changed at least a thousand times probably before, during, and since you looked at it Sophist because the emphasized->selection->emphasized event handling was recursively triggering on other platforms and causing massive slowdowns. And then changed again because of broken behavior. Etc. etc. rinse, repeat.

Sophist-UK:
My understanding of `emphasised` is that it is children of selected nodes **and** equivalent nodes in the other part of the tree i.e. opnodes for selected / emphasised element nodes and possible (or maybe not) vice versa.

It seems to me that there needs to be clarity on the definition of these things otherwise:

a. The code to emphasise nodes may end up being incorrect;
b. Selected nodes may be used when emphasised nodes should be used or vice versa.

jpirnay:
We can leave this open for 0.7, but it's confirmedly fixed in 0.8

PR #1327, opened 2022-09-11
Legacy7: 0.7.8

tatarize:
Check the 0.7.x issues. Fix some incompatible version issues.

- [x] Correct check for py3.10 wx4.1.1. This should prevent execution of incompatible versions.
- [ ] #1320
- [x] #1296
- [ ] #1293 
- [ ] #1165 
- [ ] #1162 
- [x] Still needs to check for crashes in the GUI from other TypeError things that aren't corrected.
- [x]  Ruidacontrol should be given the correct response for read-run.


tatarize:
#1330

tatarize:
0.7.7 also can suffer save crash if never mode shifted for the image and attempting to save CYMK which cannot be used for

PR #1328, opened 2022-09-12
Default operations

jpirnay:
image

Default Start / End operations are back (with a twist, you can add any console command now to both start and end).

Needs some validation, therefore no immediate merge

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.formatting.provider": "autopep8"
+}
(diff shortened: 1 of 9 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.formatting.provider": "autopep8"
+}
```
tatarize, line 3: This is probably better excluded since we don't generally require developers to use anything though we do apply black and isort here and there.

PR #1335, opened 2022-09-17
User defined node-labels

jpirnay:
image


tatarize:
How do you know what {} are allowed for the particular node. There are a number of them but they don't all list or get used?

jpirnay:
I create a dummy instance of each node type and query its dictionary

meerk40t/gui/choicepropertypanel.py
```diff
@@ -20,6 +20,68 @@ class ChoicePropertyPanel(ScrolledPanel):
     either registered in the Kernel as a choice or called directly on the ChoicePropertyPanel you can make dynamic
     controls to set various properties. This avoids needing to create a new static window when a panel is just
     providing options to change settings.
+    The choices need to be provided either as list of dictionaries or indirectly via
```
tatarize, line 23: This is fine for including anywhere even directly into main. Documentation, especially useful documentation basically doesn't need a PR. That's just great.

PR #1342, opened 2022-09-20
Add viewport parameters to moshi & grbl (changed)

jpirnay:
Reestablish/Introduce consistent home_bottom, home_right, flip_x, flip_y, swap_xy to moshi and grbl driver.

### DIFF
--- a/meerk40t/gui/wxmscene.py
+++ b/meerk40t/gui/wxmscene.py
@@ -805,23 +805,8 @@ class MeerK40tScenePanel(wx.Panel):
 
     @signal_listener("bedsize")
     def on_bedsize_simple(self, origin, *args):
-        dimx = None
-        if hasattr(self.context.device, "bedwidth"):
-            dimx = self.context.device.bedwidth
-            dimy = self.context.device.bedheight
-        elif hasattr(self.context.device, "lens_size"):
-            dimx = self.context.device.lens_size
-            dimy = dimx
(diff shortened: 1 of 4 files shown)


meerk40t/gui/wxmscene.py
```diff
@@ -812,6 +812,10 @@ def on_bedsize_simple(self, origin, *args):
         elif hasattr(self.context.device, "lens_size"):
             dimx = self.context.device.lens_size
             dimy = dimx
+        if hasattr(self.context.device, "home_bottom") and hasattr(self.context.device, "origin_y"):
+            self.context.device.origin_y = 1.0 if self.context.device.home_bottom else 0.0
+        if hasattr(self.context.device, "home_right") and hasattr(self.context.device, "origin_x"):
+            self.context.device.origin_x = 1.0 if self.context.device.home_right else 0.0
```
tatarize, line 818: This part is the one part that sort of weirds me out and I need to double check.

tatarize (changes requested):
At a minimum that section in wxmscene has way too much coupling. Moreover it doesn't see that home_right /flip_x is really a thing that demands the origin switch if that stuff is properly needed it would exist within self.context.device.realize() and certainly has no place in scene itself.

meerk40t/gui/wxmscene.py
jpirnay: That has been eliminated

tatarize (approved):

PR #1348, opened 2022-09-26
Translation and final checks

tatarize:


meerk40t/core/elements.py
```diff
@@ -311,6 +311,20 @@ def plugin(kernel, lifecycle=None):
...
+            {
+                "attr": "classify_debug",
+                "object": elements,
+                "default": False,
+                "type": bool,
+                "label": _("Debug Classification"),
+                "tip": _("Shows info about the classification process in the console"),
+                "page": "Classification",
+                "section": "_AA_Debug",
+                "hidden": True,
```
tatarize, line 324: We're now giving ourselves fancy special options in preferences so if you have developer mode on you have the convenience of pressing this button rather than doing something like `batch add channel print classification`

meerk40t/core/elements.py
```diff
@@ -8943,6 +8949,15 @@ def classify(self, elements, operations=None, add_op_function=None):
                 because of a lack of classification options.
         @return:
         """
+        def emptydebug(value):
+            return
+
+        # I am tired of changing the code all the time, so let's do it properly
```
tatarize, line 8955: Do it *really* properly. `debug = context.channel("classification")`

Then it's a channel. You can print that, save to disk, open in the console to be watched.

meerk40t/core/spoolers.py
```diff
@@ -812,7 +815,11 @@ def remove(self, element, index=None):
                     )
                 except AttributeError:
                     pass
-                self._queue.remove(element)
+                try:
+                    self._queue.remove(element)
+                except ValueError:
+                    # We might have waited for too long, the job is no longer there...
```
tatarize, line 821: with self._lock:

tatarize: We have a thread.Lock() value that we used above. If for some reason the job could have been removed elsewhere while we're doing the work that elsewhere needs to be also put into the synchronization gate. Whatever changes the queue should be locked so only one thing fiddles with that at a time.

jpirnay: Its a simple case to test, where a lock makes limited sense: run job, right click in spoolerpanel while it still runs, select but don't execute yet the option to cancel the job, job terminates while context menu is still open, click...
A lock would not help.

meerk40t/core/elements.py
```diff
@@ -8943,6 +9007,12 @@ def classify(self, elements, operations=None, add_op_function=None):
                 because of a lack of classification options.
         @return:
         """
+        def emptydebug(value):
```
tatarize, line 9010: pointless function now

meerk40t/gui/choicepropertypanel.py
```diff
@@ -50,9 +51,18 @@ def __init__(
         if choices is None:
             return
         if isinstance(choices, str):
-            choices = self.context.lookup("choices", choices)
+            tempchoices = self.context.lookup("choices", choices)
+            # we need to create an independent copy of the lookup, otherwise
+            # any amendments to choices like injector will affect the original
+            choices = []
+            for c in tempchoices:
+                choices.append(c)
```
tatarize, line 59: Should probably just do `choices = list(tempchoices)`

meerk40t/gui/wxutils.py
```diff
@@ -471,21 +520,42 @@ def set_range(self, range_min, range_max):
...
+            txt != "" and self._check == "length" and self.extend_default_units_if_empty
+        ):
+            # Do we have non-existing units provided? --> Change content
+            purenumber = True
+            unitstr = "".join(ACCEPTED_UNITS)
+            for c in unitstr:
+                if c in txt:
+                    purenumber = False
+                    break
+            if purenumber and hasattr(self.parent, "context"):
```
tatarize, line 541: You can't type in balor properties for lens size given this code.

PR #1353, opened 2022-09-30
Additional Fixes for Final hopefully beta

tatarize:


meerk40t/gui/choicepropertypanel.py
```diff
@@ -567,7 +567,7 @@ def check(event=None):
                         current_value = getattr(obj, param)
                         if current_value != current:
                             setattr(obj, param, current)
-                            self.context.signal(f"{param}", v)
+                            self.context.signal(f"{param}", v, obj)
```
Sophist-UK, line 570: Unclear why this is not `self.context.signal(param, v, obj)` as elsewhere.

tatarize: There's a bit of obscurity here really he should have used the origin. Since that would avoid the issue. All listeners are given origin and any operands they were given. Here he's using the object as a semaphore but really the origin would work as an equally good semaphore. In theory if something listened to choice panel changes without being a choice panel it might now crash, but nothing really came to mind as doing that. And while this added some extra complexity we sort of swept it under the rug. --- Or if you mean the pointless f"{param}" which is identical to param... I don't know it's not going to harm anything but it is sort of pointless.

Sophist-UK: yes - the pointless string substitution.

tatarize: fixed

PR #1356, opened 2022-10-01
Future bug fixes.

tatarize:


meerk40t/extra/vectrace.py
```diff
@@ -144,3 +123,234 @@ def _vectrace(pixels, width, height):
...
+            )
+        return "elements", paths
+
+"""
+Potracer routines, please be aware that potrace is not part of the standard
+distribution of meerk40t due to the more restrictive licensing of potrace
+If you have installed it yourself via 'pip install potracer', well why not...
+potracer is a pure python port of Peter Selinger's Potrace by Tatarize
+potrace:    https://potrace.sourceforge.net/
+potracer:   https://github.com/tatarize/potrace
```
tatarize, line 167: We should probably check if pypotrace works and make sure they are somewhat interchangeable. Pypotrace is potrace compiled from c-source with python bindings. It often crashes the compile and doesn't work consistently. Which is why replacing it with `potracer` would be doable.

meerk40t/extra/vectrace.py
```diff
@@ -144,3 +123,234 @@ def _vectrace(pixels, width, height):
...
+        turdsize=None,
+        alphamax=None,
+        opticurve=None,
+        opttolerance=None,
+        color=None,
+        invert=None,
+        blacklevel=None,
+        **kwargs,
+    ):
+        if turnpolicy not in (
```
tatarize, line 283: While not wrong, this should probably be a function that gets registered into the lookup. Or something that would let us access the code through a text-to-path routine.

meerk40t/extra/vectrace.py
```diff
@@ -1,40 +1,19 @@
...
-                )
-            return "elements", paths
+        setup_vectrace(kernel)
 
+        # A very simple hack....
+        try:
+            import potrace
+            setup_potrace(kernel)
+            print("Good news - potrace found...")
+        except ModuleNotFoundError:
```
tatarize, line 15: This probably should be in it's own file completely and loaded independently and if it can't do the import, return False during the plugin initialization routines.

PR #1368, opened 2022-10-08
Allow optional signal to untoggle ribbonbar button

tatarize:
The light job outlines do not toggle back when finished by running a normal job or directly canceling them in spooler; they should.


tatarize:
This does not unregister the listener during a normal shutdown.

PR #1369, opened 2022-10-08
Required Fixes for next beta

tatarize:
- [x] the stroke none issue in simulation
- [x] the reported multi-pass bug
- [ ] if you take multi-device seriously all the logic that is based on just a single device (red background, leaving while another job is active)
- [ ] interrupt/beep kicking off, although the device is still working
- [ ] Cyclocycloid pentagrams


tatarize:
d8ad7d53fe69a03f96cad7a757c3602e698af692 should read by the sign of the determinant. It doesn't matter if it's even or odd, just whether it's positive or negative.

PR #1376, opened 2022-10-12
0.8.x bug fixes sync

jpirnay:


tatarize:
Some conflicts now exist. I updated shxparser to deal with the potential crashes. It shouldn't crash even if the files are gibberish now.

PR #1395, opened 2022-10-18
Deal better with images with transparency

jpirnay:


tatarize (changes requested):
Could you simplify this rather than adding very similar code 3 different places. ImageNode is basically our home for image. So it's not weird to request a non-transparent image from the node itself. `image_node.opaque_image` to give us the original image with the guarantee that it has no transparency. This would also let you centralize some stuff a bit more.

meerk40t/core/node/elem_image.py
```diff
@@ -297,7 +297,21 @@ def process_image(self, step_x=None, step_y=None, crop=True):
...
+        if img is not None:
+            if img.mode == "RGBA":
+                r, g, b, a = img.split()
+                background = Image.new("RGB", img.size, "white")
+                background.paste(img, mask=a)
+                img = background
+        return img
+
     def _convert_image_to_grayscale(self, image):
+        from PIL import Image
```
tatarize, line 313: pointless import

meerk40t/core/node/elem_image.py
```diff
@@ -526,7 +540,7 @@ def _process_image(self, step_x, step_y, crop=True):
 
         transparent_mask = self._get_transparent_mask(image)
 
-        image = self._convert_image_to_grayscale(image)
+        image = self._convert_image_to_grayscale(self.opaque_image)
```
tatarize, line 543: Recursive and will crash

meerk40t/image/imagetools.py
```diff
@@ -211,6 +211,7 @@ def image_lock(command, channel, _, data, **kwargs):
     def image_threshold(
         command, channel, _, data, threshold_max=None, threshold_min=None, **kwargs
     ):
+        from PIL import Image
```
tatarize, line 214: pointless import

meerk40t/image/imagetools.py
```diff
@@ -1325,10 +1328,12 @@ def dither(image, method="Floyd-Steinberg"):
     :copyright: 2016-2017 by hbldh <henrik.blidh@nedomkull.com>
     https://github.com/hbldh/hitherdither
     """
+
     diff_map = _DIFFUSION_MAPS.get(method.lower())
     if diff_map is None:
         raise NotImplementedError
-    diff = image.convert("F")
+    img = image
+    diff = img.convert("F")
```
tatarize, line 1336: this temp variables does not seem to improve anything

tatarize (changes requested):
I can't get around the idea that that will crash. Also, somewhat pointless imports.

meerk40t/core/node/elem_image.py
tatarize: Eh, it may have presented that wrong.

tatarize (approved):
No crash, rest is fairly fine.

meerk40t/core/node/elem_image.py
jpirnay: Done

meerk40t/image/imagetools.py
jpirnay: Done

PR #1406, opened 2022-10-21
Passthrough images

tatarize:


meerk40t/core/node/node.py
```diff
@@ -211,6 +212,50 @@ def parent(self):
     def root(self):
         return self._root
 
+    @property
```
tatarize, line 215: All nodes were given can_action() values for emphasize, highlight, target, move, scale, rotate, skew, modify, alter, update, remove. (These were never used). The idea was to allow checking their action and by default locked objects could do none of these things.

meerk40t/core/svg_io.py
```diff
@@ -862,6 +862,11 @@ def parse(self, element, context_node, e_list, uselabel=None):
                         _lightness = float(element.values.get("lightness"))
                     except (ValueError, TypeError):
                         pass
+                    _passthrough = None
```
tatarize, line 865: SVG IO would get the passthrough value.

tatarize (commented):
The primary change of this now dead PR was to expand the concept of lock to the concept of each node having a series of actions they could perform or not. And this would be checked rather than permitted and every node would have their own lock, label, and id.

PR #1410, opened 2022-10-23
Fix missing delegates

jpirnay:


meerk40t/moshi/gui/moshidrivergui.py
```diff
@@ -32,8 +33,13 @@ def __init__(self, *args, context=None, **kwds):
         self.SetSizer(sizer_main)
         self.Layout()
         self.SetupScrolling()
+        self.parent.add_module_delegate(self.panel_pref1)
```
tatarize, line 36: likely wouldn't need both.

tatarize (approved):

meerk40t/moshi/gui/moshidrivergui.py
jpirnay: It does.

PR #1411, opened 2022-10-23
Fix paint_bounds

jpirnay:
- Fix calculation of paint bounds
- Make sure changes to stroke / stroke_width lead to a recalculation of the bounds

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -3954,6 +3954,7 @@ class Elemental(Service):
             node = self.elem_branch.add(shape=poly_path, type="elem polyline")
             node.stroke = self.default_stroke
             node.fill = self.default_fill
+            node.altered()
             self.set_emphasis([node])
             node.focus()
             if data is None:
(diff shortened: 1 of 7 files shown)


meerk40t/core/elements.py
```diff
@@ -3978,8 +3979,10 @@ def render_elements(command, channel, _, dpi=500.0, data=None, **kwargs):
             if not make_raster:
                 channel(_("No renderer is registered to perform render."))
                 return
-
+            bounds_regular = Node.union_bounds(data)
             bounds = Node.union_bounds(data, attr="paint_bounds")
+            for idx in range(4):
+                print (f"Bounds[{idx}] = {bounds_regular[idx]:.2f} vs {bounds_regular[idx]:.2f}")
```
tatarize, line 3985: debug

tatarize (approved):

meerk40t/core/elements.py
jpirnay: Gone

PR #1412, opened 2022-10-23
Outline command

jpirnay:
grafik

and fixed existing ouline command and renamed it to ``frame``


tatarize (dismissed):
Lol. That's kind of clever.

jpirnay:
Still suffers from incorrectly generated make_raster images. Does not appear if you create stuff from scratch, but as soon as you transform them it becomes visible. Needs to be fixed first

tatarize (approved):

PR #1428, opened 2022-10-24
Suppress UI-elements if no potrace available

jpirnay:


meerk40t/gui/propertypanels/imageproperty.py
```diff
@@ -1,8 +1,10 @@
 # import threading
 import wx
+from PIL import Image
```
tatarize, line 3: I think this might harden the requirement for Pillow.

tatarize: Nm. It's part of gui.

meerk40t/core/element_treeops.py
```diff
@@ -1450,6 +1450,7 @@ def wlist_minus(node, **kwargs):
                 property_op(self.kernel.root, node)
             self.signal("element_property_update", [node])
 
+    @tree_conditional(lambda node: has_vectorize(node))
```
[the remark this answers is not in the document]
tatarize: the lamba node: could just get called with has_vectorize just as simply.

meerk40t/gui/propertypanels/imageproperty.py
```diff
@@ -956,7 +995,10 @@ def generate_preview(self):
 
     @staticmethod
     def accepts(node):
-        if node.type == "elem image":
+        # Changing the staticmethod into a regular method will cause a crash, so therefore this circumvention
+        # Not the nicest thing in the world, as we need to instantiate the class once to reset the status flag
+        global HAS_VECTOR_ENGINE
```
tatarize, line 1000: This is only needed if we're going to set that value. It's unneeded if we're just checking its value.

meerk40t/core/element_treeops.py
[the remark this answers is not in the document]
jpirnay: Agree, lazy me just copied and pasted a single line...

PR #1434, opened 2022-10-25
outline --outer can deal now with more topologies

jpirnay:
The subvariant --outer requires one additional pass where we disassemble
        the first outline and fill the subpaths, This will effectively deal with
        donut-type shapes

        The need for --inner was't high on my priority list (as it is somewhat
        difficult to implement, --outer just uses a clever hack to deal with
        topology edge cases. So if we are in need of inner we need to create
        the outline shape, break it in subpaths and delete the outer shapes
        manually. Sorry.

### DIFF
--- a/meerk40t/core/element_commands.py
+++ b/meerk40t/core/element_commands.py
@@ -3290,6 +3290,16 @@ def init_commands(kernel):
 
         This two phase approach is required as not all nodes have
         a proper stroke-width that can be adjusted (eg text or images...)
+
+        The subvariant --outer requires one additional pass where we disassemble
+        the first outline and fill the subpaths, This will effectively deal with
+        donut-type shapes
+
+        The need for --inner was't high on my priority list (as it is somwhat
+        difficult to implement, --outer just uses a clever hack to deal with
+        topology edge cases. So if we are in need of inner we need to create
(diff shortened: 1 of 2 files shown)


meerk40t/core/element_commands.py
```diff
@@ -3278,7 +3278,7 @@ def element_outline(
         invert=None,
         blacklevel=None,
         outer=None,
-        times=None,
+        steps=None,
```
tatarize, line 3281: Much better!

meerk40t/gui/laserrender.py
```diff
@@ -782,13 +782,15 @@ def measure_text(self, node):
         @param node:
         @return:
         """
-
-        bmp = wx.Bitmap(1000, 500, 32)
+        dimension_x = 1000
```
tatarize, line 785: We could probably get away with a smaller initial box if we're actually guessing from the revision factor to start with. The box is so big because we're trying to avoid the bounds.

tatarize (approved):

meerk40t/gui/laserrender.py
jpirnay: Agree, that was a 45 seconds fix

PR #1450, opened 2022-10-29
Time Based Simulation Viewing

tatarize:
Please have a look at the bugfix branch, there's a beta of the updated simulation panel.
Allow a time-based view of simulation.
While the usual simulation jumps from one completed cutcode operation to the next, you can now choose a timed mode: the simulation jumps from one second to the next and displaying the progress so far:
 
simulation

There are some limitations still:
- intermediate cutcode operations are interpolated with a linear move from cutcode start to end 
- Raster operations are only shown as a growing rectangle
- Very fast units like balor might need a subsecond resolution
Image
Would like to get some feedback what to change, thanks


meerk40t/gui/simulation.py
```diff
@@ -51,6 +55,8 @@ def __init__(self, *args, context=None, plan_name=None, auto_clear=True, **kwds)
             if isinstance(c, CutCode):
                 self.cutcode.extend(c)
         self.cutcode = CutCode(self.cutcode.flat())
+        self.statistics = self.cutcode.provide_statistics()
```
tatarize, line 58: Does this update if the cutcode updates with the recalculate button?

meerk40t/gui/simulation.py
```diff
@@ -383,14 +419,54 @@ def fit_scene_to_panel(self):
...
+            idx = 0
+        # print(
+        #     f"Cut-Mode={self._playback_cuts}, prog={progress}, idx={idx}, stats={len(self.statistics)}"
+        # )
+        return idx, residual
+
     def on_mouse_right_down(self, event=None):
+        def cut_before(event):
+            step, residual = self.progress_to_idx(self.progress)
+            self.context(f"plan{self.plan_name} sublist {step} -1\n")
```
tatarize, line 459: Does this make sense in time mode? If we are not at a specific point in the cutcode where we can delete half the cutcode and be at this location, we might well try to do a `delete cuts before` partway through a raster or something which could be a bit strange. Though disabling this feature in that mode would suffice.

meerk40t/gui/simulation.py
```diff
@@ -473,11 +567,10 @@ def _refresh_simulated_plan(self):
             if isinstance(c, CutCode):
                 self.cutcode.extend(c)
         self.cutcode = CutCode(self.cutcode.flat())
-        self.max = max(len(self.cutcode), 0) + 1
-        self.progress = self.max
-        self.slider_progress.SetMin(0)
-        self.slider_progress.SetMax(self.max)
-        self.slider_progress.SetValue(self.max)
+        self.statistics = self.cutcode.provide_statistics()
```
tatarize, line 570: Ah it does

meerk40t/gui/simulation.py
```diff
@@ -703,12 +824,72 @@ def __init__(self, scene, sim):
         self.matrix.post_cat(scene.context.device.device_to_scene_matrix())
 
     def process_draw(self, gc: wx.GraphicsContext):
-        if self.sim.progress > 1:
+        if self.sim.progress >= 0:
```
tatarize, line 827: Could do:
```python
if self.sim.progress < 0:
    return
```
Then spare everything else in this block a tab.

tatarize:
I would expect 100% speed and 1 second ticks to move at a speed of 1 second per tick.

jpirnay:
Replaced by updated one

PR #1465, opened 2022-10-31
Spooler

jpirnay:
a) Send signal if a laser is paused/resumed and change some buttons accordingly
b) Issue estop if removal of active job is requested

### DIFF
--- a/meerk40t/gui/laserpanel.py
+++ b/meerk40t/gui/laserpanel.py
@@ -225,6 +225,23 @@ class LaserPanel(wx.Panel):
             self.combo_devices.Append(spools[i])
         self.combo_devices.SetSelection(index)
         self.button_save_file.Enable(hasattr(self.context.device, "extension"))
+        self.set_pause_color()
+
+    def set_pause_color(self):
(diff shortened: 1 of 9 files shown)


meerk40t/gui/laserpanel.py
```diff
@@ -225,6 +225,23 @@ def spooler_lookup(self, *args):
             self.combo_devices.Append(spools[i])
         self.combo_devices.SetSelection(index)
         self.button_save_file.Enable(hasattr(self.context.device, "extension"))
+        self.set_pause_color()
+
+    def set_pause_color(self):
+        new_color = None
+        new_caption = _("Pause")
+        try:
+            if self.context.device.driver.paused:
```
tatarize, line 234: This isn't consistent for all drivers. Though in an attribute test so it won't matter much.

tatarize (approved):

meerk40t/gui/laserpanel.py
jpirnay: It is for most now, ruida the most notable exception

PR #1469, opened 2022-11-02
Fix enabled status for read util ops

jpirnay:


meerk40t/core/svg_io.py
```diff
@@ -967,6 +967,12 @@ def parse(self, element, context_node, e_list, uselabel=None):
 
                 try:
                     op = self.elements.op_branch.add(type=node_type, **attrs)
+                    overlooked_attributes = [
+                        "output",
+                    ]
+                    for overlooked in overlooked_attributes:
+                        if overlooked in element.values and hasattr(op, overlooked):
+                            setattr(op, overlooked, element.values.get(overlooked))
```
tatarize, line 975: Hm. Seems like there should be a better way to do this.

tatarize (approved):

meerk40t/core/svg_io.py
jpirnay: If you have one, go ahead - that's just a quickly hacked together fix.

PR #1470, opened 2022-11-02
Allow tolerance for inner burn optimisations

jpirnay:
image

### DIFF
--- a/meerk40t/core/cutplan.py
+++ b/meerk40t/core/cutplan.py
@@ -429,7 +429,7 @@ class CutPlan:
                 tolerance = float(Length(stol)) * 2 / (self.context.device.native_scale_x + self.context.device.native_scale_y)
             except ValueError:
                 pass
-        print(f"Tolerance: {tolerance}")
+        # print(f"Tolerance: {tolerance}")
 
         channel = self.context.channel("optimize", timestamp=True)
         grouped_inner = self.context.opt_inner_first and self.context.opt_inners_grouped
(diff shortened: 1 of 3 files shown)


meerk40t/core/cutplan.py
```diff
@@ -422,12 +422,21 @@ def optimize_cuts(self):
...
         @return:
         """
+        tolerance = 0
+        if self.context.opt_inner_first:
+            stol = self.context.opt_inner_tolerance
+            try:
+                tolerance = float(Length(stol)) * 2 / (self.context.device.native_scale_x + self.context.device.native_scale_y)
+            except ValueError:
+                pass
+        print(f"Tolerance: {tolerance}")
```
tatarize, line 432: We don't tend to print things.

meerk40t/core/cutplan.py
```diff
@@ -444,12 +453,21 @@ def optimize_travel(self):
...
         except AttributeError:
             last = None
+        tolerance = 0
+        if self.context.opt_inner_first:
+            stol = self.context.opt_inner_tolerance
+            try:
+                tolerance = float(Length(stol)) * 2 / (self.context.device.native_scale_x + self.context.device.native_scale_y)
+            except ValueError:
+                pass
+        print(f"Tolerance: {tolerance}")
```
tatarize, line 463: Print.

meerk40t/core/cutplan.py
```diff
@@ -478,15 +496,19 @@ def clear(self):
         self.commands.clear()
 
 
-def is_inside(inner, outer):
+def is_inside(inner, outer, tolerance=0):
```
tatarize, line 499: So if we have two overlapping circles are they both inside each other? Then which goes first? I guess it doesn't matter since they are overlapping.

tatarize (dismissed):
I dunno about the print statements.

meerk40t/core/cutplan.py
jpirnay: The tpology is a bit hard to imagine where that would be the case, in essence if the diameter of the circle would be smaller than the given tolerance then that woul happen. Will try it out what happens...

jpirnay:
> I dunno about the print statements.

Removed

PR #1475, opened 2022-11-03
Fix crash for stroke-property display

jpirnay:


jpirnay:
Will force that one through as it is a critical crash.

Issue #1476, opened 2022-11-03
[Bug]: Stroke_width inconsistently saved


jpirnay:
### Summary Description

Stroke_width is stil inconsistently treated
Inkscape file: 

![line_width](https://user-images.githubusercontent.com/2670784/199702601-05eb114d-eeab-4dd1-8861-7d7212106623.svg)
![bug](https://user-images.githubusercontent.com/2670784/199702907-6a1802f4-af2c-4f40-ae0a-2eabea0a6aa4.gif)

### MeerK40t Version

0.8.1 mainbranch (development)

### MeerK40t Type

Source (zip/tar file from Github)

### Your Operating System

Windows

jpirnay:
PR1475, 1477 and 1478 should fix all of this

PR #1477, opened 2022-11-03
Fix Stroke Width (in native no scale)

tatarize:


jpirnay (approved):

tatarize:
#1476 fixed. 

It's two different errors.
Both are caused by switching to corrected scales.
So all the paths are generally allowed to be whatever units with a matrix and we don't constantly reify on load.
So for saving this was multiplied by the old stroke-scale which was needed because they were saved at px level rather than native. I commented out that explicit units section because we were no longer in px so converting to px was unneeded. We were 1 px = 1 tat.
For the other issue it was actually the stroke-width command never took the matrix into account. This probably was the reason the status-bar strokewidth thing was a pain to use sometimes.

PR #1479, opened 2022-11-03
Fix drag&drop for regmark nodes

jpirnay:
a) Drag & drop of regmark nodes directly to an operation was possible -> prevented
b) Drag & drop of regmark nodes back to elements did not classify (if option set)

### DIFF
--- a/meerk40t/core/node/op_cut.py
+++ b/meerk40t/core/node/op_cut.py
@@ -108,7 +108,7 @@ class CutOpNode(Node, Parameters):
     def drop(self, drag_node, modify=True):
         # Default routine for drag + drop for an op node - irrelevant for others...
         if drag_node.type.startswith("elem"):
-            if not drag_node.type in self._allowed_elements_dnd:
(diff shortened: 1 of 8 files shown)


meerk40t/core/node/op_cut.py
```diff
@@ -108,7 +108,7 @@ def default_map(self, default_map=None):
     def drop(self, drag_node, modify=True):
         # Default routine for drag + drop for an op node - irrelevant for others...
         if drag_node.type.startswith("elem"):
-            if not drag_node.type in self._allowed_elements_dnd:
+            if not drag_node.type in self._allowed_elements_dnd or drag_node._parent.type == "branch reg":
```
tatarize, line 111: PEP8 and Codacy tend to agree that if `drag_node.type not in ....` is better. This doesn't get auto fixed by black though I did go through and change them at some point.

tatarize (dismissed):

meerk40t/core/node/op_cut.py
jpirnay: Done

PR #1488, opened 2022-11-05
Undo Correction

tatarize:
* Fix remaining issues in undo


tatarize:
I'm fairly close it needs to save the classifications which it currently doesn't do.

PR #1497, opened 2022-11-18
Add in PostExec kernel commands

tatarize:
The use case of `classify` is that it should happen once a command completing data is executed. However, due to the nature of command chaining there's not a good indication of when that would be.

`circle 3in 3in 1in stroke blue fill green rotate 0.4turn` -- is 4 commands. If classification happened after `circle` then it might miss that `stroke blue` happened which changed the classification. If it happened after `stroke blue` it might miss the `fill green` command which could change the classification. And if we looked at the `rotate 0.4turn` we might not consider that this shape requires classification at all.

We expand the kernel commands adding in `post` which is a list of commands to be executed. This command is fed the final chained information, channel, data, data_type it ended with and all the commands from `post_data`, which could contain nearly anything. The `post_data` could also be used as a simple chained dictionary from the previous commands. So checking it for flags or other data that could have been set by previous commands would be acceptable.

### DIFF
--- a/meerk40t/core/element_commands.py
+++ b/meerk40t/core/element_commands.py
@@ -52,6 +52,26 @@ def init_commands(kernel):
 
     _ = kernel.translation
 
+    def classify_new(data):
+        """
+        Why are we doing it here? An immediate classification
+        at the end of the element creation might not provide
+        the right assignment as additional commands might be
+        chained to it:
+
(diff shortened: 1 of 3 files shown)


tatarize:
The first commit there got X because it failed the test with the signals doing the classification out of order.

PR #1500, opened 2022-11-18
Living hinges

jpirnay:
Not completely ready, but good enough for wider testing and feedback:
Missing: 
a) ~~Saving presets per hinge-type~~
b) ~~Bowling pin shape not ready~~
c) ~~Circle shape not ready~~
d) Application to non rectangular shapes

Everything else (including clipping) should be functionally okay, but might = will require usability tuning.

Usage: you can either select an existing shape (eg a rectangle) or call the tool directly without an existing selection (will use a predefined size) under Tool - Lasertool - Living Hinges
image


meerk40t/tools/livinghinges.py
```diff
@@ -0,0 +1,1129 @@
...
+        self.param_b = 0.7
+        # Requires recalculation
+        self.path = None
+        self.pattern = []
+        self.cutshape = ""
+        self.set_cell_values(10, 10)
+        self.set_padding_values(5, 5)
+        self.set_predefined_pattern("line")
+
+    def get_patterns(self):
```
tatarize, line 62: These should probably be dynamic, and store their name and the appended pattern in the same routine.

meerk40t/tools/livinghinges.py
```diff
@@ -0,0 +1,1129 @@
...
+                            partial += 1
+                current_x= p[0]
+                current_y = p[1]
+            if current_x != part_of_path.end[0] or current_y != part_of_path.end[1]:
+                add_move(newpath, part_of_path.end)
+            # print (f"From iterated line: added={added}, partial={partial}, ignored={ignored}")
+
+        def add_move(addpath, destination):
+            # Was the last segment as well a move? Then just update the coords...
+            if len(addpath._segments)>0:
```
tatarize, line 524: You can access this through the path itself. `len(addpath)` is equal to this also `addpath[-1]` is equal. to the segments bits.

meerk40t/tools/livinghinges.py
```diff
@@ -0,0 +1,1129 @@
...
+                elif statex == "inside" and statey == "inside":
+                    # Fully inside, so append
+                    newpath.cubic(e.control1, e.control2, e.end)
+                    not_deleted += 1
+                else:
+                    approximate_line(e, current_x, current_y)
+                    partial_deleted += 1
+                current_x = e.end[0]
+                current_y = e.end[1]
+            elif isinstance(e, Arc):
```
tatarize, line 661: There's also a command that replaces arcs with the cubic curves before hand since that's generally an easier way to avoid them.

tatarize (dismissed):

tatarize:
While there's things to be maybe improved it's self-enclosed and can't hurt anything.

meerk40t/tools/livinghinges.py
jpirnay: Done

PR #1569, opened 2022-12-19
Geomstr-Polybool Implementation

tatarize:
The goal is to implement Martinez-Rueda polygon clipping with multiple polylines.

https://en.wikipedia.org/wiki/Martinez%E2%80%93Rueda_clipping_algorithm

See:
https://sean.cm/a/polygon-clipping-pt2

The primary method of the algorithm is to find all intersections between all geometries such that each event of the scanline cannot itself possess an intersection. We do this for all lines and we like need to divide all cubics and quads to be purely monotonic and rational shapes. Such that there are no inflection points with regard to the scanline, by splitting at the extrema points.

segments3 (2)

Also, the subject and clip distinctions are less needed because we can take many different polygons and combine them such that that we have the N bits of insideness information for each side of every segment. As well as a N bits of belongs information.

So if we combine three different geometries, every segment has 9 bits of data.
Right: G1, G2, G3.
Left. G1, G2, G3
Belongs: G1, G2, G3

A segment can overlap and thus belong to more than one geometry. With this data we can solve for every different type of boolean operation as well as any multi-operations. If performing a union of an entire stack of geometries we can solve for whether one side, but not the other has G1, G2, and G3. We could also solve for the intersection of G1 and G3 or G2 and G1 without needing to compute anything else. 

Further the combined geometries should allow for different fill types. Not just `even-odd` or `non-zero` but weirder things like `open` and `zero`. Where we supply the segments but nothing is ever inside them, or a closed path which includes the universe except for this one tiny area. With these we can do things like polyline clipping. We match everything that has both sides included in G1. This omits the original closed path, but captures all segments contained within that closed path. If we do the same operation between an `open` and a `zero` filled geometry, we get a polyline occlusion.


tatarize:
May be worth revisiting, I am advised to check into polygon polycut and living hinges code.

tatarize:
I did a bunch of work, and can confirm this work is largely unneeded. Basically acceleration structure one is concerned with is the event/active table based on x,y position of the event and the list of active segments at at those positions (sorted based on their x), such that no segment ever rises above the other segments during that event space.

With a structure like this we merely look up the values and can do PIP and we can also do CAG with winding, even/odd, or polycutting. The simple structure tells us a lot about the underlying polygon and makes most of the calculations much simpler.

PR #1570, opened 2022-12-20
Extended path simplification

jpirnay:
Extend simplify to use a multi-pass approach:
1) Delete zero length lines generated from the algorithms (new)
2) Simplify lines that have a lot of in-between points (old)
3) Join path segments that have a coinciding end points (new)
4) Join subpaths that end in a line segment that are coinciding
image

### DIFF
--- a/meerk40t/core/elements.py
+++ b/meerk40t/core/elements.py
@@ -2869,7 +2869,6 @@ class Elemental(Service):
                 filetypes.append(f"*.{extension}")
         return "|".join(filetypes)
 
-
     def simplify_node(self, node):
         def my_sign(x):
             # Returns +1 for positive figures, -1 for negative and 0 for Zero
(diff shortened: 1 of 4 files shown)


meerk40t/core/elements.py
```diff
@@ -2964,4 +3223,10 @@ def my_sign(x):
                 pt_older = pt_old
                 pt_old = pt
             after = len(obj.points)
+            time_pass1 = time()
```
tatarize, line 3226: left over debug code. If we aren't printing we don't need this either.

meerk40t/core/elements.py
```diff
@@ -2874,15 +2874,59 @@ def my_sign(x):
             # Returns +1 for positive figures, -1 for negative and 0 for Zero
             return bool(x > 0) - bool(x < 0)
 
-        from meerk40t.svgelements import Line
+        def list_subpath_bounds(obj):
```
tatarize, line 2877: The API for Path objects basically treats them like a list anyway. You don't need to dig into the class and access the `._segements` attribute. This bit of code seems like it could use some `path.as_subpath()` code.

meerk40t/core/elements.py
```diff
@@ -2924,6 +2968,222 @@ def my_sign(x):
...
+
+            redo = True
+            while redo:
+                # Dont do it again unless indicated...
+                redo = False
+                reason = ""
+                results = list_subpath_bounds(obj)
+                # print (f"Examine {len(results)} chains")
+                if len(results) <= 1:
+                    print("only one chain, exit")
```
tatarize, line 2986: Debugging code still stuck in.

tatarize (approved):

tatarize:
Generally it's fine, the debug code like needs to be removed. Also, I each pass is completely independent. You should actually make those different functions. And have simplify call each of these functions.

tatarize:
Each pass should just be a different function. Since they aren't each doing something that depended on the previous pass. They are actually each doing a very specific optimizations.

meerk40t/core/elements.py
jpirnay: Removed

meerk40t/core/elements.py
jpirnay: Did this as I needed to reshuffle the segments anyway, so I assume at that point I would need _segments

meerk40t/core/elements.py
jpirnay: Removed

jpirnay:
> Each pass should just be a different function. Since they aren't each doing something that depended on the previous pass. They are actually each doing a very specific optimizations.

Done

PR #1575, opened 2022-12-22
Balor: Some gui fixes / better connectivity recovery

jpirnay:
- Fix usage of wrong 'connected'-attribute
- Better recovery if no driver is found

### DIFF
--- a/meerk40t/balormk/gui/balorcontroller.py
+++ b/meerk40t/balormk/gui/balorcontroller.py
@@ -104,7 +104,7 @@ class BalorControllerPanel(wx.ScrolledWindow):
         if status is None:
             status = "Unknown"
         try:
-            connected = self.service.connected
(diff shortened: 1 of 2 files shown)


tatarize (approved):

meerk40t/balormk/gui/balorcontroller.py
```diff
@@ -104,7 +104,7 @@ def on_usb_update(self, origin=None, status=None):
         if status is None:
             status = "Unknown"
         try:
-            connected = self.service.connected
+            connected = self.service.driver.connected
         except AttributeError:
```
tatarize, line 108: The except attribute error here would have let this error be detected otherwise. Though I guess a different `device` might invoke something different for that, and crash there. Seems like it should maybe not check for that error but rather correctly check if the type of device is actually a balor device somehow.

PR #1577, opened 2022-12-22
Polybool clipping

tatarize:


jpirnay (approved):

jpirnay:
Looks good, although I would add a message if no solution was found:
```
[...]
            else:
                channel(_("No solution found (empty path)"))
                return "elements", []
```

tatarize:
Yeah, I don't know if this should just be switched out. I might need to run some speed checks on the data to see if there's some performance costs or benefits. Also, I added `-c` to main and to this branch and sort of wonder if that should be made the default instead. Replacing the original pieces.

PR #1588, opened 2022-12-26
Preparations for node edit

jpirnay:


tatarize (approved):
Other than the print statement these are fine changes that should affect or do nothing, just make the code look a bit prettier.

Issue #1400, opened 2022-10-19
M2 Nano - Override rapid movements


tatarize:
> Override rapid movements does not work correctly - multiple issues, Travel moves using drag pane are SLOOOOOWWWWW and have a twitch at the beginning. No diagonal moves. drag pane moves, some further drag pane moves do not happen.

Override rapid movement needs fully testing. It should be used to limit the jogging speed in both x and y.

tatarize:
This is accurate, fixed in #1593.

PR #1593, opened 2022-12-27
Lihuiyu: Correct Override-Speed Fix

tatarize:
* Jogs between objects ignored the settings for traveling at restricted speed.
* Correct the jog directions to combine the x and y at the speed of the slower when the slower is longer.
* If x is faster and y is slower but x is longer, does a proper zig at the respective speeds.


tatarize:
Fixes #1400

PR #1594, opened 2022-12-28
Balor Backbone Error, Threading wait/notify

tatarize:
There are four crash logs in the latest beta for Balor:

```python
MeerK40t crash log. Version: 0.8.1002 git on Windows: Python 3.9.10: AMD64 - wxPython: 4.1.1 msw (phoenix) wxWidgets 3.1.5
  File "C:\_development\meerk40t-2\meerk40t\balormk\controller.py", line 328, in connect_if_needed
    raise ConnectionError
ConnectionError

  File "C:\_development\meerk40t-2\meerk40t\gui\wxmribbon.py", line 356, in button_click
    self._button_click_id(evt_id, event=event)
  File "C:\_development\meerk40t-2\meerk40t\gui\wxmribbon.py", line 392, in _button_click_id
    button.action(event)
  File "C:\_development\meerk40t-2\meerk40t\balormk\gui\gui.py", line 115, in <lambda>
    "action": lambda v: service("red on\n"),
  File "C:\_development\meerk40t-2\meerk40t\kernel\context.py", line 33, in __call__
    return self._kernel.console(data)
  File "C:\_development\meerk40t-2\meerk40t\kernel\kernel.py", line 2097, in console
    self._console_parse(command, channel=self._console_channel)
...
  File "C:\_development\meerk40t-2\meerk40t\balormk\device.py", line 1142, in balor_on
    self.driver.connection.write_port()
  File "C:\_development\meerk40t-2\meerk40t\balormk\controller.py", line 1355, in write_port
    return self._command(WritePort, self._port_bits)
  File "C:\_development\meerk40t-2\meerk40t\balormk\controller.py", line 496, in _command
    return self.send(cmd, read=read)
  File "C:\_development\meerk40t-2\meerk40t\balormk\controller.py", line 350, in send
    self.connect_if_needed()
  File "C:\_development\meerk40t-2\meerk40t\balormk\controller.py", line 339, in connect_if_needed
    raise ConnectionRefusedError(
ConnectionRefusedError: Could not connect to the LMC controller.
```

This is triggered by hitting red-light on with the device not actually able to connect.

* To correct this bug we are switching the spooler over to a wait/notify system.
* If the spooler queue is empty then we wait.
* We try again when the spooler gets anything added to it (notify).

Within the Galvo driver:
* We also switch from returning when we invoke backbone error (really probably backend error, or we-gave-up state) to throwing the ConnectionRefusedError.

Note that the correction to the spooler code to allow ConnectionRefused to force the thread to wait, might make the entire backbone error setup moot. Since the ConnectionRefusedError stops the spooler from attempting additional connections until something else is submitted. In our case you must submit something to the queue (in our case a `connect` command) and set the backbone error to false.

This also means the background drivers will do less work and won't be alive, waiting, and running in circles. If they aren't started they will simply be stuck in a wait, until the shutdown command tells them to wake up and die.


meerk40t/balormk/controller.py
```diff
@@ -339,6 +340,7 @@ def connect_if_needed(self):
                     raise ConnectionError
                 self.init_laser()
             except (ConnectionError, ConnectionRefusedError):
+                time.sleep(0.3)
```
tatarize, line 343: We hammer the connect less hard with a tiny sleep each time we failed to connect.

meerk40t/balormk/device.py
```diff
@@ -1132,16 +1132,20 @@ def balor_goto(command, channel, _, x=None, y=None, remainder=None, **kwgs):
...
-                reply = self.driver.connection.light_off()
-                self.driver.connection.write_port()
-                self.redlight_preferred = False
-                channel("Turning off redlight.")
-            else:
-                reply = self.driver.connection.light_on()
-                self.driver.connection.write_port()
-                channel("Turning on redlight.")
-                self.redlight_preferred = True
+            try:
```
tatarize, line 1135: Specific code that directly sends will trigger the ConnectionRefusedError if you can't send that data.

meerk40t/balormk/gui/gui.py
```diff
@@ -2,10 +2,6 @@ def plugin(service, lifecycle):
     if lifecycle == "service":
         return "provider/device/balor"
     if lifecycle == "invalidate":
-        try:
-            import numpy  # pylint: disable=unused-import
-        except ImportError:
-            return True
```
tatarize, line 8: numpy isn't actually a requirement here.

tatarize:
The test timeouts implies the quitting no longer works correctly. I did change how the individual threads managed to let the spoolers die. Since they can't just recheck the pipe to see if the kernel was shutdown.

PR #1604, opened 2022-12-30
Add support for physical home

jpirnay:
a) Add physical_home command to drivers, allowing override of logical home.
b) Add tooltip and right mouse click action to home button in jog panel

### DIFF
--- a/meerk40t/balormk/driver.py
+++ b/meerk40t/balormk/driver.py
@@ -377,6 +377,12 @@ class BalorDriver:
         """
         self.move_abs("50%", "50%")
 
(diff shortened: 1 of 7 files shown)


tatarize:
We only have `origin` and `home`. The concept of physical home sort of died a bit. I guess there's a distinction but home is `physical_home` and non-physical home is `origin` which currently goes to 0,0 but could be set to travel other locations.

tatarize:
Checked. G28 is the internal origin command without using software origin. $H is physical home.

tatarize (approved):

PR #1605, opened 2022-12-30
Switch Camera button to first identified camera

jpirnay:
After 'Identify cameras' the ribbonbar button switches to the first identified camera


tatarize (approved):
Had to check that it properly got rid of the listeners but yeah, that's solid.

PR #1635, opened 2023-01-03
materialtest: allow maintaining existing elements

jpirnay:


tatarize (approved):
I'd sort of prefer Yes/No/Cancel since you're really giving 3 choices, but suddenly cancel means do this thing but with these caveats.

jpirnay:
> I'd sort of prefer Yes/No/Cancel since you're really giving 3 choices, but suddenly cancel means do this thing but with these caveats.

Cancel still means cancel, so should be fine

PR #1640, opened 2023-01-06
Some blob handling + blop property panel

jpirnay:
a) Fix involuntary deletion of blob when executing 'Remove unused' from tree
b) Add blob deletion option in tree
c) (Untested) code change to blob execution in grbl
d) Add property display panel for blobs (Hex-Codes)
grafik

### DIFF
--- a/meerk40t/core/element_treeops.py
+++ b/meerk40t/core/element_treeops.py
@@ -618,6 +618,20 @@ def init_tree(kernel):
         self.signal("operation_removed")
 
     @tree_conditional(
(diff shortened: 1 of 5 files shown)


tatarize (approved):
While it differs from blob to blob, I'm not sure hex edit is the right view. I mean that's not really how I'd edit or view a gcode. Some other formats like moshi or egv might not have much of an edit and things like ruida would work but generally be pure hex and mostly gibberish looking.

jpirnay:
It was never intended to act as an editor, I just wanted to fill the void. A pure ascii display is most likely a good alternative option.

PR #1643, opened 2023-01-07
GRBL Parser - some fixes

jpirnay:
# Blob Property Panel
Speedup + alternative view
grafik
grafik

# GRBL Parser (as in convert blob to elements)
Some fixes to GRBL parser to allow at least basic import, some fixes outstanding:
a) $J unrecognised
b) Import does not acknowledge different origin, so mirrored in Y-axis (for standard GRBL origin bottom left)
c) Import has some flaws in recognising moves?!
MK: 
grafik
LaserGRBL:
grafik

But going in the right direction and fit for further testing

### DIFF
--- a/meerk40t/grbl/parser.py
+++ b/meerk40t/grbl/parser.py
@@ -29,13 +29,7 @@ The channel callable is given any additional information about the gcode.
 import re
 
 from meerk40t.svgelements import Color
-
-MM_PER_INCH = 25.4
-MIL_PER_INCH = 1000.0
-
-MM_PER_MIL = MM_PER_INCH / MIL_PER_INCH
-INCH_PER_MIL = 1.0 / MIL_PER_INCH
-MIL_PER_MM = MIL_PER_INCH / MM_PER_INCH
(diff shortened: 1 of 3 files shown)


tatarize (approved):

meerk40t/grbl/parser.py
```diff
@@ -258,10 +253,15 @@ def parse(self, data, elements):
...
                 d = d.decode("utf-8")
-            self.process(d)
+            # Lets split lines...
+            splitted_lines = d.splitlines()
+            for singleline in splitted_lines:
+                self.process(singleline)
+        # We need to add a matrix to scale grbl coordinates ?!
         elements.elem_branch.add(
-            type="elem path", path=plotclass.path, stroke=Color("blue")
+            type="elem path", path=abs(plotclass.path), stroke=Color("blue"), stroke_width=UNITS_PER_PIXEL,
```
tatarize, line 262: This should probably use `stroke_width=context.elements.default_strokewidth`

Since that exists much as `default_stroke` and `default_fill` exist.

PR #1644, opened 2023-01-08
Update Spooler History to Direct Tracking

tatarize:
Use direct contact with logs, and reference logs more exactly for spooler history.

* Using signals requires the spooler to be open to record the entry.
* Using signals is potentially lossy if many things are cancelled.
* Adds in UID for each job, so they can be more directly referenced without ambiguity.
* Better readability in meerk40t.log


tatarize:
See:

https://github.com/meerk40t/meerk40t/pull/1540

meerk40t/gui/spoolerpanel.py
```diff
@@ -743,235 +654,64 @@ def datestr(t):
...
-                    f"{info[IDX_HISTORY_STEPS_DONE]}/{info[IDX_HISTORY_STEPS_TOTAL]}"
-                )
-                self.list_job_history.SetItem(
-                    list_id, self.column_history["steps"], stepinfo
-                )
-            self.list_job_history.SetItemData(list_id, idx - 1)
+            self.list_job_history.SetItemData(list_id, idx)
 
     def reload_history(self):
-        def fixitems():
```
jpirnay, line 871: I was able to crash meerk40t consistently with an improperly stored logging string, part of this routine is not only to get backward compatibility (which is no longer an issue) but to deal with malformed entries. So the:
```
           if isinstance(self.history, str):
                # Flawed !! At least recover gracefully
                self.history = []
```
should be reestablished

tatarize:
Yeah, a lot of this was just to get some stuff a bit more graceful than they currently felt like they were. Especially relying on signals to convey the data to then get stored. This would, in theory store that data even with CLI jogs, etc.

In theory the csv stuff would be best done as a sort of report. Like info on all jobs, or just important jobs (I didn't automatically delete them (but probably should)), all those with this flagged `info` bit.

jpirnay (dismissed):

PR #1658, opened 2023-01-14
New CH341 with challenge codes

tatarize:
* Challenge codes respond true with a 204 for about half a second after the challenge is issued. This means to read or process a device by the serial number we must perform this protocol step. If it remains 206 the challenge was failed, if it was successful then we mark the device based on the known serial number.
* The issue with this is that the core CH341 does the decision of the connection/non-connection of the different elements. This PR is an attempt to simplify the CH341 interfacing to use a simple connection rather than a series of handlers.
* Also we move the connection/non-connection into the lihuiyu controller.
* The original code is left intact for Moshi which also uses CH341. 

Due to the nature of this PR it can't be merged without being fully tested and vetted and probably can't be added until after the next full version release.


tatarize:
This might actually be close. It was actually appearing to check for serial numbers and rejecting devices based on the reply.

tatarize:
Should need full testing anyway.

PR #1660, opened 2023-01-15
Speedup load and cut

tatarize:


meerk40t/core/node/rootnode.py
```diff
@@ -111,6 +111,27 @@ def notify_modified(self, node=None, **kwargs):
             if hasattr(listen, "modified"):
                 listen.modified(node, **kwargs)
 
+    def notify_translated(self, node=None, dx=0, dy=0, **kwargs):
```
tatarize, line 114: This trick might work also work for uniform scaling. To know that the tree doesn't need new pictures.

meerk40t/gui/wxmmain.py
```diff
@@ -353,6 +353,19 @@ def register_options_and_choices(self, context):
         # ]
         # context.kernel.register_choices("preferences", choices)
         choices = [
+            {
```
tatarize, line 356: This is probably huge.

meerk40t/core/elements.py
```diff
@@ -452,6 +454,40 @@ def __init__(self, kernel, index=None, *args, **kwargs):
...
+        if key in self._timing_stack:
+            stime = self._timing_stack[key]
+            etime = time()
+            duration = etime - stime[0]
+            stime[0] = etime
+            stime[1] += duration
+            stime[2] += 1
+            if display:
+                # print (f"Duration for {key}: {duration:.2f} sec - calls: {stime[2]}, average={stime[1] / stime[2]:.2f} sec")
+                self.kernel._console_channel(
```
tatarize, line 475: This should probably have it's own channel rather than running things through console. If people want to open that channel and listen to the logging they could.

tatarize (commented):
I like the suspending and resuming channels. I'm not quite sure on `freeze_tree` as a signal. I think it might be more appropriate to suspend the tree with a direct call. And I'm a little bit confused as to what is happening there, pretty sure it's just preventing tree updates, but a bit of documentation wouldn't hurt outlining what your protocol is there. 

Also, some of the suspending and resuming seems like it pales in comparison to mini-icons. It seems like that's going to do 98% of the speed up and some of the deck chair rearranging is going to get tiny improvements beyond that.

PR #1665, opened 2023-01-16
Add more information to file/group property

jpirnay:
Show information how many elements (recursively) this node contains


jpirnay:
This one was actually helpful for me as I wanted to have an oveview of the file-size and the amount of elements

PR #1669, opened 2023-01-16
Speedup #2

jpirnay:
a) Miscellaneous performance improvements for scale operations as they no longer require a recalculation of bbox (which is still a very expensive operation for paths)
b) Multiple bugfixes around this
c) Less time consumption in attractionwidget if snap to points is disabled
d) Alignment of emphasized objects on bed (right mouse click on alignment buttons) is avoiding unnecessary startup calculations which made them effectively useless for large selections
e) Unrelated bugfix for invalid matrix during linetext creation
f) Fix some bugs / deficiencies around attraction points on elements


tatarize:
unit test failure is legit.

tatarize (changes requested):
unit test is failing, some difference exists

jpirnay:
Unittests work again

meerk40t/gui/scenewidgets/attractionwidget.py
```diff
@@ -259,7 +272,9 @@ def process_draw(self, gc):
                     if pts[2] in (TYPE_POINT, TYPE_BOUND):
                         self.draw_caret(gc, pts[0], pts[1], closeup)
                     elif pts[2] == TYPE_MIDDLE:
-                        self.draw_midpoint(gc, pts[0], pts[1], closeup)
+                        self.draw_midpoint(gc, pts[0], pts[1], closeup, False)
```
tatarize, line 275: def draw_midpoint(self, gc, x, y, closeup):

What's the False or True there for? Is that *also* closeup?

This will crash.

meerk40t/gui/scenewidgets/selectionwidget.py
```diff
@@ -498,14 +505,14 @@ def tool(self, position, nearest_snap, dx, dy, event=0, modifiers=None):
...
             # b = self.reference_rect.bbox()
-
-            for e in elements.flat(types=elem_nodes, emphasized=True):
-                try:
-                    if e.lock:
-                        continue
-                except AttributeError:
-                    pass
-                e.matrix.post_rotate(delta_angle, self.rotate_cx, self.rotate_cy)
+            with elements.undofree():
```
tatarize, line 508: Clever!

tatarize (changes requested):
Draw middle there will crash.

meerk40t/gui/scenewidgets/attractionwidget.py
jpirnay: That's an artefact from a test - and yes it will crash - fixed

jpirnay:
image

PR #1682, opened 2023-01-22
Allow saving of a plain svg file

jpirnay:


meerk40t/core/elements.py
```diff
@@ -223,30 +223,30 @@ def plugin(kernel, lifecycle=None):
                 "page": "Classification",
                 "section": "_90_Auto-Generation",
             },
-            {
```
tatarize, line 226: This rather large commenting out doesn't seem to be related or reflected in the PR title and seems like it might want to be a different PR.

meerk40t/core/svg_io.py
```diff
@@ -154,19 +154,22 @@ class SVGWriter:
     @staticmethod
     def save_types():
         yield "Scalable Vector Graphics", "svg", "image/svg+xml", "default"
+        yield "SVG-Plain (no extensions)", "svg", "image/svg+xml", "plain"
```
tatarize, line 157: I basically forgot I spent a while writing this whole save_types protocol.

tatarize (approved):

meerk40t/core/elements.py
jpirnay: Agree

PR #1687, opened 2023-01-24
Acknowledge classify_new flag on load of file

jpirnay:
A plain file was classified even if the 'Classify on creation' flag was unset
grafik
That effectively prevented to have a complete manual assignment logic


tatarize (dismissed):
The GRBL changes don't belong in this PR. There's a `control-servers` branch but it would certainly need more work than just that little change.

jpirnay:
That was fully unintended. I will revert the grbl commits

PR #1697, opened 2023-02-11
Ensure cache generation has a synced matrix between node.matrix and n…

jpirnay:
…ode.shape.transform


tatarize:
Certainly don't need a PR for this. In pycharm, menu: Git, merge. Select main. Merging into branches from main should be beyond harmless.

PR #1702, opened 2023-02-20
Add suppression of elements in disabled operation

jpirnay:
If an operation is disabled you can opt to suppress the display of contained elements in the scene:
image
This option is also available in the operation property:
image
**This option is only available if the operation has been disabled, active operations will always be displayed.**
layered_op_display


tatarize:
Is that icon a ghost? Rather than an eyeball?

meerk40t/gui/icons.py
```diff
@@ -3638,3 +3638,13 @@ def RemoveAlpha(self, image, black_bg=False):
     b"gYtSN6IY+kRjYW6sBVtK3W7AKIHermBNRQPgEsWqNfv6vJUMDVVETDVlpCdYUxEiH2XE+Z7p"
     b"jikTgSOdGv4BIclJf+4n3KEAAAAASUVORK5CYII="
 )
+
+icons8_ghost_20 = PyEmbeddedImage(
```
tatarize, line 3642: Ghost for `not-shown`. Seems like crossed-out-eyeball iconography.

tatarize (approved):

PR #1703, opened 2023-02-21
Update Laserposition in Jog-Pane

jpirnay:
image


tatarize:
I think we had partially opted against this because it had an issue that you'd lose the laser-goto location.

tatarize (approved):

PR #1711, opened 2023-02-24
Fix treatment of fill='currentColor'

jpirnay:
Fixes treatment of files like this:
bugreport
where ``fill='currentColor'`` or ``stroke='currentColor'`` are used.
Addresses: https://github.com/meerk40t/meerk40t/discussions/1657

### DIFF
--- a/meerk40t/core/svg_io.py
+++ b/meerk40t/core/svg_io.py
@@ -1164,13 +1164,16 @@ class SVGLoader:
             else:
                 width = None
                 height = None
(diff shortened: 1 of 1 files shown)


tatarize (approved):
I was worried this would be weird or broken or you'd try to mess with elements where that's precarious but the most right way to do that. Nope. Perfect. It's an option because what the color is for currentColor is out-of-context for the parsing and that's why that option exists.

PR #1717, opened 2023-02-27
Add speed/power scaling option for all ops

jpirnay:
The use case to adjust some speed / power settings globally to acknowledge a different material can be quite cumbersome to perform, especially if the laser does not have a global external power setting as found in most K40s, like for Diode Lasers.

grafik


tatarize (approved):
This works, but seems a bit like it should do more fancy scaling that can be set and reset rather than set everything at this value. It seems like this could be done with selecting every operation and setting all their power settings with the same tree_operation or whatever.

PR #1718, opened 2023-02-28
Divide Emulator and Job for GRBL and Ruida

tatarize:
A principle reason why realtime commands don't work in the 0.8.2 version is that the emulator is really two parts. One governs the control over the laser. This is the reverse of the controller. Whereas the other governs the execution of code, this is the inverse of the driver. It's possible to use GRBL-like control interfaces while sending data that is not strictly grbl (except jog which is written in grbl code but is done in a non-modal fashion. 

So even with properly different threads to write the code to the emulator as to execute it, the realtime commands cannot execute until the execution catches up. The result is that we need a realtime commands to control the laser and the programmed commands to send to the laser as a program.

### DIFF
--- a/meerk40t/core/element_treeops.py
+++ b/meerk40t/core/element_treeops.py
@@ -1021,7 +1021,8 @@ def init_tree(kernel):
     )
     def blob_execute(node, **kwargs):
         spooler_job = self.lookup(f"spoolerjob/{node.data_type}")
-        job_object = spooler_job(self.device.driver, self.device.scene_to_show_matrix())
+        matrix = self.device.scene_to_device_matrix()
+        job_object = spooler_job(self.device.driver, matrix)
         job_object.write_blob(node.data)
         self.device.spooler.send(job_object)
 
--- a/meerk40t/core/element_treeops.py
+++ b/meerk40t/core/element_treeops.py
@@ -1011,6 +1011,20 @@ def init_tree(kernel):
             d2p.parse(node.data_type, node.data, self)
         return True
 
+    @tree_conditional_try(
+        lambda node: kernel.lookup(f"spoolerjob/{node.data_type}") is not None
+    )
+    @tree_operation(
+        _("Execute Blob"),
+        node_type="blob",
+        help=_("Run the given blob on the current device"),
+    )
+    def blob_execute(node, **kwargs):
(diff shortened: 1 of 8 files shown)


meerk40t/core/element_treeops.py
```diff
@@ -972,7 +972,7 @@ def lasercode2cut(node, **kwargs):
         node.replace_node(CutCode.from_lasercode(node.commands), type="cutcode")
 
     @tree_conditional_try(
-        lambda node: kernel.lookup(f"emulator/{node.data_type}") is not None
+        lambda node: kernel.lookup(f"spoolerjob/{node.data_type}") is not None
```
tatarize, line 975: Switched to spooler-job since emulator isn't the important bit for this now.

meerk40t/core/element_treeops.py
```diff
@@ -1011,6 +1011,21 @@ def blob2path(node, **kwargs):
...
+        _("Execute Blob"),
+        node_type="blob",
+        help=_("Run the given blob on the current device"),
+    )
+    def blob_execute(node, **kwargs):
+        spooler_job = self.lookup(f"spoolerjob/{node.data_type}")
+        matrix = self.device.scene_to_device_matrix()
+        job_object = spooler_job(self.device.driver, matrix)
+        job_object.write_blob(node.data)
+        self.device.spooler.send(job_object)
```
tatarize, line 1027: With a spoolerjob, we can send that directly into the spooler with the blob.

meerk40t/core/spoolers.py
```diff
@@ -371,6 +371,71 @@ def home_dot_test():
             return "spooler", spooler
 
 
+class SpoolerJob:
```
tatarize, line 374: Moved test spooler job to actual as an example of what a spoolerjob looks like.

meerk40t/core/spoolers.py
```diff
@@ -552,15 +617,20 @@ def command(self, *job, priority=0, helper=True):
             self._lock.notify()
         self.context.signal("spooler;queue", len(self._queue))
 
-    def send(self, job):
+    def send(self, job, prevent_duplicate=False):
```
tatarize, line 620: We need to be able to put the same job into the spooler repeatedly since we'll stream as such.

PR #1725, opened 2023-03-08
Laserpanel

tatarize:


meerk40t/grbl/driver.py
```diff
@@ -650,3 +652,43 @@ def _g21_units_mm(self):
...
+        self.grbl_realtime("\x90")
+        if ifactor < 10:
+            for idx in range(10 - ifactor):
+                self.grbl_realtime("\x92")
+        elif ifactor > 10:
+            for idx in range(ifactor - 10):
+                self.grbl_realtime("\x91")
+
+    @staticmethod
+    def has_adjustable_power():
```
tatarize, line 689: Might want to not be static. Since you might need one driver to have adjustable power and another one not to.

PR #1731, opened 2023-03-13
Perspective Viewports

tatarize:
Using perspective 4->4 matrix conversions convert scene space to laser space.


tatarize:
Initial work with this assumed that the things like flip-x and flip-y wouldn't be needed but they clearly are. Even though the mapping is easier that makes making the matrix code significantly easier it does not actually make those flips unneeded.

PR #1740, opened 2023-03-18
Rotary

jpirnay:


tatarize:
Lihuiyu rotary panel shows up blank. Others too.
image

tatarize:
They were flagged as hidden. This seems in error. I removed the flag.

PR #1744, opened 2023-03-23
Signals RibbonToggle State

tatarize:
The goal of this PR is to ensure that the ribbon toggle state is compatible with the choices panel signals of the same varieties.

Currently the state is the opposite of the sent signal state. So `stop_tracing True` would tell make the trace button state toggle to unpressed. This is the opposite of what the choice panel would send for the same sort of state. So if they reference the same variable they cannot be made to agree as things are currently set.

This changes that.

* We add `toggle_attr` for triggering the setting the toggle state's attr rather than `identifier` which is typical for multi-based buttons.
* We flip the current buttons for Galvo-Trace buttons. GRBL-RedLight buttons. And we will also switch Newly AutoExecute button.

### DIFF
--- a/meerk40t/gui/wxmribbon.py
+++ b/meerk40t/gui/wxmribbon.py
@@ -403,14 +403,14 @@ class RibbonPanel(wx.Panel):
             return
         button.toggle = not button.toggle
         if button.toggle:
-            if button.identifier is not None:
-                setattr(button.object, button.identifier, True)
-                self.context.signal(button.identifier, True, button.object)
+            if button.toggle_attr is not None:
+                setattr(button.object, button.toggle_attr, True)
+                self.context.signal(button.toggle_attr, True, button.object)
             self._restore_button_aspect(button, button.state_pressed)
         else:
-            if button.identifier is not None:
-                setattr(button.object, button.identifier, False)
-                self.context.signal(button.identifier, False, button.object)
+            if button.toggle_attr is not None:
+                setattr(button.object, button.toggle_attr, False)
+                self.context.signal(button.toggle_attr, False, button.object)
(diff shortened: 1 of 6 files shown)


meerk40t/gui/wxmribbon.py
```diff
@@ -403,8 +403,14 @@ def _button_click_id(self, evt_id, event=None):
...
         button.toggle = not button.toggle
         if button.toggle:
+            if button.toggle_attr is not None:
+                setattr(button.object, button.toggle_attr, True)
+                self.context.signal(button.toggle_attr, True, button.object)
             self._restore_button_aspect(button, button.state_pressed)
         else:
+            if button.toggle_attr is not None:
+                setattr(button.object, button.toggle_attr, False)
+                self.context.signal(button.toggle_attr, False, button.object)
```
tatarize, line 413: We flip the states here. But then also send the button.object (which is needed for choice panels)

meerk40t/gui/wxmribbon.py
```diff
@@ -665,11 +671,11 @@ def _create_signal_toggler(self, button, signal):
...
             def toggle_click(origin, set_value, *args):
                 if set_value:
-                    _tb.toggle = False
-                    self._restore_button_aspect(_tb, _tb.state_unpressed)
-                else:
                     _tb.toggle = True
                     self._restore_button_aspect(_tb, _tb.state_pressed)
+                else:
+                    _tb.toggle = False
+                    self._restore_button_aspect(_tb, _tb.state_unpressed)
```
tatarize, line 678: The sent state, `set_value` is specifically applied to the toggle to set the same state as the given value.

PR #1745, opened 2023-03-23
Allow conversion of unsupported svg features

jpirnay:
Quick-parse for culprit svg features at load time with an option to convert these features via inkscape:
image

image

### DIFF
--- a/meerk40t/extra/inkscape.py
+++ b/meerk40t/extra/inkscape.py
@@ -1,9 +1,10 @@
+import gzip
 import os.path
 import platform
-from subprocess import PIPE, run
+from subprocess import PIPE, run, TimeoutExpired
 
 from meerk40t.core.exceptions import BadFileError
-
+from meerk40t.kernel.kernel import get_safe_path
 
 def plugin(kernel, lifecycle):
     if lifecycle == "register":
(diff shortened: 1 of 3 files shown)


meerk40t/extra/inkscape.py
```diff
@@ -218,3 +227,253 @@ def inkscape_base(**kwargs):
             root_context = kernel.root
             root_context.setting(str, "inkscape_path", "inkscape.exe")
             return "inkscape", (root_context.inkscape_path, None)
+
+
+def register_inkscape_conversion(kernel):
+
+    import wx
```
tatarize, line 234: Why is this a wx required extension? 

Get translation and the dialog don't seem like they would be at all required for this to work.

tatarize (approved):
I don't hate it. But, I feel like this would maybe need be to be revisited later since this should be possible to do without needing the wxPython gui system at all. So first time somebody wants to use a load-type preprocessor, it would need to exclude the wxPython stuff and add that as a seemingly optional later bit to this stuff. You're just using it for a dialog, which seems like that might better be built.

For example, you might be able to pre-process other things externally. Process an image with imagemagick, svgnest to nest the various svg shapes, and all these things are doing is calling an external program with an input and output file (likely stored in the safe directory since otherwise I you can get permission errors) and processing them that way.

tiger12506:
Now that you mention it, it seems very much like a feature that should not require the GUI, and would probably be even more useful for someone running MeerK40t headless. To have it automatically run problematic svg features through inkscape and ensure more proper laser output when you can't see the simulation would be highly desirable for those running that way. (I don't know how many people still do though)

PR #1750, opened 2023-03-29
Expand Devices with Priority

tatarize:


meerk40t/lihuiyu/device.py
```diff
@@ -21,6 +21,7 @@ class LihuiyuDevice(Service, ViewPort):
     """
     LihuiyuDevice is driver for the M2 Nano and other classes of Lihuiyu boards.
     """
+    priority = 1
```
tatarize, line 24: device.priority is increased at the class level. This value can be checked and viewed prior to initializing the class.

tatarize:
The friendly name might be better to merely add to the class specification. Like I did for the priority there. Such that if there is a friendly name it could be checked on the device class description without activation of the object.

tatarize:
never mind, I think tiger's right about the many to 1.

tatarize:
nevermind that nevermind. I think friendly name probably needs its own registration and to point to the device-driver name and not register under friendly name there and shouldn't expand the kernel to do that.

PR #1751, opened 2023-04-01
More sophisticated device selection dialog during creation

jpirnay:
image


tatarize (approved):
Bravo! This is clearly needed. The Newly driver works for example, but is entirely spammy in the device selection.

PR #1763, opened 2023-04-05
Parse .ezd files.

tatarize:
* Parse some `.ezd` files.
https://github.com/shark92651/LaserParamsConverter/issues/9


tatarize:
Only still remaining major issue is that bitarray is not a standard library and is not trivially supported.

PR #1775, opened 2023-04-06
Fix Linetext tool if no/an invalid font was set

jpirnay:


jpirnay:
This still has artifacts of the recent gitfu in it, so will be closed and reissued

PR #1776, opened 2023-04-07
Fix and use check_for_updates

jpirnay:
grafik

grafik

```
[16:49:03] help check_for_updates 
[16:49:03]     	This command checks for updates and outputs the results to the console.  If check_for_betas setting is true OR we run a beta then we get the release list and iterate through to find the latest full and beta releases that are > currently running  If check_for_betas is false and we are not currently running a beta then we check only for a full release > currently running
[16:49:03]     
[16:49:03]     
[16:49:03]     	check_for_updates 
[16:49:03]     	(None) -> check_for_updates -> (None)
[16:49:03]     	Option: bool ('--beta', '-b'):
[16:49:03]     		Check for betas
[16:49:03]     	Option: bool ('--popup', '-p'):
```
The ``popup`` option will ask if you want open the corresponding webpage is an update is found.


tatarize (approved):

tatarize:
Check for update was mostly @Sophist-UK's code, I don't think I messed with it nearly at all.

PR #1785, opened 2023-04-09
Placements

tatarize:
* Add in `place point` nodes.
* Add in cutplan code to use the different placements.

image


meerk40t/core/svg_io.py
```diff
@@ -1086,6 +1086,13 @@ def is_child(candidate, parent_node):
                     return
                 node_type = f"op {op_type.lower()}"
                 element.values["attributes"]["type"] = node_type
+            if node_type == "place":
```
tatarize, line 1089: This fallback code doesn't seem like it could ever happen. We never had a node type called "place". This does this because in 0.7.x the node type was "operation" and rather than `op cut` etc and that stuff needed to be converted over in case you were loading a very old file.

PR #1816, opened 2023-04-21
German Update

tatarize:
* German Update using poboy.

Since the German language stuff is @jpirnay's domain he probably needs to sign off on this, or at least view the few things that were found and not yet translated and do his own update.


jpirnay:
This has some holes compared to the manual version I put forward. The reason being (fully my own doing) that there are some routines in the code that construct a string to be translated (most notable exmaple: the page/section/subsection routines in choicepropertypanel take some leading indicators that allow sorting regardless of any to be appilied translations. So we might need to parse some comments with poboy directives?!

jpirnay:
Additionally translations for addons (like barcode) get lost too. So another thing to consider how to solve it.

tatarize:
Hm. Well, I'm thankful to at least get poboy working again. There are a series of issues with translations. The most pressing of which is that we cannot triage. You need a lot of understanding of the code to know how useful or unuseful a section of code would be. And we likely need to know this information and convey it to a would-be translator. So if we can get somebody to do 500 of the most used bits of text that will actually make a reasonable translation without doing the full 3000 or so.

In theory we might need to flag some translations with specialty comments, or figure out methods of translation that would work for plugins. 

There's also some very long bits that got put in `meerk40t/gui/help_assets/help_assets.py` which can basically just set a global in that file and write exceedingly long bits of text and translate them with `asset(context, "wordlist_howto")` or whatever.

tatarize:
Great feedback. Won't merge will work on 1818 issues.

PR #1817, opened 2023-04-21
Fix plan update crash in Laserpanel

jpirnay:


tatarize (commented):
More changes than I would have expected. Though it looks more professionally done with a more proper panel

PR #1824, opened 2023-04-22
Use existing pane for size in transformation

jpirnay:
grafik

Remove SizerPanel in Navigation and use another (reduced) instance of PositionPanel instead


tatarize (approved):

tatarize:
It's better to reuse things that work, and deprecate things that don't.

PR #1852, opened 2023-04-28
Fix 'simplify' by bypassing faulty code for now

jpirnay:


meerk40t/core/elements/elements.py
```diff
@@ -3198,7 +3198,13 @@ def list_subpath_bounds(obj):
                 result.sort(key=lambda a: a[2])
                 return result
 
+            # TODO: This is not working properly yet, so we skip this for now
```
tatarize, line 3201: todo.

tatarize (changes requested):

meerk40t/core/elements/elements.py
jpirnay: Fine I will comment the complete section out as it is not working properly in cases of subpaths yet

jpirnay (commented):
Removed comment and no longer call the culprit function

PR #1868, opened 2023-05-02
Add repeated button press logic to Navigationpanels

jpirnay:


tatarize (approved):
Works fine for the M3 Nano (WASD keys don't for it). I would have expected TimerButton to be added as one of the classes in wxutils

PR #1888, opened 2023-05-12
Geomstr shapes: 2nd attempt

tatarize:
Rather than actually swap fully over and break everything. We create scaffolding to create the nodes from default attributes.


tatarize:
Even with maximum safety this actually still breaks a lot of things and should be made 0.9.x, since proper rewriting of the sections will take time.

PR #1903, opened 2023-05-17
Codecy legacy8 patch

tatarize:
Patch Legacy8 Codecy Checks


tiger12506 (approved):
Looks ok. Tested as running in Windows.

PR #1907, opened 2023-05-18
TCP Galvo Driver Connection

tatarize:
* This likely requires `size, checksum, <payload>` chunks.
* Would require that the galvo driver not necessarily expect a reply that stalls the driver.
* This likely requires a protocol of send and response which also provides the original query command.
* The data would perhaps be better off sent according to the driver protocols. So "function()", "payload" information.


tatarize:
This could work, but it would likely require a sort of `galvoplotter` like sending mechanics. With a full protocol.

PR #1910, opened 2023-05-19
Mac OSX: Tree too large vs. Laser Pane

tatarize:
Default start on Mac fails


joerlane (approved):
I see a change in minwd and MinSize. The PR has been tested as working on macOS. Approved.

PR #1913, opened 2023-05-19
Replace "shapes" command-context with "geometry"

tatarize:
`"shapes"` is/was a list of svgelements.Shape objects. `"geometry"` is purely a geometry of a single `geomstr` object. 

Shapes is only used in a few places:
* Balor driver uses it for the ElementLightJob
* Balor uses it to creates hull shapes.
* Balor uses it to create ant-points.
* elements/shapes has a conversion of shape objects to path-based shape objects.
* elements/shapes has is also a `path` command that converts elements -> shapes
* elements/trace, `trace` command accepts shapes input.
*  elements/trace, `tracegen` command accepts shapes input.
---

Requires retesting, balor looping and tracing functions.

### DIFF
--- a/meerk40t/balormk/device.py
+++ b/meerk40t/balormk/device.py
@@ -1684,7 +1684,7 @@ class BalorDevice(Service, ViewPort):
             channel(_("Element bounds: {bounds}").format(bounds=str(bounds)))
             geometry = Geomstr.rect(xmin, ymin, xmax - xmin, ymin - ymax)
             if count > 1:
-                geometry.duplicate(count)
+                geometry.copies(count)
             return "geometry", geometry
 
         @self.console_command(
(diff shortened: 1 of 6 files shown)


meerk40t/balormk/device.py
```diff
@@ -1772,26 +1759,24 @@ def ant_points(points, steps):
...
-                    path = e.as_path()
+                    path = e.as_geometry()
                 except AttributeError:
                     continue
-                for i in range(0, quantization + 1):
-                    x, y = path.point(i / float(quantization))
-                    points.append((x, y))
-                points_list.append(list(ant_points(points, int(quantization / 2))))
-            return "shapes", [Polygon(*p) for p in points_list]
+                ants = list(ant_points(path.as_interpolated_points(interpolate=quantization), int(quantization/2)))
```
tatarize, line 1776: This is not 1:1 with the old code. We may not be interpolating parts of different lines whereas previously we may have interpolated long lines.

PR #1970, opened 2023-06-18
Ezcad3 Driver

tatarize:
Initial tests of establishing an ezcad3 dlc-controller driver.

The goal here is to implement the major features that are more or less parallel to those in the ezcad2 lmc-controller. A lot of the controller is weirdly overbuilt and the function set is apparently bordering on being Turing complete.

* This PR is more a feasibility study than an honest attempt.
* We have nobody with the given hardware controller to test against.


tatarize:
This would absolutely require hardware to complete.

PR #1972, opened 2023-06-19
Restore the Win build architecture: x86 for libusb

tiger12506:
Tested. This fixes the libusb issue.

Recommendations:
Don't fix stuff that ain't broke.
Python architecture should show in About -> System Information? Is this possible?


tatarize:
If we're switching back to 32 bit, we need to properly recompile the 32bit bootloader.
        python3 ./waf distclean all --target-arch=32bit

PR #1982, opened 2023-07-14
Cosmetics for Ribbonbar

jpirnay:
This is addressing some of the inconsistencies of buttons on the new ribbonbar:
a) Tab-Width takes real text width into consideration
b) Button heights will be all the same in one page across multiple panels
c) Wrong iconsize for Ref Object button
d) Wrong tooltip if hovering over disabled buttons

Before:
image

image

image

After:
image
image


tatarize:
Entirely implemented in #1983 (except for the all same height routine since that was mooted by the changes in the PR natively)

Issue #1943, opened 2023-06-02
[0.8.5] Suggestion: Show power as %, DPI as integer to avoid visual confusion 


Sophist-UK:
![image](https://github.com/meerk40t/meerk40t/assets/3001893/11dad183-1022-4396-a5d0-2ced1a1b92d0)

I have added "dpi" to my labels to assist with this, but I think power as a % makes more sense across various controllers and might make it more intuitive for Lightburn users.

PR #2008, opened 2023-07-24
Device specific power display in tree

jpirnay:
Allow a per device setting whether power should be displayed as ppi (default) or as percentage.
Addresses [Issue #1943](https://github.com/meerk40t/meerk40t/issues/1943)

power_setting


tatarize:
I'm not opposed but this seems like it could be confusing. Since the % would be only in the tree. But, wouldn't be entered into properties or really used there.

tatarize (approved):

jpirnay:
> I'm not opposed but this seems like it could be confusing. Since the % would be only in the tree. But, wouldn't be entered into properties or really used there.

I am working on this

PR #2045, opened 2023-08-14
Bug fixes 0.8.x

tatarize:
(cherry picked from commit 26e0ec83a40cc4219eca75376496f4d6c36856e5)


meerk40t/core/svg_io.py
```diff
@@ -1049,8 +1050,8 @@ def is_child(candidate, parent_node):
                         lock=_lock,
                     )
                     e_list.append(node)
-            except OSError:
-                pass
+            except OSError as e:
+                raise e
```
Sophist-UK, line 1054: Isn't this the same as not doing an except?

tatarize: Yes. This like was debug. I needed to see what e actually said as part of the debugging.

Sophist-UK: No - my point is that trapping and then raising the same exception is EXACTLY the same as not trapping it at all.

tatarize: I did that to look at the current value of `e` which let me see what the thrown error was. Which is slightly different. But, yes, you are correct. I will remove that change.

tatarize: This and all changes to svg_io were reverted.

PR #2049, opened 2023-08-16
Fix inkscape 1.3 image load bug

jpirnay:


tatarize:
This is probably worth fixing correctly for all svgelement users.

tatarize:
Fixed via https://github.com/meerk40t/meerk40t/commit/9e75875a7c5c1033c3ef55e73dfa6c129203d204
(The bug is in svgelements, needed to be fixed in svgelements)

Co-organized with meerk40t/svgelement#240

PR #2052, opened 2023-08-18
Add in PPI-Phase for lihuiyu devices.

tatarize:


joerlane (approved):
Reviewed source changes, and tested features. Extremely low PPI values, looped, can now hit every spot along a path. Tested PPI as low as 12, with a few loops, and could get a solid line instead of dots spaced apart. Should even work fine for PPI values much below 12. Very nice.

Issue #2058, opened 2023-08-21
[Bug]: Operation stops / Systematic USB errors


zufus:
### Summary Description

When executing cut operations I get a deterministic "USB Error" which causes the board to disconnect and stops the job. I tried to change cable and PC. Disconnection always occurs at the same point during the JOB. So I exclude cable issues. The same drawing seems to run smoothly when using k40whisperer. However, I would prefer using meerk40t.

Another issue I found is that when I launch the job the laser head stars moving/cutting: the main screen background becomes green and I see a cursor moving along the drawing path: however, the cursor on the screen runs much faster than the laser head.

Depending on the drawings, the job is completed, but it seems that on the PC it is executed much faster: result is that meerk40t sees the job as ended, while the job is still running. In this case, the Abort button is not functional anymore..

it seems to me that meerk40t sends commands at a rate which is faster then the rate that can be handled by the machine.

My laser machine has a M2 nano board. I don't see useful error messages on the USB log trace.

can you please guide me to create a meaningful bug-report? 

The same behavior occurs either with 0.8 and the latest 0.9.0009 git version.

I am using it on a Debian GNU/Linux machine. Other details:

Python 3.11.4
wxPython gtk3 wxWidgets 3.2.2

Thanks!

Marco

### MeerK40t Version

0.9.0009

### MeerK40t Type

Git

### Your Operating System

Linux

tatarize:
> When executing cut operations I get a deterministic "USB Error" which causes the board to disconnect and stops the job. I tried to change cable and PC. Disconnection always occurs at the same point during the JOB. So I exclude cable issues. The same drawing seems to run smoothly when using k40whisperer. However, I would prefer using meerk40t.

1. This is extremely unusual. If the file that does it is consistent can I get a copy? Either here or tatarize at my gmail account. There is nothing as part of cut operations that should cause a USB error. 


> Another issue I found is that when I launch the job the laser head stars moving/cutting: the main screen background becomes green and I see a cursor moving along the drawing path: however, the cursor on the screen runs much faster than the laser head.

2. It's generally the write position rather than the actual location of the physical device. There's a bit of a buffer between the time it starts writing and the time position of the laser-head.

> result is that meerk40t sees the job as ended, while the job is still running. In this case, the Abort button is not functional anymore..

[...]

PR #2066, opened 2023-08-26
0.8.x Bug Fixes Addendum

tatarize:
Correct some issues outlined in:
#2058


tatarize:
This includes the version bump and is pretty much effectively 0.8.11 but this will be held until issue #2078 is dealt with.

Sophist-UK:
You can disable the crash logging and publish 0.8.11 and then do 0.8.12 with the crashlogs enabled. Or leave them disabled and enable them again in 0.9.x on the basis that soon you probably won't be bothered about crashes in 0.8.x (which are pretty rare now anyway). (Aside from the crash I had yesterday on 0.8.9 I cannot remember the last time I had a crash in 0.8.x.)

PR #2071, opened 2023-08-27
Add mm/min as speed display unit

jpirnay:
speed_options


jpirnay:
The number of files changed are due to the translation changes, all po and mo got touched

Issue #2072, opened 2023-08-27
[0.8.9] Weird extra job spooled


Sophist-UK:
I just hit the BGB and got this in the spooler:
![image](https://github.com/meerk40t/meerk40t/assets/3001893/40a14175-4472-4e77-a2ad-f8638a06f087)
I have no idea what the second job is.

tatarize:
I assume BGB is big go button... Yeah, that's rightly weird.

Sophist-UK:
Big Green Button.

I may have double-clicked it. The second job was a duplicate. So the primary issue is that double-click = double job. Secondary issue is why the steps were wrong.

tatarize:
That happens. I think it's actually because you may have double clicked a thing and or triggered it inadvertently that it's sketchy with double clicking. It may actually just be there to prevent you from double clicking like that.

tatarize:
I think this relates to #1870 to only allow conscious clicks. Still spools it but disables that because your click was invalid or something.

tatarize:
Nope. Didn't pan out. Maybe `plan clear` is breaking the LaserJob already generated?

Yeah, you can click the BGB and get it to happen. So it's quite minor, and easy to reproduce.

jpirnay:
Double Click should be suppressed now.
The steps will only be updates, when the job starts.

PR #2074, opened 2023-08-27
Suppress unwanted double clicks in Big Run Button

jpirnay:


tatarize:
No joy. Might be fine but doesn't fix issue #2072

tatarize:
Or never mind... they look weird but work fine.

image

tatarize (approved):

PR #2082, opened 2023-08-31
Prevent endless loop if scene is rotated

jpirnay:
Ugly fix: no guides will be displayed if scene is rotated. Could need some distinction for 90 / 180 / 270 deg.


tatarize:
Really I would opt for a much more robust and correct solution.

It's usually not worth it but, often times it can pay some interesting dividends when you do all the math correctly. Though in this case it's a little strange because moving when the viewport is tilted means that we'd actually change both x and y if purely an ortholgonal view. There's only a very cryptic command that would rotate that view.  And there's really every reason we'd have guides if and only if the view angle is 0.

It seems like solving this stuff would be doable, though at 90 degrees, it would depend on what the bars actually mean. If they are actual distance values that different than if they are distances in x direction (at which point they wouldn't be defined).

PR #2083, opened 2023-08-31
Device Views

tatarize:
Cursory test for Device Views.

The goal of this is to replace the Device as a subclass of ViewPort with device.view objects. These are the same view class as is used for `context.space.display`.

* Expectation that all Devices have a `.view` attribute.
* Views are calculated by applying operations on the bounding points of a class, this serves to simplify the methods.
* Rather than a complex `scene_to_device_transform` code we use a much simpler set of underlying operations.


tatarize:
# rotary_active = self.rotary_active,
        # rotary_scale_x = self.rotary_scale_x,
        # rotary_scale_y = self.rotary_scale_y,
        # rotary_flip_x = self.rotary_flip_x,
        # rotary_flip_y = self.rotary_flip_y,
        
        Rotary elements used to be toggled on viewports, currently these are now unused.

PR #2089, opened 2023-09-05
Basic Operations Panel / Default Operations

jpirnay:
- Attempt to provide a simpler interface to operations, similar to the layer concepts of other popular laser software.
grafik

basic_ops
BTW a right click on the icons in the operation panel opens the usual context menu for operations as in the tree:
grafik

- the statusbar icons for existing operations have disappeared as this funcionality is  now available from the operations pane 
- a new set of icons have been introduced in the statusbar with a list of default operations to choose from
grafik

- this panel has been combined with the regular color / line adjustment panel
- the context menu for the element branch contains menu entries to add hatch effects 
grafik

- a somewhat helpful tooltip for effect hatch has been added
grafik

- the property dialog for a hatch effect contains now a preview and allows to change the hatch type
grafik

The default operations are editable, they use the special material list name ``_default``. If this does not exist then some useful (?) default operations are created and stored under this name. If you load this list and save it again under this slot, these operations become then the new default (program restart needed)
grafik

The Id of these operations is used for the text in the icons and is hence editable if required, as are colors etc. The label will be used for the tooltip.

Functionalitywise a click on one of the default operations in the statusbar will assign the selected elements to:

- an existing operation with the same id as the one used (ie the label shown)
- a newly created operation with the same parameters as the default operation

So you can click and use/create a non-existing operation, which can then be amended and will remain unchanged by any future click on the same icon.


meerk40t/core/node/effect_hatch.py
```diff
@@ -5,7 +5,7 @@
 from meerk40t.core.node.node import Node
 from meerk40t.core.units import Angle, Length
 from meerk40t.svgelements import Color, Matrix
-from meerk40t.tools.geomstr import Geomstr, Scanbeam
+from meerk40t.tools.geomstr import Geomstr  # ,  Scanbeam
```
tatarize, line 8: You can just remove unused imports. No need for comments.

tatarize (approved):

Issue #2091, opened 2023-09-09
Rubber Stamp Mode


tatarize:
It may be reasonable to create an effect that primarily runs within image operation. This would do the primary effects needed for stamps:

1. Image thresholding. All image goes to either black or white.
2. Image flip. We must flip horizontally so the stamp comes out correctly.
3. Image invert. The black sections are marked to not contain inked sections.
4. Image ramping. We create ramping from a certain distance away from the black sections. Largely this would be like a blur with white sections mixing with black, but no black mixing with the white sections. This allows a gentle slope of stamp rather than a steep cliff.

In theory this should be done as an extension of effect image operations, running sort of as a parallel to image-wizard as such.

Sophist-UK:
I would certainly be interested in this.

I would also be interested in a similar mode for carving other materials whereby the layers are generated using increasing (or decreasing) shades of grey. So each layer would be black or white based on a grey threshold, and the threshold would increase on higher layers e.g. on a 4 layer carve, 1st layer would be light grey or darker, 2nd would be mid-grey or darker, 3rd would be dark grey/black and the 4th would be black.

Sophist-UK:
P.S. This would need to apply to Raster ops as well as image ops. (For example, if I want to create a stamp with lettering on it.

The UI would need to be carefully considered. The scene would need to give an accurate representation of what will be burned.

I am not sure it would be parallel to image-wizard - partly because it would need to be available to Raster Ops too. For image Ops, I am not sure whether it would be a replacement for Image-Wizard or run in parallel with it.

tatarize:
In theory we could also, nominally render this data, if that was needed. Basically if you have just an image or set of images we process them according to these rules. If you have vector objects in there, just render them.

The idea generally for how to have `effect hatch` and `op hatch` sort of agree with each other is to allow engrave to have a `effect hatch` node within it. We'd then be able to deprecate `op hatch` if we merely allowed hatch effect. Currently if you type `effect-hatch` it adds to the elements an `effect hatch` node. Effect nodes basically rely on the 0.9.x change that all element nodes contain an `as_geometry()` or `as_image()` function which provides a geomstr object or an image object. The `effect hatch` node is an element which accepts as operands closed shapes and will fill them with it's hatchy goodness. This happens in the regular scene and does not require a special operation to do this.

[...]

PR #2094, opened 2023-09-10
Hatch Multi-Effect

tatarize:
* Intending to lay the groundwork for #2091
* `op hatch` is totally deleted.
* `op engrave` pre-apply a sibling references if they contain an effect node (a real effect node, non-referenced).
* Loading of older `op hatch` objects replace these with an `op engrave` containing a hatch-effect node.

---

- [ ]  Test material test with hatch operation.
- [x]  Fix loading and saving of `effect *` being inside an operation. 
- [x]  Fix persistence of `effect *` between loading and saving.
- [x]  Do not allow dropping of node into hatch-effect if the hatch-effect is inside an operation.


meerk40t/extra/ezd.py
```diff
@@ -1096,18 +1096,15 @@ def parse(self, ez, element, elem, op):
                 # (self, ez, element, elem, op)
                 self.parse(ez, child, elem, op)
         elif isinstance(element, EZHatch):
-            elem = elem.add(type="group", label=element.label)
```
tatarize, line 1099: This change requires testing.

meerk40t/gui/materialtest.py
```diff
@@ -187,7 +187,9 @@ def size_it(ctrl, value):
         self.color_scheme_free.append(False)
         self.default_op.append(ImageOpNode())
         self.color_scheme_free.append(True)
-        self.default_op.append(HatchOpNode())
+        op = EngraveOpNode()
```
tatarize, line 190: This change requires testing.

tatarize: Does not seem to work, the operations were empty.

PR #2101, opened 2023-09-12
Some fixes / improvements for default / basic ops

jpirnay:
a) Centralize code to load/save/generate default operations to elements.py
b) Allow to have device dependent default operations (save them under ``_default_``_devicelabel_) (so if your current device is called 'balor-device' then use ``_default_balor-device``)
c) a fix for the derivable function in the settings class to make the above work

### DIFF
--- a/meerk40t/gui/statusbarwidgets/defaultoperations.py
+++ b/meerk40t/gui/statusbarwidgets/defaultoperations.py
@@ -41,110 +41,6 @@ class DefaultOperationWidget(StatusBarWidget):
             slabel = ""
         return slabel
 
-    def init_nodes(self):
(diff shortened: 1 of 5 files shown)


meerk40t/gui/statusbarwidgets/defaultoperations.py
```diff
@@ -175,7 +71,7 @@ def size_it(ctrl, dimen_x, dimen_y):
         self.first_to_show = 0
         self.page_size = int((self.width - 2 * self.buttonsize_x) / self.buttonsize_x)
 
-        self.init_nodes()
+        self.context.elements.init_default_operations_nodes()
```
tatarize, line 74: This should probably get called by elements itself.

PR #2105, opened 2023-09-13
Adjust node assignment logic for default operations

jpirnay:
a) always create a copy of a default node
b) adjust the copy logic of node.py to have a full set of attributes

### DIFF
--- a/meerk40t/core/elements/elements.py
+++ b/meerk40t/core/elements/elements.py
@@ -112,6 +112,10 @@ def plugin(kernel, lifecycle=None):
         kernel.register("format/elem polyline", "{element_type} {desc} {stroke}")
         kernel.register("format/elem rect", "{element_type} {desc} {stroke}")
         kernel.register("format/elem text", "{element_type} {desc} {text}")
+        kernel.register(
+            "format/effect hatch",
+            "{effect}{element_type} - {distance} {angle} ({children})",
+        )
         kernel.register("format/reference", "*{reference}")
         kernel.register(
             "format/group", "{element_type} {desc} ({children} children, {total} total)"
@@ -309,6 +313,18 @@ def plugin(kernel, lifecycle=None):
                 )
                 + "\n"
                 + _("by disabling this option."),
+                "page": "Classification",
+                "section": "_30_GUI-Behaviour",
+            },
+            {
+                "attr": "remove_non_used_default_ops",
+                "object": elements,
+                "default": True,
+                "type": bool,
+                "label": _("Remove unused default operations"),
+                "tip": _(
+                    "If a default operation is no longer used it will be removed from the list of active operations"
+                ),
                 "page": "Classification",
                 "section": "_30_GUI-Behaviour",
             },
@@ -693,7 +709,7 @@ class Elemental(Service):
         ly = float(lly)
         return lx * ly
 
-    ### Operation tools
+    # ---- Operation tools
 
     def assign_operation(
         self,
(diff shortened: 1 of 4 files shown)


jpirnay:
Still has some issues. Needs further refinement

jpirnay:
After the fix of a stupid coding error, things work as expected

meerk40t/core/elements/elements.py
```diff
@@ -312,6 +316,18 @@ def plugin(kernel, lifecycle=None):
...
+                "object": elements,
+                "default": True,
+                "type": bool,
+                "label": _("Remove unused default operations"),
+                "tip": _(
+                    "If a default operation is no longer used it will be removed from the list of active operations"
+                ),
+                "page": "Classification",
+                "section": "_30_GUI-Behaviour",
+            },
```
tatarize, line 330: While this isn't bad. I suggestion there was generally to not-show them. You are making a list of operations that you care about. You can just filter out the empty operations you don't care about.

meerk40t/core/elements/elements.py
```diff
@@ -1345,6 +1360,110 @@ def create_image(oplist):
         if needs_signal:
             self.signal("default_operations")
 
+    def create_usable_copy(self, sourceop):
+        from copy import copy
```
tatarize, line 1364: Copy can go in the top, it's standard for python

meerk40t/core/elements/elements.py
```diff
@@ -1345,6 +1360,110 @@ def create_image(oplist):
...
+        id_existing = True
+        while id_existing:
+            id_existing = False
+            id_candidate += 1
+            op_id = f"{pattern}{id_candidate}"
+            for op in list(self.ops()):
+                if op_id == op.id:
+                    id_existing = True
+                    break
+        targetop.id = op_id
```
tatarize, line 1386: Slight observation, not really actionable. But, `elements.validate_ids()` exists. It's typically though the use might not include default ops though `targetops` there might. There's some overlap there. `validate_ids" typically does `meerk40t:{idx}` in those cases when the id is missing or is duplicated and thus must be cleared. This is typically for saving an svg file and allowing valid references. But, there's not really much of an issue combining these schemes and calling the same code line when we're saving an SVG as we use other places to ensure our nodes have IDs when they are used for something.

meerk40t/core/elements/elements.py
```diff
@@ -1345,6 +1360,110 @@ def create_image(oplist):
...
+        emphasize_mode = False
+        if data is None:
+            emphasize_mode = True
+            data = list(self.flat(emphasized=True))
+        if len(data) == 0:
+            return
+        emph_data = [e for e in data]
+        op_id = targetop.id
+        if op_id is None:
+            # WTF, that should not be the case
```
tatarize, line 1398: If a thing shouldn't happen. Crashing may be correct, then fixing the root cause. I'm guessing since the default ops were being moved directly into tree, the modifications of the tree could clear out the ID and cause issues. If I'm reading the PR correctly.

meerk40t/core/elements/elements.py
jpirnay: I have left the option, but set the default to False and make it visible only in expert mode.
Operation panel will now allow to filter unused ops (default behaviour)

grafik


grafik

meerk40t/core/elements/elements.py
jpirnay: Extended and used the standard method

meerk40t/core/elements/elements.py
jpirnay: As I have changed a couple of things to address the root causes (wrong undo, direct attachment of a default operation to the tree instead of creating and using a copy), that should really never happen. But it does not hurt either...

meerk40t/core/elements/elements.py
jpirnay: Done

PR #2108, opened 2023-09-16
Improve/Standardize Op Dictionary Copying

tatarize:
Several operations have .attr values like `id` and `stopop` and these would be lost since they aren't stored in the .settings dictionary. This is not a problem with Elem nodes, where they copy the node dictionary correctly.

Here we standardize the op dictionary copying for operations.

### DIFF
--- a/meerk40t/core/elements/elements.py
+++ b/meerk40t/core/elements/elements.py
@@ -1190,7 +1190,11 @@ class Elemental(Service):
         op_tree = dict()
         for section in list(settings.derivable(name)):
             op_type = settings.read_persistent(str, section, "type")
-            op = Node().create(type=op_type)
+            try:
+                op = Node().create(type=op_type)
+            except ValueError:
+                # Attempted to create a non-boostrapped node type.
+                continue
             op.load(settings, section)
             op_tree[section] = op
         op_list = list()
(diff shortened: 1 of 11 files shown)


meerk40t/core/elements/elements.py
```diff
@@ -1190,13 +1190,7 @@ def load_persistent_op_list(self, name):
...
         for section in list(settings.derivable(name)):
             op_type = settings.read_persistent(str, section, "type")
-            # That should not happen, but it happens nonetheless...
-            # So recover gracefully
-            try:
-                op = Node().create(type=op_type)
-            except (AttributeError, RuntimeError, ValueError) as err:
-                print(f"That should not happen, but ops contained: '{op_type}' [{err}]")
-                continue
+            op = Node().create(type=op_type)
```
tatarize, line 1193: This needs to check with the bootstrapping to make sure the type is a real node and that it correctly read.

meerk40t/core/node/node.py
```diff
@@ -437,8 +437,10 @@ def _build_copy_nodes(self, links=None):
                 "stroke_width",
                 "stroke_scaled",
             ):
-                if hasattr(c, attr):
-                    setattr(node_copy, attr, getattr(c, attr))
+                if getattr(node_copy, attr, None) != getattr(c, attr, None):
```
tatarize, line 440: This debug code should eventually be removed.

PR #2114, opened 2023-09-21
Parametric Editing

jpirnay:
Introduction pf parametric editing, ie post-creation GUI updates of complex graphics. Routines (as well as basic nodes) can store 'parameter' information, that can be used to recrete the shape with different parameters.
This requires two basic routines: a) the initial creation of a node and b) a routine to look at the parameters in the node and to recreate the shape according to this set.

param_basic

Examples of such routines can be found in ``/meerk40t/extra/param_functions.py``:
a) a simple fractal tree drawing:
param_fractal

b) a grid routine to cerate a grid copy of a given node.
param_grid

The parameters will be saved and loaded in a SVG file.


meerk40t/core/elements/elements.py
```diff
@@ -6,6 +6,7 @@
 
 
 import contextlib
+import math
```
tatarize, line 9: pointless

meerk40t/core/node/elem_ellipse.py
```diff
@@ -66,6 +66,30 @@ def __init__(self, **kwargs):
             self.ry = 0
         if self.matrix is None:
             self.matrix = Matrix()
+        if self.rx == self.ry:
```
tatarize, line 69: Internal... very nice.

meerk40t/core/node/node.py
```diff
@@ -92,7 +92,7 @@ def __init__(self, *args, **kwargs):
         self._can_update = True
         self._can_remove = True
         self._is_visible = True
-
+        self.mkparam = None
```
tatarize, line 95: This seems a little extreme. mkparam for *all* nodes. Maybe a mixin like `Stroked` or something. Most nodes would never utilize this.

jpirnay: Agree, that makes sense, although you could extend the concept if you really wanted to images too, but operations do not make a lot of sense

meerk40t/core/elements/elements.py
jpirnay: Came from a previous version, before I internalised the ellipse code

meerk40t/core/node/mixins.py
```diff
@@ -3,6 +3,9 @@
 
 
 class Stroked(ABC):
+    def __init__(self, *args, **kwargs):
+        self.mkparam = None
```
tatarize, line 7: Better, but maybe you could have a non-stroked element node with functional parameters. This isn't really a requirement of stroked node.

meerk40t/core/parameters.py
```diff
@@ -66,7 +66,7 @@
 
 COLOR_PARAMETERS = ("color", "line_color")
 
-LIST_PARAMETERS = ("allowed_attributes",)
+LIST_PARAMETERS = ("allowed_attributes", "mkparam")
```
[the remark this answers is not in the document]
tatarize: Also, mostly they seem to affect elem nodes which are notably not sub-elements of parameters.

meerk40t/core/svg_io.py
```diff
@@ -683,10 +684,18 @@ def process(self, svg, pathname):
     def check_for_mk_path_attributes(self, node, element):
         for prop in element.values:
             lc = element.values.get(prop)
-            # print (f"Property: {prop} = {lc}")
             if prop.startswith("mk"):
+                # print (f"Property: {prop} = [{type(lc).__name__}] {lc}")
                 if lc is not None:
                     setattr(node, prop, lc)
+                    # This needs to be done as some node types are not based on Parameters
+                    # and hence would not convert the stringified tuple
```
tatarize, line 692: Arent the functional parameters inherent to the node types? Why does it need to be done at all?

meerk40t/gui/scenewidgets/selectionwidget.py
```diff
@@ -2458,6 +2458,8 @@ def event(
         modifiers=None,
         **kwargs,
     ):
+        if self.scene.pane.suppress_selection:
```
tatarize, line 2461: This can actually be done, probably in `ToolContainer` since we're dealing with what basically amounts to switching the effective selection widget. If you see the code in ToolContainer, it already accepts that some tools may have a `selection_mode` set. NodeMoveTool for example has `select_mode = "vertex"` which lets you select the relevant nodes only.

meerk40t/gui/toolwidgets/toolcontainer.py
```diff
@@ -36,17 +36,25 @@ def signal(self, signal, *args, **kwargs):
...
         self.remove_all_widgets()
         if tool is not None:
             new_tool = self.scene.context.lookup("tool", tool)
-            self.mode = getattr(new_tool, "select_mode", "selection")
+            if new_tool is None:
+                response = f"Such a tool is not defined: {tool}, leave unchanged"
+                return False, response
+            else:
+                self.mode = getattr(new_tool, "select_mode", "selection")
+                response = f"Tool set to {tool}, selection mode: {self.mode}"
```
tatarize, line 53: Given this is seems the suppressing the selection is odd... perhaps I don't understand.

meerk40t/extra/param_functions.py
```diff
@@ -0,0 +1,757 @@
...
+                if iterations is None:
+                    iterations = 10
+                iterations = int(iterations)
+            except ValueError:
+                channel("Invalid data provided")
+                return
+            start_pt = Point(ssx, ssy)
+            end_pt = Point.polar(start_pt, 0, blen)
+            geom = create_fractal_tree(start_pt, end_pt, iterations, ratio)
+            node = self.elem_branch.add(type="elem path", geometry=geom)
```
tatarize, line 265: Ah, this is why the parameters aren't inherent to the code, since it's creating these things as paths and then massaging the functions to be sort of equal. --- This feels like these should be `elem fractal`.

meerk40t/gui/toolwidgets/toolparameter.py
```diff
@@ -0,0 +1,528 @@
...
+
+import wx
+
+from meerk40t.gui.scene.sceneconst import RESPONSE_CHAIN, RESPONSE_CONSUME
+from meerk40t.gui.toolwidgets.toolwidget import ToolWidget
+
+_ = wx.GetTranslation
+
+
+class SimpleSlider:
```
tatarize, line 11: This belongs in it's own more general set of scene objects to add to a scene as a generalized thing. Fine for now.

PR #2134, opened 2023-10-01
Crash info

jpirnay:


meerk40t/gui/wxmeerk40t.py
```diff
@@ -952,8 +967,13 @@ def handleGUIException(exc_type, exc_value, exc_traceback):
...
     error_log += "".join(traceback.format_exception(exc_type, exc_value, exc_traceback))
-    print("\n")
-    print(error_log)
+    variable_info = ""
+    try:
+        frame = exc_traceback.tb_frame
+        variable_info = "\nLocal variables:\n"
+        variable_info += _variable_summary(frame.f_locals)
+    except Exception:
+        pass
```
tatarize, line 976: The f_locals in the crash frame are probably guaranteed to be safe. No real harm though.

tatarize (changes requested):
I fixed a couple helpful things like putting the branch into the Application version which corrects some code duplication.

However,  the rejection is based on a disclosure issue. 

The adding of the frame.f_locals is *not* reflected in the popup. The popup must disclose what will actually be sent to the developers. Omitting that we will now also be sending the local values is a breach of trust.

tatarize (approved):

PR #2140, opened 2023-10-06
Check for Update Frequency

jpirnay:
Allow specification how often mk should look for new versions: every startup, once per day, once per week


tatarize (commented):
This should probably also let you set this in the popup that said the time period, and also let you mark the check_for_update to be effectively never.

tatarize (approved):

PR #2145, opened 2023-10-10
Add kernel.is_shutdown test

jpirnay:


tatarize:
This is probably fighting the symptom of the problem rather than the problem itself. Life cycles are supposed to prevent this and are not doing that.

tatarize (changes requested):
Fights symptoms rather than the root cause.

PR #2152, opened 2023-10-11
Bug fixes for 0.9.2

tatarize:


jpirnay:
Things to address:
- [ ] dxf input crashes on invalid picture import
- [ ] invalid treatment of geomstr / svgelements hull
- [ ] runtime error if accessing already destroyed controls (during shutdown?)
- [ ] geomstr.Clip.clip not working reliably when having non-line primitives

tatarize:
I probably need an example of dxf with an image. I vaguely recall this was a thing but I couldn't really find one and considered having ezdxf make one but didn't get around to it.

tatarize:
> Things to address:
> 
> * [ ]  dxf input crashes on invalid picture import
> * [ ]  invalid treatment of geomstr / svgelements hull
> * [ ]  runtime error if accessing already destroyed controls (during shutdown?)
> * [ ]  geomstr.Clip.clip not working reliably when having non-line primitives

Pushing a beta instead to test high-dpi mode.

These are still needed and the PR isn't completed.

PR #2156, opened 2023-10-12
Balor: Fallback to bounds when live full is too long

jondale:
When running Live Full red line on balor device, large paths such as hatching are not feasible to show and cause Meerk40t to freeze.   This change causes it to fall back to bounds operation when the path length is larger than a set amount.

### DIFF
--- a/meerk40t/balormk/livelightjob.py
+++ b/meerk40t/balormk/livelightjob.py
@@ -15,6 +15,8 @@ from meerk40t.core.units import UNITS_PER_PIXEL, Length
 from meerk40t.svgelements import Matrix
 from meerk40t.tools.geomstr import Geomstr
 
+MAX_FULL_LENGTH = 2000000
+
 
 class LiveLightJob:
     def __init__(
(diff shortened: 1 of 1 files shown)


meerk40t/balormk/livelightjob.py
```diff
@@ -325,13 +327,24 @@ def _light_elements(self, con, elements):
...
             if self.stopped:
                 return False
             if self.changed:
                 return True
             geometry = Geomstr(node.as_geometry())
 
+            if self.mode == "full":
+                # Is path length too large for full?
+                for segment in range(0, len(geometry)):
+                    segment_length = geometry.length(segment)
```
tatarize, line 341: This `length()` call can be expensive. Like prohibitively, especially if you have cubic curves since they lack a closed form solution. You probably want to make this metric called based on `len(geometry)` or so. Also, if it's huge calculating the bounds might be somewhat costly too.

jondale: The original desire for this pr was because in the 0.8.x branch it would freeze during live full.  In the 0.9.x branch, it will run slowly on large paths but not freeze up the program.  The user can see this and switch modes manually so I don't think choosing that for them is useful.  Closing this pr.

PR #2174, opened 2023-10-19
Correct issues for the 0.9.3 merges

tatarize:
* Restore default image load class is `elem image`
* Fix display of `elem image`
* Permit creation of `image raster` from `elem image`.


tatarize:
See #2131

PR #2194, opened 2023-10-24
Add a bit more control over job loops

jpirnay:
image


meerk40t/gui/propertypanels/opbranchproperties.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, *args, context=None, node=None, **kwds):
                 "default": 1,
                 "type": int,
                 "conditional": (self.operation, "loop_enabled"),
-                "label": _("Loop"),
+                "label": _("Passes"),
```
tatarize, line 41: There's a use of `passes` inside the operations that usually just copies that operation. Which actually sort of sucks since you need a discrete amount rather than an infinite amount. But, that'd also suck with regards to optimization. Loops was used here since it's the same but different than pass. Also, hatches use the term loop, but there it's shared with ezcad terminology.

I'm not sure it's wildly important but eventually calling a bunch of different things `passes` just like we call a bunch of different things `operations` is going to catch up with us.

meerk40t/gui/wxmscene.py
```diff
@@ -1292,10 +1292,16 @@ def on_close(self, event):
 
     @signal_listener("pipe;running")
     def on_driver_mode(self, origin, state):
+        new_color = None
         if state:
-            self.widget_scene.overrule_background = wx.RED
+            new_color = wx.RED
         else:
-            self.widget_scene.overrule_background = None
+            try:
```
tatarize, line 1299: bit out of scope with the rest of the PR

tatarize (approved):

tatarize:
Big thing is spooler needs to not assume every job is a laser job since they could be things like RDJobs or Lightjobs. But, you correctly checked for it having the right function so it should be fine.

meerk40t/gui/wxmscene.py
jpirnay: It was too simple to leave out...

meerk40t/gui/propertypanels/opbranchproperties.py
jpirnay: Reverted the change of terms back to 'loop'

PR #2195, opened 2023-10-24
Allow deletion / duplication of entries in newly raster chart

jpirnay:
grafik


tatarize (approved):
Yeah, was sort of needed.

PR #2201, opened 2023-10-26
Py37

tatarize:


Sophist-UK:
I can fully understand why you are switching to Py37 however anyone running from source or PyPi on an ancient version of Linux with a previous version of 37 is going to have it stop working, so this probably needs a but more highlighting in the release notes and a bit of publicity.

jpirnay:
You may want to have a look at the commit, it's actually bringing down the need from py 3.8 to 3.7.1 just for the creation of the exe file under windows.

Sophist-UK:
Ah yes - apologies, I should indeed have looked at the changes.

tatarize:
Nah, he's not wrong this was a disaster and I screwed up. I meant to merge this into the local but it got merged upstream. I then had to quickly paper over that stuff. And restore it back to 3.8 or so. In theory this should be compatible, but we certainly don't want to do this across the board. I rapidly reverted this code.

tatarize:
I was trying to generate a version of meerk40t that would work for Windows 7 (Pre-SP1) and it did not generate correctly and when I tried to merge this into `tatarize:main` I accidently merged it to `meerk40t:main` and then had to revert that and it really didn't work correctly.

https://github.com/meerk40t/meerk40t/commit/b6bbd32190cbb3739837c6d02692a9ee380f544c

PR #2208, opened 2023-10-31
Next batch of vector icons

jpirnay:


meerk40t/gui/icons.py
```diff
@@ -4174,11 +3012,11 @@ def make_geomstr(self, gc, path):
...
 )
 
-icons8_laser_beam_20 = icons8_laser_beam_52
+icons8_laser_beam = icons8_laser_beam
```
tatarize, line 3019: Not technically wrong. But, not needed.

tatarize (approved):

meerk40t/gui/icons.py
jpirnay: Not sure from where that came from, will remedy it at the next iteration

Issue #2203, opened 2023-10-27
Is Meerk40t support Marlin firmware?


kpbicka:
### Summary Description

I didn't find in case how to connect 3D-printer Laser with Marlin firmware to meerk40t software 

### MeerK40t Version

0.9.2000

### MeerK40t Type

Executable

### Your Operating System

Mac (before M1/Bigsur)

Sophist-UK:
Marlin firmware is a one of a number of firmwares that support use of GCODE for lasering.

So really you need to be asking whether MeerK40t supports driving GCODE controllers - which it does in theory, though I am not sure exactly what the status is for that functionality and what the reliability is. 

A complementary use case is would be for MeerK40t to emulate a GCODE controller - AFAIK this functionality has not (yet) been written.

tiger12506:
> Marlin firmware is a one of a number of firmwares that support use of GCODE for lasering.
> 
> So really you need to be asking whether MeerK40t supports driving GCODE controllers - which it does in theory, though I am not sure exactly what the status is for that functionality and what the reliability is.

Pretty solid, actually. Tested with FluidNC and GRBL controllers both.

> A complementary use case is would be for MeerK40t to emulate a GCODE controller - AFAIK this functionality has not (yet) been written.

Actually, one does exist. grblcontrol is the gcode controller equivalent of ruidacontrol, because Lightburn also supports controlling grbl lasers. It was my understanding that this works, too.


I'm not certain if there are any tweaks that would need to be made to support Marlin, but I'm sure they would be fairly minimal.

tatarize:
Current no.

We support gcode and our gcode support is actually really quite good. You could save out some of our gcode from the normal setup and it would properly work. However most of our interfacing is built around GRBL which is mostly non-compatible with Marlin. The gcode is compatible with either though.

It's a little weird to understand but there's two main parts of any laser cutting control protocols. There's the program code, which is sequential and issued one line at a time, you complete one line, you move on to the next line until you finish the execution. And secondly there's the realtime non-out-of-band non-sequential commands. This usually includes things like `estop`, `pause`, `resume` and in GRBLs case some stuff like speed up, slow down, increase power. But, this is true for *all* laser controllers, and I've reversed engineered a lot of them by this time.

Our gcode support is quite good at this point, mostly because Jpirnay bought me a cheap Ortur2 diode laser, so the weird points got nicely softened. The `grblcontrol` stuff was also mostly finished and I couldn't poke a hole in it.

[...]

PR #2209, opened 2023-11-01
Marlin: Driver Basic

tatarize:
Best I can figure an extremely pared down GRBL driver could do most of the Marlin work.

It's a lot more like a super-generic gcode sender, with some SD card commands.

See analysis on #2203

---

`G4 I` -- Needs to be used to start.


tatarize:
Table until I dig out the 3d printer.

PR #2218, opened 2023-11-02
Rename icons to remove unneeded size indicator

jpirnay:


meerk40t/device/gui/defaultactions.py
```diff
@@ -4,16 +4,16 @@
...
     icons8_close_window_20,
-    icons8_down_50,
-    icons8_home_20,
+    icons8_down,
+    icons8_home_filled,
     icons8_input_20,
     icons8_output_20,
     icons8_remove_25,
     icons8_return_20,
     icons8_stop_gesture_20,
```
tatarize, line 13: These have size too.

tatarize (approved):

meerk40t/device/gui/defaultactions.py
jpirnay: And rightly so, as they are still bitmapped icons

PR #2222, opened 2023-11-03
Hatch classify

tatarize:
Add classification for hatches


jpirnay:
Hatches will be classified on creation and will be reclassified on a color-change within the property window.
Clicking on an element with a hatch, ie emphaszing it, and a subsequent click of any of the operations buttons in the statusbar will still classify the shape (and not the hatch). If you hold down the Shift-key while clicking on one of these then the hatch will be used for color-change/classification.

PR #2223, opened 2023-11-04
MButton: Custom Button

tatarize:
See Pause.

image

Due to the implementation in OSX we cant do color buttons and so we get a white box. And where we thought the color would be white on dark red, it's actually white on white box with a dark red border. This stuff messes up the consistency and we're doing more work than we should want to do.

However, we did that whole ribbonbar so we should be able to do a tiny custom button. 

* We could more consistently show the icons, texts, restricted sizes and enforce interoperability with signals and other mkernal internals.
* We can ensure a consistent look and feel between different OS platforms.
* We can send out signals for when the buttons are pressed in a consistent manner.
* We can make more robust buttons, for example, we can make an Outline button that lets you choose between the outline type, like a multi-button.
* We can directly use VectorIcons with the known colors so we don't hard-code the force_darkmode information.

### DIFF
--- a/meerk40t/gui/button.py
+++ b/meerk40t/gui/button.py
@@ -968,7 +968,7 @@ class Art:
                 i += 1
         dc.DestroyClippingRegion()
 
-    def layout(self, dc: wx.DC):
+    def layout(self, dc: wx.DC, button):
         self.button_layout(dc, self.parent.button)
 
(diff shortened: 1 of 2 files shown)


tatarize:
Table this can of worms. Current implementation removed the pulldown for multi-buttons which is wrong, so this likely would need more work than potentially just starting over or doing something else.

PR #2229, opened 2023-11-07
Consistent use of pause / laser_status

jpirnay:


meerk40t/device/gui/defaultactions.py
```diff
@@ -61,6 +61,8 @@ def __init__(self, *args, context=None, **kwds):
         sizer_after = StaticBoxSizer(self, wx.ID_ANY, _("At job end"), wx.VERTICAL)
         sizer_middle = wx.BoxSizer(wx.VERTICAL)
 
+        iconsize = dip_size(self, 30, 20)
```
tatarize, line 64: seems unrelated

tatarize (approved):
This seems correct, but it feels like some of the controllers that are pausing should maybe be a bit more self-defined. Also, I think there may have been calls to `.device.driver` in some of the gui parts that should probably be more overly found. This is certainly more consistent between devices but without the signal would seem more passive but can certainly update that status correctly. 

This also feels like it leaves some of the status information for the devices on the table. Some devices can query the laser for information and that call must be made async but this doesn't feel like it solves that.

PR #2235, opened 2023-11-09
Line Tool and ImageCut Tool, other background improvements.

tatarize:
* Add `tool line`
* Add `tool imagecut`
* `render_keyhole` command has more consistent options.
* Adds `image linecut` command to slice an image via a line.
* Line uses Functional Parameters.
* Improve `imageactions.py` calls to be documented and simplified.
* Add in Geomstr `append_lines` command
* Add in Geomstr `divide()` command
* Add in optional `breaks` flag for split.
* Add in optional `breaks` flag for polycut.

TODO:
- [ ] Imagecut does not retain position.
- [x] Single click crash for `imagecut`
- [ ] Implement `breaks` for quad and cubic curves.

### DIFF
--- a/meerk40t/core/node/elem_line.py
+++ b/meerk40t/core/node/elem_line.py
@@ -80,6 +80,9 @@ class LineNode(Node, Stroked, FunctionalParameter):
             0,
             self.x2,
             self.y2,
+            0,
+            (self.x1 + self.x2) / 2.0,
+            (self.y1 + self.y2) / 2.0,
         )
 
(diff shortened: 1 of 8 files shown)


tatarize:
The use of single click should probably be used in a slightly different mode where click puts down `p1` and then lets you click on `p2` without holding the mouse down, where right-click cancels the process. So you either do `click`, `move` `click` or you do `click, drag, release` but either should draw lines.

PR #2239, opened 2023-11-11
0.9.3 Beta 2: Final Touches

tatarize:
- [ ] Tiger Test: GRBL
- [x] Tiger Test: Lihuiyu
- [x] Tiger Test: Balor
- [x] Bug Checks 0.9.3

### DIFF
--- a/.github/workflows/win/file_version.txt
+++ b/.github/workflows/win/file_version.txt
@@ -6,8 +6,8 @@ VSVersionInfo(
   ffi=FixedFileInfo(
     # filevers and prodvers should be always a tuple with four items: (1, 2, 3, 4)
     # Set not needed items to zero 0.
-    filevers=(0, 9, 2001, 0),
-    prodvers=(0, 9, 2001, 0),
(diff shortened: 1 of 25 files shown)


tatarize:
GRBL needs added robustness.

Even though the GRBL device can technically pass it's a bit fickle to get and keep connected correctly.

- Usually connecting to the device does not trigger the welcome message. Only reset triggers the message.
- After connecting, if greet is not processed a status message should be requested to determine whether the device is correctly connected.
- The connection should not count as `connected` unless the device validates in some manner. Otherwise it should count as connecting.
- If it tried to send G91 without the device correctly connected it believes the G91 mode is engaged when it is not. This requires a confirmation `ok`.
- The connect button needs to show the com-port being connected to.
- The `?` data should be processed and potentially polled. For example, to query whether a job is finished or not.
- Serial connections do not detect the connection, they only detect the data.
- The status message reply will provide Pn:XY data as a reply.
- Cluster mode can be detected with `$I` and should be implemented.
- The device can provide several bits of information which should be used directly.

PR #2240, opened 2023-11-11
Raster optimisations: clustering of elements

jpirnay:
raster_opt

### DIFF
--- a/meerk40t/core/cutplan.py
+++ b/meerk40t/core/cutplan.py
@@ -175,6 +175,8 @@ class CutPlan:
         #     axis = rotary.axis
 
         original_ops = copy(self.plan)
+        if self.context.opt_raster_optimisation:
+            self.optimize_rasters(original_ops)
         self.plan.clear()
 
         idx = 0
@@ -638,6 +640,74 @@ class CutPlan:
         self._previous_bounds = None
         self.plan.clear()
         self.commands.clear()
+
+    def optimize_rasters(self, operation_list):
+        from meerk40t.core.node.op_raster import RasterOpNode
+
+        def generate_clusters(operation):
+            clusters = list()
+            cluster_bounds = list()
+            for node in operation.children:
+                if node.type == "reference":
+                    node = node.node
+                try:
+                    bb = node.paint_bounds
+                except AttributeError:
+                    continue
+                overlap = False
+                idx = 0
+                while idx < len(cluster_bounds):
+                    cc = cluster_bounds[idx]
+                    # The rectangles don't overlap if
+                    # one rectangle's minimum in some dimension
+                    # is greater than the other's maximum in
+                    # that dimension.
+                    flagx = (bb[0] > cc[2]) or (cc[0] > bb[2])
+                    flagy = (bb[1] > cc[3]) or (cc[1] > bb[3])
+                    # print (f"Checking\n"
+                    #        f"({bb[0]:.0f}, {bb[1]:.0f}, {bb[2]:.0f}, {bb[3]:.0f}) vs\n"
+                    #        f"({cc[0]:.0f}, {cc[1]:.0f}, {cc[2]:.0f}, {cc[3]:.0f}) "
+                    #        f"Overlaps: x={flagx}, y={flagy}"
+                    #     )
+                    if not (flagx or flagy):
+                        # print (f"Reuse cluster for {node.type}")
+                        overlap = True
+                        clusters[idx].append(node)
+                        cluster_bounds[idx] = (
+                            min(bb[0], cc[0]),
+                            min(bb[1], cc[1]),
+                            max(bb[2], cc[2]),
+                            max(bb[3], cc[3]),
+                        )
+                        break
+                    idx += 1
+
+                if not overlap:
+                    # print (f"Generate new cluster for {node.type}")
+                    cluster_bounds.append(
+                        ( bb[0], bb[1], bb[2], bb[3] )
+                    )
+                    clusters.append( [node] )
+            return clusters
+
+        idx = 0
+        for idx in range(len(operation_list) - 1, -1, -1):
+            op = operation_list[idx]
+            if not isinstance(op, RasterOpNode):
+                continue
+            clusters = generate_clusters(op)
+            if len(clusters) > 0:
+                # Create cluster copies of the raster op
+                for entry in clusters:
+                    newop = copy(op)
+                    newop._references.clear()
+                    for node in entry:
+                        newop.add_reference(node)
+                    newop.set_dirty_bounds()
+                    operation_list.insert(idx + 1, newop)
+
+                # And remove the original one...
+                operation_list.pop(idx)
 
 
 def is_inside(inner, outer, tolerance=0):
--- a/meerk40t/core/cutplan.py
+++ b/meerk40t/core/cutplan.py
@@ -17,7 +17,7 @@ CutPlan handles the various complicated algorithms to optimising the sequence of
 from copy import copy
 from math import isinf
 from os import times
-from time import time
+from time import time, perf_counter
 from typing import Optional
 
 import numpy as np
@@ -176,7 +176,9 @@ class CutPlan:
 
         original_ops = copy(self.plan)
         if self.context.opt_raster_optimisation:
-            self.optimize_rasters(original_ops)
+            self.optimize_rasters(original_ops, "op raster")
+            # We could do this as well, but images are burnt separately anyway...
+            # self.optimize_rasters(original_ops, "op image")
         self.plan.clear()
 
         idx = 0
@@ -641,9 +643,7 @@ class CutPlan:
         self.plan.clear()
         self.commands.clear()
 
-    def optimize_rasters(self, operation_list):
-        from meerk40t.core.node.op_raster import RasterOpNode
-
+    def optimize_rasters(self, operation_list, op_type):
         def generate_clusters(operation):
             clusters = list()
             cluster_bounds = list()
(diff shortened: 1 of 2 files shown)


Sophist-UK:
Personally I don't see a downside from this, and so don't see the need for an Execute option for it (though useful for the video example above). Users will just find that the burn takes less time.

Comments (if any) on the code to follow.

Sophist-UK:
Personally I don't see a downside from this, and so don't see the need for an Execute option for it (though useful for the video example above). Users will just find that the burn takes less time.

Comments (if any) on the code to follow.

Sophist-UK:
I am not tied to the specific algorithm and implementation I wrote for 0.7.x and I have no issues with the idea of a different algorithm and different code being written this time around.

However, a quick look at the code suggests that the algorithm is flawed. Because once a cluster is created it is never going to be merged with another cluster even if they overlap - and you can get different results with the same elements in different orders.

Imagine a square frame consisting of 8 slightly overlapping black-filled square boxes - 4 corners and 4 sides. 

If the element order starts with a box and goes sequentially around the frame, you will end up with 1 cluster. However, **if I read the code correctly**...

If the element sequence is the 4 corners first and then the 4 sides (which means that the 4 side boxes are "on top" of the 4 corner boxes), the 4 corners (which do not overlap) will create 4 clusters, and then the sides will be added to the first cluster they overlap with. And you end up with 4 clusters and 4 images - images which overlap and will create a double burn. Or if the corner squares are black-filled, and the sides white-filled and on top, then you will probably get areas burned which should have been masked.

It should be possible to create two SVGs identical except having different element orders and see what happens in the simulation.

Regarding other changes of approach, from memory I think my old code was inserted into the `blob` phase (because that is where images are created from the raster elements), and this code runs earlier in the `preprocess`, so I suspect that the change to splitting the groups into separate ops which later get turned into separate images is not in principle going to be materially different from my original algorithm, and indeed because this change is entirely contained in this class and doesn't require changes in the `as_cutobjects` method of operations (which my code previously did), then it seems to be a better approach.

jpirnay:
I think you are right, it needs reiterating the clusters and combining them if needed

Sophist-UK:
It was a couple of years ago now, but I seem to recall that this was why I created single entry groups and then iterated to merge them rather than attempt to create merged groups from the start.

Sophist-UK:
I would also suggest that you implement the advanced version (which iterates inside each group to double check overlap by individual rasters) whilst you are at it.

jpirnay:
This should fully cover overlaps **within a single raster operation** - without any safety margin, so adjacent rectangles with a microscopic gap between them would still be burnt separately.
There's no check across multiple raster operations.
test
raster_opt

jpirnay:
Additionally this isn't in any form or shape numpy optimised, but this example took just a few milliseconds:
```
[05:59:33]     Optimise op raster finished after 0.01 seconds, inflated 1 operations to 4
```

tatarize (approved):
This actually leaves a bit of optimization on the table. Since the original 0.7.x code was not expanding the bounding box to the bigger one of the full group but finding all objects which had bounding boxes and linking them together if those bounding boxes overlapped.

The somewhat confusing part was that if you linked 1 and 2 then linked 4 and 5. Then found a bounding box that linked 2 and 4, They all ended up in the same group. And did so without expanding the underlying bounding box.

I will note that the technically correct method is actually to put the function of the `optimize_rasters()` into the `self.commands` they will all be executed during the `process` stage which executes all of that data.

The previous distinctions aren't that important since we're not doing step by step. But, the work is supposed to be done in the `process` and `optimize` steps and the `preprocess` and `preopt` steps are mostly just setting up the commands to be executed rather than executing them directly.

tatarize:
The cutplanning steps:
copy: moves the stuff over.
preprocess: figure out all the commands to be run and add everything to the `.commands`
validate: execute all those functions, just run each function there. (this would better be called `process`).

blob: convert everything to cutcode.

preopt: figure out all the commands to be run on the cutcode this time around, add that to `.commands`
optimize: execute all the preopt commands, just run each function there.
spool: send this whole thing to the spooler.

tatarize:
Yeah, I wouldn't really expect numpy to be that much help. I'm pretty sure even with thousands of objects it might still be remarkably fast. But, the current code is non-existing so you have it beat for a feature Sophist really wants back working. If I want to write a numpy-optimized section of it later that reduces the bounding box footprint (make groups out of sets of overlapping bounding boxes regardless of order to identify clusters) and takes even less time than very very little time, that could be written later.

I'm not sure group by by overlap is that useful outside of this functionality, or that this code is hard enough to warrant taking basically no time at all.

Sophist-UK:
> The somewhat confusing part was that if you linked 1 and 2 then linked 4 and 5. Then found a bounding box that linked 2 and 4, They all ended up in the same group. And did so without expanding the underlying bounding box.

If they were grouped but the overall bbox was not expanded, then that would have been a bug.

meerk40t/core/cutplan.py
```diff
@@ -175,6 +175,10 @@ def preprocess(self):
         #     axis = rotary.axis
 
         original_ops = copy(self.plan)
+        if self.context.opt_raster_optimisation:
+            self.optimize_rasters(original_ops, "op raster")
+            # We could do this as well, but images are burnt separately anyway...
```
Sophist-UK, line 180: There is actually a second potential optimisation algorithm which could be applied to large images which is a whitespace optimisation to chop images with large areas of whitespace (think of a large picture frame) into separate smaller images.

So in the picture frame example, you might chop the image into two smaller images and raster the top and left side in one image, and the bottom and right in another. The potential issue is that when you move the head back to the origin point, any sloppiness in the hardware or loss of position might result in the two images not aligning exactly.

This is image analysis, and so would need to be done in Numpy or similar where it is done in compiled code using all cores in parallel.

Another image optimisation (but not needed for rasters converted to images) which would be useful for cut inner first processing, is to remove any vertical and horizontal white margins in the image. This could be done at load time at the same time as changing white to transparent - though the change of dimensions (but not image size) might be confusing if the user has specifically sized the image with margins in e.g. Inkscape.

meerk40t/core/cutplan.py
```diff
@@ -639,6 +643,88 @@ def clear(self):
...
+            clusters = list()
+            cluster_bounds = list()
+            for node in operation.children:
+                if node.type == "reference":
+                    node = node.node
+                try:
+                    bb = node.paint_bounds
+                except AttributeError:
+                    continue
+                clusters.append([node])
```
Sophist-UK, line 657: In my original code I think I had the node and its bbox as a tuple in a single array. There is no reason why they can't be held separately like this, but it does mean that the cluster merges need to be done to the two arrays in sync. and that is a tiny bit more likely to go astray than holding them together as a tuple.

With a bit more insight into modern coding techniques, I think that if I was recoding the 0.7 version now I would use a data-object to hold a cluster rather than a tuple.

meerk40t/core/cutplan.py
```diff
@@ -639,6 +643,88 @@ def clear(self):
...
+                cluster_bounds.append((bb[0], bb[1], bb[2], bb[3]))
+
+            needs_repeat = True
+            while needs_repeat:
```
Sophist-UK, line 661: A very very minor point, but would a `while true:` and use of `break` be a better code style?

meerk40t/core/cutplan.py
```diff
@@ -639,6 +643,88 @@ def clear(self):
...
+                needs_repeat = False
+                for outer_idx in range(len(clusters) - 1, -1, -1):
+                    # Loop downwards as we are manipulating the arrays
+                    bb = cluster_bounds[outer_idx]
+                    inner_idx = outer_idx - 1
+                    while inner_idx >= 0:
```
Sophist-UK, line 667: `for inner_idx in range(outer_idx - 1, -1, -1):` ??

Sophist-UK (commented):
Overall the code looks like a good implementation of my basic algorithm however:

1. I still believe that use of a small **fixed** margin (rather than the dynamicly calculated margin in my 0.7.x version) is worth including. Imagine the case of two squares which are positioned with a corner touching but not overlapping - would we really want these put into separate rasters (which would happen with the current code) or done in a single raster). With different rasters, any hardware sloppiness or loss of position would result in misalignment of two images which should literally be touching).

2. I still believe that implementation of another loop checking overlap of individual element bboxes before merging clusters would be beneficial - and if you create an `are_bboxes_overlapping(bb1, bb2): boolean` function it would only be (say) 3-5 extra lines of code.

3. I would recommend creation of an SVG and / or some automated test cases for the clustering code which will test for all these edge cases.

meerk40t/core/cutplan.py
tatarize: I'm adding but not totally done with `tool imagecut` which will let you manually chop up your images in place. Basically this would let you slice your image with a line. There's a still broken version of this code in main (and mostly hiding as a hidden feature). Also `tool line` but that works fine.

meerk40t/core/cutplan.py
tatarize: He's breaking the inner loop there at a particular point. So it's a lot harder to break correctly, and no. You'd use a function for this for better code style. Nearly always when you're using `while true` and `break` you're better off writing the inner loop as a function and returning.

Though you might be able to do multiple merges if you iterated backwards.

Sophist-UK:
> I will note that the technically correct method is actually to put the function of the `optimize_rasters()` into the `self.commands` they will all be executed during the `process` stage which executes all of that data.

If there is going to be a second PR to address my review comments, then I would suggest that this is included.

meerk40t/core/cutplan.py
Sophist-UK: > He's breaking the inner loop there at a particular point. So it's a lot harder to break correctly, and no. You'd use a function for this for better code style. Nearly always when you're using `while true` and `break` you're better off writing the inner loop as a function and returning.
> 
> Though you might be able to do multiple merges if you iterated backwards.

Use of a function - I agree. This comment was (as stated) a very minor one.

Multiple merges - I think the code has removed the entry you are comparing with so you need to break and move to the next one down the list.

meerk40t/core/cutplan.py
Sophist-UK: I imagine that there may be some overlap of code here.

PR #2245, opened 2023-11-12
Additional margin + refined overlap test for raster optimisation

jpirnay:


meerk40t/core/cutplan.py
```diff
@@ -643,8 +647,21 @@ def clear(self):
...
 
-    def optimize_rasters(self, operation_list, op_type):
+    def optimize_rasters(self, operation_list, op_type, margin):
         def generate_clusters(operation):
+            def overlapping(bounds1, bounds2, margin):
+                # The rectangles don't overlap if
+                # one rectangle's minimum in some dimension
+                # is greater than the other's maximum in
+                # that dimension.
+                flagx = (bounds1[0] - 0.5 * margin > bounds2[2] + 0.5 * margin) or (
```
tatarize, line 657: Margin is the extra space between the stuff and the edge, not necessarily the sum of the total margins. I see the logic but I think firstly, if you do half margins you should do `m = margin / 2` assignment above this because the readability here is damaged, but rather should just use the actual literal margins and just use `margin`.

tatarize: Actually if you wanted to use half margins you'd probably just subtract the margin from one side and not the other since `q -half_margin > r +half_margin` is the same as `q > r +margin`

tatarize (approved):
half margins seem weird, but *shrug* -- 1 inch margins leave increase the total size by 2 inches along each dimension.

meerk40t/core/cutplan.py
jpirnay: Simplified code

jpirnay:
You may want to review the changes again as some fixes / code refactoring were introduced

tatarize (approved):

PR #2253, opened 2023-11-15
Add a lock around icon generation

jpirnay:


meerk40t/gui/icons.py
```diff
@@ -569,216 +586,207 @@ def GetBitmap(
...
-                # values beyond 100 indicate lighter
-                # no change = 100
-                # What about black? This is a special case, so we only consider
-                ialpha = info_bright / 100.0
-                if color.red == color.green == color.blue == 0 and info_bright > 100:
-                    ialpha = (info_bright - 100) / 100.0
-                    cr = int(255 * ialpha)
-                    cg = int(255 * ialpha)
-                    cb = int(255 * ialpha)
+        with self._lock:
```
tatarize, line 594: * This is could still duplicate the work rather than use the cache. You sort of need to do double gated syntax here. 
* The function is also getting a lot complex and maybe should be subdivided into different functions.
* Especially if you could avoid the local caching scheme with the lru_cache decorator.  (https://realpython.com/lru-cache-python/)

tatarize: I'm a bit of a never nester so I would sort of prefer if within the lock it just calls the function that does the rest of the work stuff. You can also avoid the massive ----, ++++ in the diff log that way. And make it a bit easier to see what's being actually altered.

tatarize: The reason this could still double execute is that if two threads hit the lock around the same time, only one thread will get through the lock. Then the other thread will be forced to wait until that thread finishes. When the other thread finishes it will then release this thread which will then repeat all the same work again pointlessly even though it's already done.

Weirdly the correct method is double-gated. You check the cache, then you enter the lock, then you check the cache again. It would seem like you already know the cache was a miss, but you really don't because it could have been a while since you acquired the lock. 

So you do:
```python
        if cache_id in _CACHE:
            return _CACHE[cache_id]
        with self._lock:
            if cache_id in _CACHE:
                return _CACHE[cache_id]
            self._function_that_makes_bitmap_and_fills_cache(cache_id)
```

meerk40t/gui/icons.py
```diff
@@ -569,216 +586,207 @@ def GetBitmap(
...
+                if "join_miterclip" in attrib:
+                    spen.SetJoin(wx.JOIN_MITER)
+                gc.SetPen(spen)
+                gc.StrokePath(gp)
+            dc.SelectObject(wx.NullBitmap)
+            gc.Destroy()
+            del gc.dc
+            del dc
+            # Save bitmap for later retrieval
+            _CACHE[cache_id] = bmp
```
tatarize, line 789: Due to the extra tab, I can't tell if anything changed. I assume not.

tatarize (approved):

meerk40t/gui/icons.py
jpirnay: Removed the local caching scheme in favor of the functools lru-cache functionality.

meerk40t/gui/icons.py
jpirnay: - The double cache check should no longer be relevant, as it has been replaced by the lru-cache decorator.
- Routines have been isolated

PR #2265, opened 2023-11-17
Provide undo/redo indicator

jpirnay:


meerk40t/core/undos.py
```diff
@@ -108,3 +108,19 @@ def undolist(self):
...
+            if self._undo_index == 0:
+                # At bottom of stack.
+                return False
+            if len(self._undo_stack) == 0:
+                # Stack is entirely empty.
+                return False
+        return True
+
+    def has_redo(self):
+        with self._lock:
```
tatarize, line 123: You don't need these locks. If you're not manipulating the data it won't matter if multiple threads look at the same data.

meerk40t/core/elements/elements.py
```diff
@@ -526,14 +526,15 @@ def __init__(self, kernel, index=None, *args, **kwargs):
...
-            self.undo.mark("op-loaded")  # Mark defaulted
+        with self.undofree():
+            self.load_persistent_operations("previous")
+
+            ops = list(self.ops())
+            if len(ops) == 0 and not self.operation_default_empty:
+                self.load_default(performclassify=False)
+            # if list(self.ops()):
+            #     # Something was loaded for default ops. Mark that.
+            #     self.undo.mark("op-loaded")  # Mark defaulted
```
tatarize, line 537: No longer marks the loaded ops?

tatarize (approved):

PR #2268, opened 2023-11-18
Experimental: Remove svgelements/kernel replace with import references

tatarize:
* Use mkernel and svgelements directly via import rather than including the file in place.


tatarize:
Kind of a lateral move 6 of 1, half-dozen of another. Probably more pythonic to do it like this.

PR #2269, opened 2023-11-18
Helper Window + Online Help

jpirnay:
- Add a helper window that will display the tooltip of the control the mouse is currently hovering over. Tooltips disappear after a short period of time, while this window will have the information as long as the mouse remains over the control.
- Additionally a wiki page describing the functionality of this control / the window it is in embedded can be opened via F11 (uses the wxmain menu functionality for this overarching key).
Internally this uses the SetHelpText functionality of any wxPython window (control or panel) and uses this retrieved value to open a corresponding wiki page. This page is constructed via ``url = f"https://github.com/meerk40t/meerk40t/wiki/Online-Help:-{sect}"`` where ``sect`` is the retrieved helptext value.
Example:
grafik

Corresponding page:
https://github.com/meerk40t/meerk40t/wiki/Online-Help:-CAG


tatarize (approved):
I don't see an online help button, but seems pretty good. Does put us on the hook for a lot of documentation that we should have done.

PR #2270, opened 2023-11-19
GRBL: Imply GRBL home offsets

tatarize:
* allow forced validation
* add in origin shift offset


meerk40t/grbl/device.py
```diff
@@ -124,22 +124,22 @@ def __init__(self, kernel, path, *args, choices=None, **kwargs):
                 "signals": "bedsize",
             },
             {
-                "attr": "home_bottom",
```
tatarize, line 127: This needs a better interface. Maybe just a combo box.

meerk40t/grbl/driver.py
```diff
@@ -175,12 +175,21 @@ def move_rel(self, dx, dy):
         @param dy:
         @return:
         """
-        self._g91_relative()
+        self._g90_absolute()
```
tatarize, line 178: Revert this, just for testing.

PR #2274, opened 2023-11-19
Make m2nano controller window a bit more compact

jpirnay:
Made control sizes a bit more compact for Lihuiyu controller:

grafik

Added device info to controll button for GRBL + reduced bitmap size a bit

grafik

### DIFF
--- a/meerk40t/lihuiyu/gui/lhycontrollergui.py
+++ b/meerk40t/lihuiyu/gui/lhycontrollergui.py
@@ -20,9 +20,9 @@ from meerk40t.kernel import signal_listener
 
 _ = wx.GetTranslation
 
-_simple_width = 500
+_simple_width = 565
 _advanced_width = 952
-_default_height = 584
+_default_height = 570
 
 
 class LihuiyuControllerPanel(ScrolledPanel):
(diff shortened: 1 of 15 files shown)


meerk40t/lihuiyu/gui/lhycontrollergui.py
```diff
@@ -356,7 +367,7 @@ def on_connection_state_change(self, origin, state):
             self.button_device_connect.SetLabel(_("Connect failed"))
             self.button_device_connect.SetBitmap(
                 icons8_disconnected.GetBitmap(
-                    use_theme=False, resize=get_default_icon_size()
+                    use_theme=False, resize=0.75 * get_default_icon_size()
```
tatarize, line 370: You probably want to set `factor = 0.75` in there and just use that since otherwise these numbers are more fully hardcoded. Not that I'm recommending hardcoding them, but making it all adjustable in code from one line.

tatarize (approved):

meerk40t/lihuiyu/gui/lhycontrollergui.py
jpirnay: Done

PR #2278, opened 2023-11-21
Add in GRBL Interpolation tweaks.

tatarize:
See Discord discussion (11-21-2023 11:29 PM PST). The GRBL code interpolates a curve into 50 subpoints regardless how intricate the curve is. This means small fragments of curve are made more fragmented and does not account for the speed change within bezier curves ( https://youtu.be/aVwxzDHniEw?si=nUytSTNjguAWS1nn&t=907 ).

The algorithm around this is usually to use a lookup table of distance and slice into that lookup table. Here I did a bit of a different thing but it mostly would work slightly inferior.


tatarize:
Fix the distance incorrectness due to distance from mark rather than distance through curve. Fix the distance sum LUT algorithm working correctly/fully.

PR #2280, opened 2023-11-21
Exif image rotation

tatarize:
Some images have an exif rotation applied to them purely in metadata. This will provide those rotations if they exist.


tatarize:
image
Portrait_2

Main thread previously:
image

PR #2282, opened 2023-11-22
Tips at startup

jpirnay:
Display tips at startup (can be suppressed, will still be accessible via the Panes-Help menu)

image

The tips can be maintained outside of the regular release cycle by updating a file on github: https://github.com/meerk40t/meerk40t/blob/main/locale/tips.txt
The logic will once a week query that file and hold a local copy. It will look first in the language-specific subdirectory and then fallback to the english version.
```
# This file contains tips to be displayed at program start. Every entry has up to 3 lines:
#   tip=xxxx Tip to be displayed
#   version=x.x.x (optional) Minimum version to apply, you can just omit that line to allow it for any version
#   cmd=xxxx Command to execute for the 'Try it out' button
# NB: Every tip= will reset cmd and version, so please use the sequence tip, cmd, version
#
# Example:
# tip=Did you know, these tips can be maintained and extended online.\nWow, that's great.
# cmd=
# version=0.9.3
#
# tip=That won't be visible for a while
# version=1.0
```


tatarize (changes requested):
I don't love the idea of the program phoning home like that.

If we have tips we should hard code them at build time.

jpirnay (commented):
Changes done according to agreement for intentional download

meerk40t/gui/tips.py
```diff
@@ -0,0 +1,461 @@
...
+                parts = path.split("/")
+                if len(parts) > 0:
+                    basename = f"_{counter}_" + parts[-1]
+                    local_path = os.path.join(self.cache_dir, basename)
+                    # Is this file already on the disk? If not load it...
+                    if not os.path.exists(local_path):
+                        loaded = False
+                        opened = False
+                        try:
+                            with urllib.request.urlopen(path) as file:
```
tatarize, line 172: Could still download things without strict authorization.

tatarize (changes requested):
If you deleted the images it could still download replacement images without explicit permission.

meerk40t/gui/tips.py
jpirnay: resolved

tatarize (changes requested):
I feel like this is really close. But, I can't easily revoke the ability to download those automatically. They should either have to press `update` to get the missing images or they should have to hit `update` on that individual image.

Once permissions are given here to automatically download the image it can't be revoked. Even just a text box that says `Image not found. Try updating.` would meet the otherwise seemingly stringent rules. Since it could just download the images after the user requested this.

* The program should not contact the internet without express permission.
* Permission should be revocable, but strongly preferred user initiated.

tatarize:
Even a checkbox that defaults to off, that gives a blanket `automatically update` would probably suffice. Just ask before you download, and let people suspend that access. Rather than begging them to click update a lot. Just `request access` and `allow revoke access`.

tatarize (approved):

Issue #2293, opened 2023-11-23
[Bug]: Window formatting contents incorrectly  


jpagliaccio:
### Summary Description

Window formatting contents (gadgets) incorrectly. For example the Navigation pop-out window contents are cropped no matter the window size. Many Gtk warnings are produced for width and size. The Navigation window produces the most Gtk warnings.   

### Additional Details

For example: 

![image](https://github.com/meerk40t/meerk40t/assets/7680161/205db47e-5a10-47f3-90cc-504975c4568b)

In my shell window there are many repeated Gtk width warnings (eg: Negative content width -9).

### Crash logs

```shell
-------------------------------------------------------
Clicking the Navigate Button (a subset of the output)
-------------------------------------------------------
(MeerK40t-Linux-32:109753): Gtk-WARNING **: 08:56:28.475: Negative content width -14 (allocation 20, extents 17x17) while allocating gadget (node button, owner GtkButton)

(MeerK40t-Linux-32:109753): Gtk-WARNING **: 08:56:28.510: Negative content width -9 (allocation 25, extents 17x17) while allocating gadget (node button, owner GtkButton)

(MeerK40t-Linux-32:109753): Gtk-WARNING **: 08:56:28.510: Negative content width -14 (allocation 20, extents 17x17) while allocating gadget (node button, owner GtkButton)

(MeerK40t-Linux-32:109753): Gtk-WARNING **: 08:56:28.539: Negative content width -9 (allocation 25, extents 17x17) while allocating gadget (node button, owner GtkButton)

(MeerK40t-Linux-32:109753): Gtk-WARNING **: 08:56:28.539: Negative content width -14 (allocation 20, extents 17x17) while allocating gadget (node button, owner GtkButton)

(MeerK40t-Linux-32:109753): Gtk-WARNING **: 08:56:28.573: Negative content width -9 (allocation 25, extents 17x17) while allocating gadget (node button, owner GtkButton)

(MeerK40t-Linux-32:109753): Gtk-WARNING **: 08:56:28.573: Negative content width -14 (allocation 20, extents 17x17) while allocating gadget (node button, owner GtkButton)

(MeerK40t-Linux-32:109753): Gtk-WARNING **: 08:56:28.624: Negative content width -9 (allocation 25, extents 17x17) while allocating gadget (node button, owner GtkButton)

[...]

PR #2294, opened 2023-11-23
Arbitrary size correction for Linux

jpirnay:
Addresses issue #2293


tatarize (approved):
I can tell this is safe, I'm not sure on the essential nature of the changes.

PR #2296, opened 2023-11-24
Polygon tool

jpirnay:
Made regular polgyon design more logical. Added another tool


tatarize (approved):
Somehow I think so long as the polygon tools are different they might want to be different tools in different files with different registered names. Where tools maybe have a subclass such that we put all the polygon-classed tools into a group. With the current method this tool can't be extended by a plugin for example.

PR #2300, opened 2023-11-25
Unify all line based tools

jpirnay:
As quite a few tools use nearly the same code and the same logic (click on a couple of points and create an element from it), I have created a helper class the encapsulates all the ToolWidget logic and just needs a couple of overloaded functions per tool.

### DIFF
--- a/meerk40t/gui/toolwidgets/toolpointlistbuilder.py
+++ b/meerk40t/gui/toolwidgets/toolpointlistbuilder.py
@@ -12,6 +12,7 @@ from meerk40t.gui.scene.sceneconst import (
 from meerk40t.gui.toolwidgets.toolwidget import ToolWidget
 from meerk40t.svgelements import Point
 
+
 class PointListTool(ToolWidget):
     """
     This tool is a basic widget just compiling a list of points.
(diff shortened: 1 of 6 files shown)


meerk40t/gui/toolwidgets/toolpointlistbuilder.py
```diff
@@ -0,0 +1,238 @@
...
+        followup = self.create_node()
+        self.scene.pane.tool_active = False
+        self.point_series = []
+        self.mouse_position = None
+        self.scene.request_refresh()
+        if followup:
+            self.scene.context(f"{followup}\n")
+
+    def create_node(self):
+        print("Needs to be overloaded!")
```
tatarize, line 225: This should `raise NotImplementedError`

exception NotImplementedError[¶](https://docs.python.org/3/library/exceptions.html#NotImplementedError)
This exception is derived from [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError). In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

tatarize (approved):

meerk40t/gui/toolwidgets/toolpointlistbuilder.py
jpirnay: Okay, makes sense

PR #2302, opened 2023-11-25
Fix dxf text and image import

jpirnay:


tatarize:
* Got the DPI for Images Correct.
* Got the size for Images Correct.
* Text Orientation was fixed.

Image position is wrong.

tatarize (approved):

PR #2307, opened 2023-11-26
Different restart mechanism

jpirnay:
The current loop approach crashes under Linux all the time. Different approach with ``os.execvp`` - works both under Linux and Windows - might not work in the compiled executable - requires testing.

### DIFF
--- a/meerk40t/gui/wxmmain.py
+++ b/meerk40t/gui/wxmmain.py
@@ -228,13 +228,15 @@ class MeerK40t(MWindow):
 
         # Look at window elements we are hovering over
         # to establish the online help functionality
-        self.timer = wx.Timer(self, id=wx.ID_ANY)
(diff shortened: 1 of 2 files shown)


tatarize:
In default windows using pycharm IDE this restart mechanism loses the ide. It restarts and no longer has the regular log or debug.

One failed restart gave me:
`C:\Users\David\AppData\Local\Programs\Python\Python311\python.exe: can't open file 'C:\\Program': [Errno 2] No such file or directory`

And the other method worked perfectly. So maybe divide by OS.

tatarize:
Commandline in windows and compiled windows worked fine. It's actually only an issue for the IDE and the general rule there would just be `don't restart` that. So this may actually be a pass for windows too.

tatarize:
After some fixes to the `Mouse Query` and `Double-Buffer Signals` issues this may not be needed anymore.

Please verify there's any issue with the current mechanism on Linux.

PR #2319, opened 2023-12-02
Hatch fixes

jpirnay:
- Allow substructures (groups, files) under effect hatch and effect wobble
- Fix inconsistent angle treatment in hatch propertypanel

### DIFF
--- a/meerk40t/core/elements/element_treeops.py
+++ b/meerk40t/core/elements/element_treeops.py
@@ -171,20 +171,34 @@ def init_tree(kernel):
     @tree_operation(_("Remove hatch"), node_type=elem_nodes, help="")
     def unhatch_elements(node, **kwargs):
         for e in list(self.elems(emphasized=True)):
-            nparent = e.parent
-            if nparent.type.startswith("effect"):
-                e._parent = None  # Otherwise add_node will fail below
-                try:
-                    idx = nparent._children.index(node)
-                    if idx >= 0:
-                        nparent._children.pop(idx)
-                except IndexError:
-                    pass
-                nparent.parent.add_node(e)
-                if len(nparent.children) == 0:
-                    nparent.remove_node()
-                else:
-                    nparent.altered()
+            # eparent is the nodes immediate parent
+            # nparent is the containing hatch
+            eparent = e.parent
+            nparent = eparent
+            while True:
+                if nparent.type.startswith("effect"):
+                    break
+                if nparent.parent is None:
+                    nparent = None
+                    break
+                if nparent.parent is self.elem_branch:
+                    nparent = None
+                    break
+                nparent = nparent.parent
+            if nparent is None:
+                continue
+            e._parent = None  # Otherwise add_node will fail below
+            try:
+                idx = eparent._children.index(node)
+                if idx >= 0:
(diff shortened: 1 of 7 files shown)


meerk40t/core/elements/element_treeops.py
```diff
@@ -167,24 +167,49 @@ def image_convert_modifier(node, **kwargs):
             type="elem image",
         )
 
+    @tree_operation(_("Remove effect"), node_type=effect_nodes, help="")
+    def remove_effect(node, **kwargs):
+        childs = [e for e in node._children]
+        for e in childs:
```
tatarize, line 173: `for e in list(node._children):`

Codacy whines about the unneeded comprehension.

meerk40t/core/elements/element_treeops.py
```diff
@@ -167,24 +167,49 @@ def image_convert_modifier(node, **kwargs):
...
+            e._parent = None  # Otherwise add_node will fail below
+            node.parent.add_node(e)
```
tatarize, line 175: This is actually `append_child` it might need a bit of a better name, but append much like `insert_sibling` detaches the child, and calls the relevant notifications.

It's probably fine, but we're sort of supposed to call the notifications for detaching and attaching.

meerk40t/core/elements/element_treeops.py
```diff
@@ -167,24 +167,49 @@ def image_convert_modifier(node, **kwargs):
...
+        node._children.clear()
+        node.remove_node()
+        self.signal("rebuild_tree")
+
+
     @tree_conditional(lambda node: is_hatched(node))
```
tatarize, line 181: The is_hatched command is correct but confusingly named since it does `is_effected` now.

meerk40t/core/elements/element_treeops.py
```diff
@@ -167,24 +167,49 @@ def image_convert_modifier(node, **kwargs):
...
+                    nparent = None
+                    break
+                if nparent.parent is self.elem_branch:
+                    nparent = None
+                    break
+                nparent = nparent.parent
+            if nparent is None:
+                continue
+            e._parent = None  # Otherwise add_node will fail below
+            try:
```
tatarize, line 202: This is a lot of manipulation of the node at the level that non-node class functions probably shouldn't. Repeatedly accessing `_children` deleting parents, re-adding non-pristine freshly created nodes.

meerk40t/core/elements/shapes.py
```diff
@@ -253,21 +253,33 @@ def effect_remove(
...
-                except IndexError:
-                    pass
-                nparent.parent.add_node(node)
-                if len(nparent.children) == 0:
-                    nparent.remove_node()
-                else:
-                    nparent.altered()
+            eparent = node.parent
+            nparent = eparent
+            while True:
```
tatarize, line 258: This repeated code is also probably a node function worth writing. `find_ancestor(criteria)` or something. Also `while nparent:` is  probably better than a `while True:`

tatarize (approved):
Bit hard to read, seems like a bit of repeated code in there.

Otherwise, probably fine.

(Higher scrutiny for the beta)

PR #2321, opened 2023-12-04
GRBL Connection Robustness.

tatarize:
* Reset on Connect.
* Require starting sequence.

### DIFF
--- a/meerk40t/newly/gui/newlycontroller.py
+++ b/meerk40t/newly/gui/newlycontroller.py
@@ -149,7 +149,8 @@ class NewlyControllerPanel(wx.ScrolledWindow):
             self.context("usb_connect\n")
 
     def pane_show(self):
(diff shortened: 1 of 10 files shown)


meerk40t/newly/gui/newlycontroller.py
```diff
@@ -149,7 +149,8 @@ def on_button_start_connection(self, event):  # wxGlade: Controller.<event_handl
             self.context("usb_connect\n")
 
     def pane_show(self):
-        name = self.service.label
+        name = self.service.label.replace(" ", "-")
+        name = name.replace("/", "-")
         self.context.channel(f"{name}/usb").watch(self.update_text)
```
tatarize, line 154: This is strictly speaking allowed but a lot harder to listen to the channel since you'd need to quote the damn thing, so this is probably better.

PR #2337, opened 2023-12-11
Material Manager

jpirnay:
Current state of implementation

- [x] Filtering
- [x] Deletion
- [x] Duplication
- [x] In line editing of material ops
- [ ] Sharing to / Loading from central repository
- [x] Import ezcad / lburn
- [x] First feedback (hierarchical view, additional properties)


meerk40t/gui/materialmanager.py
```diff
@@ -0,0 +1,1216 @@
...
+        settings.write_configuration()
+        # print ("Sharing")
+        try:
+            with open(local_file, "r") as f:
+                data = f.read()
+        except:
+            return
+        self.send_data_to_developers(local_file, data)
+
+    def send_data_to_developers(self, filename, data):
```
tatarize, line 410: This is weird to have here.

jpirnay: The idea was to have an opportunity to share your settings with the mk community. It's a stub for the time being as the backend counterpart needs to be in place too.

tatarize (approved):
This is good enough and far enough along that I'm fine with merging it and fixing it going forward if there are other issues, or disabling it if there are too many issues.

meerk40t/gui/materialmanager.py
```diff
@@ -726,17 +726,176 @@ def on_delete(self, event):
...
+            return False
+        if added:
+            self.op_data.write_configuration()
+        return added
+
     def on_import(self, event):
-        print("Importing")
+       #
+        myfile = ""
+        mydlg = wx.FileDialog(
```
tatarize, line 1060: This is gui level. It might be nice to import things as a CLI process. And do the manipulations potentially from the commandline for what is essentially parsing files.

PR #2338, opened 2023-12-13
Add safeguard to signal_listener routines

jpirnay:


tatarize (changes requested):
This seems terrible. There's a different bug somewhere and certainly we should not be issuing signals when the kernel is shutdown if that's happening its indicative of a different bug which should be address correctly rather than through redundant code safeguarding. 

If nothing else this should be hacked into the kernel to pass it a function that validates that signals should be sent and maybe check `IsBeingDeleted()` there or something. Otherwise this seems like a lot of boilerplate code to try to hack around a bug that should be better documented and fixed.

jpirnay:
Not disagreeing it's ugly - but there's a not so fine line of something being ugly or horrible as in we are crashing way too often on exiting the program.  We have not understood the reason why things are still called in the shutdown phase of a program. If you figure it out can this PR (and that's the reason it's draft)

tatarize (dismissed):
Ill make a note to fix this. I thinks its an order of events problem but I guess I should figure out some way to better and more routinely cause it to happen and fix it correctly. Pretty sure I may have identified it, mostly after a shutdown is called it manages to file the request for a signal before the GUI dies, and ends up with a narrow band of time that that signal can still make it out while the kernel lives but the gui is dead.

tatarize (changes requested):
Probably fixed in a better though not very easy to validate way.

Issue #2332, opened 2023-12-09
[0.9.4b1] Suggested Image dithering improvements


Sophist-UK:
I normally work with b/w logos, so today was a bit unusual because I was working with a hi-res (1382DPI at the correct scale for the burn) greyscale drawing and trying to use the raster wizard to get the best dithering.

The first issue is performance. Just changing the dithering type on the ImageNode tab of the image properties, there was a ***MASSIVE*** difference in performance between Floyd-Steinburg (fast - c. 1s) and the rest (c. 30s). I am not sure what can be done to optimise these other algorithms, but I did note that CPU was low suggesting that Numpification could help. (Doing a newsy at a low sample size took almost an hour to complete. This sort of performance is a serious inhibitor to using MK Raster Wizards rather than e.g. Imag-R.)

One possible suggestion for making the UI faster is to apply it at the screen resolution and to the parts of the image shown - so if I am zoomed in to see what the Wizard does to individual pixels, then applying the wizard only to what I can see would make no difference. Of course if I then zoom out you would need to run it again - but you could probably calculate the full image as a background task. Similarly, if I am zoomed out to see the whole image, does it really make sense to dither it at full resolution rather than at the screen resolution (assuming that the screen resolution is lower and that the resulting displayed image looks the same.) 

The second set of issues are pure UI - on the Modification tab, if you have applied a wizard, then clearing it is done by selecting a "Set to None" value in the dropdown and clicking "Apply script", when this should probably be a "Clear" button at the bottom of the tab/window. And it might be nicer to have a small checkbox under the Apply Script button which switches between adding the selected script to the existing script or replacing the existing script with the newly selected one. (Personal preference: default is to replace, checkbox is set for replace and unset for add, so checkbox label would be "Replace existing script".) Also, when you apply a script the resulting tabs are to the left of the Modification tab when IMO they should be on the right - and when you clear a script, it switches to the ImageNode tab rather than remaining on the Modification tab.

(P.S. The names are still hopelessly useless - there is NOTHING to indicate to a user what the differences are or what they are equivalent to in other tools.)

Thirdly, having applied a script when you switch to e.g. the half-tone tab and move the sliders, the recalculation of the image happens when you have first dragged the slider to a different value not when you release the mouse button to drop the slider into a new position.  

[...]

PR #2346, opened 2023-12-16
Implement Numba Fast Dithers

tatarize:
#2332

Brings speeds of dithers down to on par with Pillow.

We fallback to to the non-dithered versions if `numba` is not installed.

`pip install numba`


tatarize:
Probably still needs more fulfilled checking but should have reasonable fallbacks.

PR #2354, opened 2023-12-20
Changed Rotary to Module

tatarize:
* Bit buggier than hoped.
* Had to use module plugin for the gui component.
* Gui bits didn't unregister since they were in their own context and thus registered in the kernel.
* Had to use `unregister` which is a really rare console command.
* Adding sub-contexts under lhystudios/rotary0 created additional devices which weren't actually real, they just enumerated strangely, or got created during launch as real devices.


tatarize:
Utterly Inferior to Service Delegate

PR #2356, opened 2023-12-20
Balor: Display input pins in config

jpirnay:


meerk40t/balormk/gui/balorconfig.py
```diff
@@ -97,13 +111,37 @@ def test_pin(self, value):
         else:
             self.context("red off\n")
 
+    def update_bit_info(self, *args):
```
tatarize, line 114: This entire thing belongs as a wait call in the controller. We are very likely waiting for a footpedal like event or something. There's some similar stuff in the controller/driver already.

tatarize: Ah, no. Okay I see what this is doing.

PR #2359, opened 2023-12-21
Settings backup

jpirnay:
if you use the ``create_backup`` parameter when instantiating a Settings object then it will maintain copies of the old config file before writing the new version: it will generate up to 6 previous versions of the file
``.bak``, ``.ba1`` .. ``.ba5`` - ``.bak`` the most recent one, ``.ba5`` the oldest

### DIFF
--- a/meerk40t/kernel/settings.py
+++ b/meerk40t/kernel/settings.py
@@ -1,4 +1,6 @@
 import ast
+import datetime
+import os
 from configparser import ConfigParser, MissingSectionHeaderError, NoSectionError
 from pathlib import Path
 from typing import Any, Dict, Generator, Optional, Union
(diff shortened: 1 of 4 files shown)


meerk40t/kernel/settings.py
```diff
@@ -19,11 +21,13 @@ class Settings:
     `write_configuration` is called.
     """
 
-    def __init__(self, directory, filename, ignore_settings=False):
+    def __init__(self, directory, filename, ignore_settings=False, create_backup=0):
         self._config_file = Path(get_safe_path(directory, create=True)).joinpath(
             filename
         )
         self._config_dict = {}
+        # 0 = No backup, 1 = keep daily, 2 = keep hourly, 3 = keep every minute
```
tatarize, line 29: Due to the way python stores strings, namely it creates a big table of them if they are explicit like "daily", "hourly", "minute" are basically equal to numbers. Since they will check `==` based on whether the objects are actually the same object if it's there. So really rather than numbers like this you can just use strings. (Though fstring versions of the same values only work at runtime, and so comparing fstrings is a longer algorithm.

Also, these are weird time periods.  And we should probably just make a copy of the last 2 or so previous settings files. Like you would only really want this if you messed them up, and then you'd have the previous ones.

jpirnay: Done

PR #2361, opened 2023-12-21
Placement improvements

jpirnay:
- Add ``placement_grid`` command (and use it in the tree context menu)
- Add ``place current`` to tree context menu
- Fix ``tree current`` operation property panel


tatarize (approved):

tatarize:
The placements themselves get queried by the call in `preprocess` and return a list of all the matrices that should be applied. 1 placement grid could do an entire grid of different placements. It could even calculate the size and fill the bed or some arbitrary size. It could do every other line off by half an object (needed to nest circles or hexagons).

PR #2367, opened 2023-12-25
Proper placement grid

jpirnay:
image

- [x] Single placement node with grid functionality 
- [x] Command support 
- [x] GUI support
- [x] Boustrodephon option
- [x] Line offset option (to allow honeycomb structure)


tatarize (approved):
Solid work. I do think this might be better as it's own object rather than `placement point` since it's not strictly a point it's a whole grid sort of thing.

PR #2376, opened 2023-12-29
Format painter

jpirnay:
A little tool to copy attributes from one element to another:

format_painter

### DIFF
--- a/meerk40t/gui/wxmmain.py
+++ b/meerk40t/gui/wxmmain.py
@@ -2,6 +2,7 @@ import datetime
 import os
 import platform
 import sys
+from copy import copy
 from functools import partial
 
 import wx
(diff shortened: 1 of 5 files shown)


tatarize (approved):
This should probably be in it's own file, rather than mostly in wxmain. I'm not sure this tool really warrants the prominent placement it's being given. This seems like a secondary tool.

meerk40t/gui/wxmmain.py
```diff
@@ -1143,6 +1310,27 @@ def register_options_and_choices(self, context):
...
+            {
+                "label": _("Paint format"),
+                "icon": icon_paint_brush,
+                "tip": _("First select your template, then every subsequent selection will apply the templates properties to the selected elements"),
+                "help": "basicediting",
+                "action": self.format_painter.on_click,
+                "identifier": "editpaint",
+                "toggle": {
+                    "label": _("Stop"),
+                    "help": "devicegrbl",
```
tatarize, line 1325: This can't be the correct help.

jpirnay: I was following the design patterns of office solutions where you have that all the time, I could add a secondary ribbon panel, that's not visible that would contain the old config without the button, so that you could pick the leaner version in the ribbon editor

PR #2377, opened 2023-12-29
Speed up font rendering

jpirnay:
The effective calculation time for the rendering of a vector font has been improved:
a) Use of LRU_CACHE to remove the need to reload the fontfile for every character changed / every recalculation
b) Migration from SVGElements path inside FontPath to pure GeomStr
Both together make a visible difference:
Before

font_before

After

font_after


tatarize (approved):
There's probably more especially since adding a letter is basically the same thing as the old path plus a couple contours.

This and allowing new-lines and alignment are things I've wanted but not enough to actually write.

PR #2378, opened 2023-12-29
Use element dragging instead of selection rectangle after some time

jpirnay:
Special case: if the user did not move the mouse within the first 0.5 seconds after
the initial mouse press then we assume a drag move.

drag

### DIFF
--- a/meerk40t/gui/scenewidgets/rectselectwidget.py
+++ b/meerk40t/gui/scenewidgets/rectselectwidget.py
@@ -1,10 +1,14 @@
 """
 Specifically draws the rectangle selection box and deals with emphasis of selected objects.
+Special case: if the user did not move the mouse within the first 0.5 seconds after
+the initial mouse press then we assume a drag move.
 """
+from time import perf_counter
 
-
+import numpy as np
 import wx
 
+from meerk40t.core.elements.element_types import elem_nodes
 from meerk40t.gui.scene.scene import (
     HITCHAIN_HIT,
     RESPONSE_CHAIN,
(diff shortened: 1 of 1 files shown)


meerk40t/gui/scenewidgets/rectselectwidget.py
```diff
@@ -162,51 +172,230 @@ def update_statusmsg(self, value):
...
+
+                self.scene.request_refresh()
+                self.scene.context.signal("select_emphasized_tree", 0)
+            else:
+                def shortest_distance(p1, p2):
+                    """
+                    Calculates the shortest distance between two arrays of 2-dimensional points.
+                    """
+                    # Calculate the Euclidean distance between each point in p1 and p2
+                    dist = np.sqrt(np.sum((p1[:, np.newaxis] - p2) ** 2, axis=2))
```
tatarize, line 246: Somebody's getting fancy.

tatarize: Most points are actually complex numbers. You could just subtract them and take the absolute power.

So if a and b are points in complex numbers the Euclidean distance is `np.abs(a - b)` even groups of points should basically be like that too. This code man actually only need to be `np.abs(p1 - p2)` assuming we're dealing with complex numbers here which most points are.

It's notably sort of complex to get a grasp on, but worth the confusing since it makes things a lot easier.

tatarize (approved):
I'm not sure drag move isn't just preferred outright.

jpirnay:
> I'm not sure drag move isn't just preferred outright.

Shrug,  we could even switch the logic in the "undefined" block

PR #2392, opened 2024-01-04
Geomstr simplify

jpirnay:
- Replace the existing simplify method of geomstr (which was based on the old svg path paradigm anyway) with an implementation of the Ramer-Douglas-Peucker algorithm to reduce the amount of points in a polyline section of the geometry.
- Changing the ``simplify`` command to use this
- Fixing a bug within ``Geomstr.as_contiguous`` 
- Kerf offset calculation will automatically call the simplify method (with minimal tolerance to just reduce the amount of points a bit)
- Added an optimisation option to apply polyline simplification. By default off

### DIFF
--- a/meerk40t/core/elements/offset_mk.py
+++ b/meerk40t/core/elements/offset_mk.py
@@ -368,8 +368,16 @@ def offset_path(self, path, offset_value=0):
         interpolation=500,
     )
     if p is None:
-        p = path
+        return path
+    g = Geomstr.svg(p)
+    # We are already at device resolution, so we need to reduce tolerance a lot
+    # Standard is 25 tats, so about 1/3 of a mil
+    g.simplify(tolerance=0.1)
+    p = g.as_path()
+    p.stroke = path.stroke
+    p.fill = path.fill
     return p
+
 
 
 def path_offset(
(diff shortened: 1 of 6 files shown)


meerk40t/core/elements/offset_mk.py
```diff
@@ -368,10 +368,18 @@ def offset_path(self, path, offset_value=0):
         interpolation=500,
     )
     if p is None:
-        p = path
+        return path
+    g = Geomstr.svg(p)
+    # We are already at device resolution, so we need to reduce tolerance a lot
+    # Standard is 25 tats, so about 1/3 of a mil
```
tatarize, line 374: This differs between devices. Galvos can have a lot more precision depending on their lenses. Though this is just 0.1 native units or so, so it's fine.

meerk40t/core/elements/shapes.py
```diff
@@ -740,36 +740,47 @@ def recalc(command, channel, _, data=None, post=None, **kwargs):
         self.signal("refresh_scene", "Scene")
         self.validate_selected_area()
 
+    @self.console_option(
+        "tolerance", "t", type=float, help=_("simplification tolerance"),
```
tatarize, line 744: Could just add a `default=25` here rather than getting None and using that to set it. It's supposed to be required to pass that to the function if the default it set.

PR #2395, opened 2024-01-04
Fontwelding

jpirnay:
- Allow "welding" of overlapping characters
- Allow kerning adjustments
- A more generic way to reload meerk40t specific attributes from svg files


tatarize (approved):
Generally fantastic.

PR #2397, opened 2024-01-05
Inkscape Layer-Support / Accelerate loading

jpirnay:
- Recognising inkscape-layers and loading them into regmarks
- Acceleration of file load for designs with a lot of elements inside groups and subgroups


tatarize (approved):
I might make  self.mk_params into  self._mk_params since it's purely internal but otherwise a very solid load time speedup.

PR #2400, opened 2024-01-08
Alarm reset

jpirnay:
Allow an optional parameter to reset the device too when clearing an alarm status as this seems to be required for some devices.
grafik


tatarize (approved):
Probably best to go in before the beta though it would void the tiger test on GRBL.

PR #2401, opened 2024-01-08
Add project sanity warnings

jpirnay:
Provides a visual indicator if mk thinks a project has issues:
a) unburnt elements (unassigned or inactive ops)
b) elements that are partially or fully outside the bed
c) operations that are too fast (grbl + lihuiyu for the time being, thresholds still too forgiving)
d) rasters that are too close to the edge so a deacceleration movement might still hit the rails

image


tatarize (approved):
This should probably hold off with regard to merging until after the beta. I can probably speed that up. Also, it's weird that this is mostly entirely a gui check, and almostly entirely stored in wxmain. But, gui checks seem pretty straight forward. This also seems to be another version of your `warning` code for various drivers exceeding a speed, etc. If replacement it should remove. Not sure if it's that different.

PR #2409, opened 2024-01-16
MWindow default window size

jpirnay:


tatarize (changes requested):
image

Several reset windows ended up looking like this super-tiny thing or whatever. That's not really possible for approval. Also scene has a super tiny view of the bed in the upper-left on default startup.

jpirnay:
Issues reolved

tatarize (approved):
Opened every window, they all seemed pretty good.

PR #2422, opened 2024-01-23
Convert regular elem text to a elem path

jpirnay:
To work properly the corresponding vector font needs to be already available, otherwise a black rectangle appears as a placeholder.


jpirnay:
Included in PR #2430

PR #2429, opened 2024-01-27
Debug: Widget Icon Viewer

tatarize:
Creates a minor scene of all the icons and their names within the scene.

image

This only works if `set debug_mode True` has been set and activates with console command `widget_icons`


jpirnay:
Nice. Just fyi: Panes - Debug - Icons allows the inspection of a single icon at larger sizes

PR #2433, opened 2024-01-30
Fix parsing of ttf:cmaps

tatarize:
If a matching cmap signature was found it would try to parse and exit. However, in many cases if this cmap used a format that was not correctly parsed it would be better to continue looking for valid cmap formats.


tatarize:
image

PR #2455, opened 2024-02-07
Feature request: abort, singleinstance

jpirnay:
- Add Quit option on erro dialog to quit meerk40t immediately
- Single Instance checker (if option is set)


tatarize (approved):
Seem fine with the option. Not entirely sure the utility of this. You can run two different copies of meerk40t with profiles in the CLI so they wouldn't overlap and could control different lasers if there's otherwise an issue doing that.

PR #2461, opened 2024-02-15
GRBL over WebSocket support implementation

rOzzy1987:
Many of DIY laser engravers use a simple ESP01 to make them wireless (including mine)
This app seems promising but it couldn't connect to WebSockets, so I added this option.

FYI, this is my first encounter with Python, so I'm happy to learn. Don't hesitate to point out the no-nos I just did.

### DIFF
--- a/meerk40t/grbl/controller.py
+++ b/meerk40t/grbl/controller.py
@@ -316,6 +316,10 @@ class GrblController:
             from meerk40t.grbl.tcp_connection import TCPOutput
 
             self.connection = TCPOutput(self.service, self)
+        elif self.service.permit_ws and self.service.interface == "ws":
+            from meerk40t.grbl.ws_connection import WSOutput
+
+            self.connection = WSOutput(self.service, self)
         else:
             # Mock
             from .mock_connection import MockConnection
--- a/meerk40t/grbl/device.py
+++ b/meerk40t/grbl/device.py
@@ -25,6 +25,7 @@ class GRBLDevice(Service, Status):
     def __init__(self, kernel, path, *args, choices=None, **kwargs):
         self.hardware_config = {}
         self.permit_tcp = True
+        self.permit_ws = True
         self.permit_serial = True
 
         Service.__init__(self, kernel, path)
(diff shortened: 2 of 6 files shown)


tatarize:
Codewise, it seems really close. Does this work for you? Like connects and controls your device with your given changes?

Error wise, there are few. The extra choice panel for the `ws` uses the same names `address` and `port` as the TCP connection and thus will share the same values, between the tcp and webserver connections. The extra requirement also needs to be added to `setup.py`.

The `read` code for websocket is probably fine and shouldn't suffer any real issues though the protocol there is technically `UTF-8` rather than the `latin-1`. The default protocol for GRBL doesn't tend to spit out any high-ascii values, though the serial would sometimes disconnect and thus be a 0xFF value and would break the UTF-8 encoding (I assume that's not an issue websockets would hit, but it's harmless copy-pasta).

We often also permit softer imports, for lesser used features. So we often do a import check and then if that import failed we disable that functionality, But, for the direct interfacing stuff like pyserial and pyusb that stuff tends to just be included directly and websocket-client is quite reasonable as an include.

---

All in all this seems like a solid PR and I'd expect it to be live by 0.9.5 beta 1.

tatarize (approved):
Should probably be merged after next full release. We may need to acquire some hardware and see if we can build in the same functionality fairly easily since the protocol isn't that large. Connect as websocket becoming a checkbox in that case.

tatarize:
Also, any more details on the project. For example, if I wanted one of these to test with, how could I vaguely retrace your steps? Does this flash full GRBL or FluidNC or something? And if not is there some customized software requirement or preference here?

I seem to recall I have a project vpype which basically does generic ascii data (https://github.com/plottertools/vpype-gcode). The methods are simple in that it simply writes out the data in profile determined ascii.

---

Also, would this stuff be common for other devices? For example we have a com-port command that sends generic data to a com-port. Would something like that be wanted or needed for websockets? Like, after this part of the project is finished, execute this command sending data X to this particular TCP address and if we get back Y data then we were successful if not we fail and abort the process.

In the original use case it was cutting and kicking different pieces of wood for a passthrough, but it seems like it could do basically anything.

PR #2465, opened 2024-02-17
Deal with nop segments

jpirnay:


tatarize (approved):
NOP is rare, but yeah, if they exist they should be treated as if they don't.

PR #2474, opened 2024-03-17
Add in Cylinder Correction for Galvo Lasers

tatarize:
See: https://docs.lightburnsoftware.com/galvo/CylinderCorrection.html

The basic idea is to a mesh sort of cylinder conversion along the x and y axis for Galvo devices to correct for x or y distortion along a narrow cylinder shape. This is very similar to the image.flatrotary feature we've had for years (but never really used).

These settings include input-> output point conversions for the galvo device.

* The math is not yet sorted.
* The use of the math is not yet hooked up.
* An icon is needed for the feature.

image


tatarize:
Requires disable and re-enable to work update, has some other issues. I'm not sure if distance is ever used since I use the known lens size.

PR #2499, opened 2024-06-03
Fix grblcontrol

jpirnay:


Alb673:
Hello, could you advise how to upgrade to latest version please. I tried pip install meerk40t[all] but i got this message
)Defaulting to user installation because normal site-packages is not writeable)
I am on version 09.3001. Tried to connect it with lightburn using the GRBL command in console, but it doesn't seems to work .

Thanks

jpirnay:
``pip`` will only refer to full versions - the fix you are looking at is included in the latest beta:
https://github.com/meerk40t/meerk40t/releases/tag/0.9.4020
In the ``Asset`` section you will find the binaries/code to use:
grafik

Alb673:
Hi, thanks for the quick reply.
I tried to run the newer version, but still having issues. In the Meerk40t console with GRBLcontrol command I got this:
[16:11:00]     Window opened: Console
[16:11:10] GRBLcontrol
[16:11:10]     GRBL Mode.
[16:11:10]     TCP Server for GRBL Emulator on port: 23
[16:11:10]     Will translate incoming to device: LihuiyuDevice

I set the machine manually as GRBL and chose USB, since it is connected with a USB cable.
Now looking at the console should select Ethernet/TCP then and not USB? If I do that is asking for a device IP address which I do not have... I wonder if it's more difficult since I am on Linux?

Cheers

jpirnay:
You need to pick a grbl via TCP device in lightburn. If it's on the same machine then add 127.0.0.1 as address

Alb673:
Well I just tried you advice, Selected Ethernet/TCP, and used 127.0.01, but now in console I get: 
[19:46:54]     Window opened: Console
[19:47:02] GRBLcontrol
and hangs there

So it doesn't look like it's even connecting...
I am on Lightburn v1.3.01, maybe it's too old version?

jpirnay:
For the avoidance of doubt:
- you setup your k40 within meerk40t as usual - via USB - please check that it works correctly (that's the k40 controller)
- you start ``grblcontrol`` in meerk40t (requires 0.9.5b2)
- you need to define a grbl-lpc laser within lightburn, this requires an ip address, that is the address of your k40 controller, if both k40 and lightburn run on the same machine, that would be 127.0.0.1

Alb673:
> For the avoidance of doubt:
> 
> * you setup your k40 within meerk40t as usual - via USB - please check that it works correctly (that's the k40 controller)
**Yes, Meerk40t is installed on laptop, and working with the machine via USB.**
> * you start `grblcontrol` in meerk40t (requires 0.9.5b2)
**The command is issued in Meerk40t v0.9.4020 console, that's what I was able to run from the app image in the link you previously supplied... How do I run the 0.95b2 then? I also tried to download the "source code.zip" but it still showing as 0.9.4020**
> * you need to define a grbl-lpc laser within lightburn, this requires an ip address, that is the address of your k40 controller, if both k40 and lightburn run on the same machine, that would be 127.0.0.1
**I first selected GRBL, now I just done it again selecting GRBL-LPC. 
Both Lightburn & Meerk40t are installed on the same laptop, connected with the K40 via USB** 

Thanks

jpirnay:
0.9.5b2 == 0.9.4020 - so you have the right version

Alb673:
Hi [jpirnay](https://github.com/jpirnay)
Following on the previous post. I just downloaded the app image from here: https://github.com/meerk40t/meerk40t/releases

Made it executable, started it and it shows up as v0.9.4042... really strange. Am I missing something? 
I'm gonna see if I can install Lightburn on my Win table and see If I am having the same issues.

PR #2506, opened 2024-06-13
Effects Toolbar Button and Default Effects Settings

jondale:
[Screencast from 2024-06-13 14-56-31.webm](https://github.com/meerk40t/meerk40t/assets/1604942/7824cd57-bc27-465f-ab07-e0e76492c6d3)

This PR makes two effective changes.

It changes the hatch button on the toolbar to an Effects Button that is populated by registered effects (currently wobble and hatch).

Additionally, it adds the ability to set default values for these effects that are device specific by adding a new Effects tab to each device config dialog.

One current weirdness that is perhaps added by this pr is that the preview in the hatch properties dialog is strange until you manually change it.  It might be an issue in conversion from str to Angle in that preview.  The hatch applies correctly to the object(s).

### DIFF
--- a/meerk40t/core/elements/shapes.py
+++ b/meerk40t/core/elements/shapes.py
@@ -386,7 +386,7 @@ def init_commands(kernel):
 
         if interval is None and hasattr(self.device, "effect_wobble_default_interval"):
             interval = getattr(self.device, "effect_wobble_default_interval")
-        elif distance is None:
+        elif interval is None:
             interval = "0.5mm"
 
         wtype = wtype.lower()
--- a/meerk40t/core/elements/shapes.py
+++ b/meerk40t/core/elements/shapes.py
@@ -307,6 +307,22 @@ def init_commands(kernel):
         """
         Add an effect hatch object
         """
+
+        # I'd like to get these settings here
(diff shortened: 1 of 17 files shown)


meerk40t/core/elements/shapes.py
```diff
@@ -361,6 +378,17 @@ def effect_wobble(
...
             wtype = "circle"
+
+        if radius is None and hasattr(self.device, "effect_wobble_default_radius"):
+            radius = getattr(self.device, "effect_wobble_default_radius")
+        elif radius is None:
+            radius = "0.5mm"
+
+        if interval is None and hasattr(self.device, "effect_wobble_default_interval"):
+            interval = getattr(self.device, "effect_wobble_default_interval")
+        elif distance is None:
```
jpirnay, line 389: Wrong variable name will crash

jpirnay (changes requested):
invalid variable name, will crash

meerk40t/core/elements/shapes.py
jondale: updated.

jpirnay (approved):

PR #2538, opened 2024-08-02
Redblue classification

jpirnay:
Simplified classification of red/blue elements

Unticked: Classify red/blue elements into rasters too
Ticked: Prefer a single cut/engrave operation (needs transparent fill)


jpirnay:
On second thoughts I don't think this is needed, the existing logic should cover it.

PR #2549, opened 2024-08-15
M3 extension to add PWM support

jpirnay:
First attempt to support m3nano PWM


jpirnay:
Seems to crash with older boards...

PR #2586, opened 2024-09-19
Adding a 3D image operation 

jpirnay:


jpirnay:
Overhead for a cutcode node of its own just for a couple of lines is too big - will be replaced by branch depthmap and PR #2588

PR #2601, opened 2024-09-27
Automatic ID generation of nodes

jpirnay:
- Automatic assignment of id on element creation
- Assignment of a derived id on node copying

Could probably be done better, as there is no mechanism in place to prevent duplicate ids.
Not to be implemented before 0.9.5b8


jpirnay:
Needs more speedups before it can be merged - too much overhead on node creation

PR #2612, opened 2024-10-02
Change driver.py and geomstr.py to do interpolation for a line.

mnoriaki:
This request is needed for Moshiboard based laser cutter since Moshiboard do not interpolate as expected as in the Meerk40t codes.

Moshiboard does not correctly cut an oblique line like in following image. 
Screenshot from 2024-10-02 13-40-22

When Moshiboard is requested to cut a oblique line, it cuts with two lines. One is 45 degrees angled line and the other is horizontal or vertical line. The following image shows this.
IMG_0577
Above shows the issue and below shows the result with the changes included in this pull request.

### DIFF
--- a/meerk40t/moshi/driver.py
+++ b/meerk40t/moshi/driver.py
@@ -235,13 +235,7 @@ class MoshiDriver(Parameters):
             self.settings.update(sets)
 
             if segment_type == "line":
-                interp = self.service.interp
-                g.clear()
-                g.line(start, end)
-                for p in list(g.as_equal_interpolated_points(distance=interp))[1:]:
-                    while self.hold_work(0):
-                        time.sleep(0.05)
-                    self._goto_absolute(p.real, p.imag, 1)
+                self._goto_absolute(end.real, end.imag, 1)
             elif segment_type == "end":
                 pass
             elif segment_type == "quad":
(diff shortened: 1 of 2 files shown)


meerk40t/tools/geomstr.py
jpirnay, line 1: as_equal_interpolated_points is expected to provide a sequence of line segments, wher only non-line artifacts will be interpolated.
Your approach would lead eg for a rectangle that is constituted of 4 lines to end up in 4xinterpolation segments.
So I wouldn't make the change here. But stick to the moshi driver instead.

meerk40t/moshi/driver.py
jpirnay, line 1: As indicated in the comment to geomstr.py you should isolate your changes here. As far as I understand your approach you are interpolating any line by a rectangular zig zag? While this generally true, afaics you will end up even for a perfectly horizontal / vertical line with a staggered line, ie instead of one line from (0, 0) to (100, 0) you will generate 100 segments (0,0) to (1,0), (1,0) to (2,0) and so on.
This probably deserves some case splits to only deal with horizontal/vertical lines.

jpirnay (changes requested):
Thank you for your bug report and your contribution. Good to see that Moshi-lasers are still around.
I left a couple of comments in the files for your consideration. A merge without amendments would cause a lot of unwanted side effects, so I would recommend isolating the changes.
Something along the lines of this fragment might be a another way of doing this. I would probaly consider isolating this this in _goto_absolute, if you comapre the new position to the new one:
```
import numpy as np

def interpolate_polyline(points, resolution):
    def is_valid_segment(p1, p2):
        dx, dy = p2[0] - p1[0], p2[1] - p1[1]
        return dx == 0 or dy == 0 or abs(dx) == abs(dy)

    def interpolate_segment(p1, p2, resolution):
        dx, dy = p2[0] - p1[0], p2[1] - p1[1]
        steps = max(abs(dx), abs(dy)) // resolution
        x_steps = np.linspace(p1[0], p2[0], steps + 1, dtype=int)
        y_steps = np.linspace(p1[1], p2[1], steps + 1, dtype=int)
        return list(zip(x_steps, y_steps))

    new_points = []
    for i in range(len(points) - 1):
        p1, p2 = points[i], points[i + 1]
        if is_valid_segment(p1, p2):
            new_points.append(p1)
        else:
            new_points.extend(interpolate_segment(p1, p2, resolution))
    new_points.append(points[-1])
    
    return new_points

# Example usage
points = [(0, 0), (3, 4), (6, 4), (6, 8)]
resolution = 1
new_points = interpolate_polyline(points, resolution)
print(new_points)
```

mnoriaki:
Thank you for your comments. 

I took quik-hack approach and found that self._goto_absolute(x, y, on & 1) at the line 398 in meerk40t/moshi/driver.py command Moshiboard to cut lines and arcs. Since Moshiboard does not have Zingl-Bresenham line draw algorithm to cut a line, we need to cut a line into segments. So I changed two files to treat a line as an arc. 

I don't intend to change a horizontal or vertical line to small zig-zag lines.  Sorry if my change do that. I just thought such a line will segmented into small horizontal or vertical lines and there should be no harm theoretically since arcs are already segmented as such.

I don't know if the change for the line 238 in moshi/driver.py I proposed is really needed.

mnoriaki:
Thank you for your comments.

I took quik-hack approach. I din't intend to change a horizontal or vertical line to small zig-zag lines. 

How about changes like attached file? It only changes moshi/driver.py.

[20241003.txt](https://github.com/user-attachments/files/17240742/20241003.txt)

mnoriaki:
Oh, please just ignore comment line below
+                    # use Zingl-Bresenham line draw algorithm

jpirnay:
> Thank you for your comments.
> 
> I took quik-hack approach. I din't intend to change a horizontal or vertical line to small zig-zag lines.
> 
> How about changes like attached file? It only changes moshi/driver.py.
> 
> [20241003.txt](https://github.com/user-attachments/files/17240742/20241003.txt)

Looking good to me, did you test it ?

mnoriaki:
Yes, I have tested it.
IMG_0578

jpirnay (approved):
Looking good, i might need to change some unit tests to take this into consideration, but that shouldn't prevent the change from being approved-
Thank you for your contribution.

Issue #2617, opened 2024-10-03
0.9.5b7: Can we resequence the context menu of File nodes:?


Sophist-UK:
Here is the file node context menu:
![image](https://github.com/user-attachments/assets/a47bc21a-14ad-4e29-b585-8a19bdac6444)

Some of these are common with a Group node context menu:
![image](https://github.com/user-attachments/assets/a13cd653-64d7-4e2d-8e8d-ab2167724ac4)

To make the UI of these two menus more consistent can we please:
1. Add any entries on the Group menu that still make sense to the File menu?
2. Move any entries unique to the file menu to the top of the menu sequenced in a sensible logical order, with a dividing line separating them from those that are also on the group menu?

Thx.

Also in terms of more general consistency, there are common or similar types of ops on all context menus i.e. 1. Group A: Assign op, duplicate, flip, scale, rotate; 2) Group B - actions to delete or remove. Can we have dividers around these groups and have them in the same order towards the bottom of the menu with more unique stuff above?

jpirnay:
That makes sense, the context menu entries have grown from a few to quite a lot and could definitely merit from a more logical sequence and optical separation. May I be so bold to ask you to look at the other cases too, so that we can bring some order into them as well. Thank you.

Sophist-UK:
> May I be so bold to ask you to look at the other cases too, so that we can bring some order into them as well.

You may certainly be that bold, however sorry but I am not a developer submitting PRs or documentation improvements any more - and until there is sufficient hard evidence that the completely toxic developers clique has been abolished for good and that code and documentation contributions from other members of the community will be welcomed and valued, this will continue to be the case. In the mean time, my contributions are now limited to raising issues and making suggestions, and will continue to be so for the foreseeable future.

Sophist-UK:
It looks like this one might actually be fixed rather than closed to get rid of it. Apparently now a rare occurence.

PR #2619, opened 2024-10-04
Sort and separate tree/node context menu according to tags

jpirnay:
The existing methodology of creating a structure inside the context menu became tiresome to maintain / to amend.

So a new grouping tag has been added to the context menu entries that will determine its position,

## Comparison
| New            | Old            |
| ------------ | ----------- |
|  |  |
|  |  |
|  |  |
|  |  |


jpirnay:
Should address issue #2617

Issue #2625, opened 2024-10-07
[Bug]: Mac OS 13.7 (22H123) - INTEL - not loading 


skasilov:
### Summary Description

Not loading/starting on Mac OS 13.7 (22H123) - INTEL

### Additional Details

Not loading/starting on Mac OS 13.7 (22H123) - INTEL

### MeerK40t Version

betta 6

### MeerK40t Type

Executable

### Your Operating System

Mac (before M1/Bigsur)

joerlane:
Do you see anything when you try to open this version of MeerK40t? Does it simply bounce in the dock and then stop; or does it show anything, or an error of some sort?

You can launch it from terminal by calling the executable directly from inside the .app bundle; that should provide a more specific error message. If MeerK40t is inside the Applications folder, this command will launch it from terminal: /Applications/MeerK40t.app/Contents/MacOS/MeerK40t

skasilov:
Good afternoon! 
Thank you so much for your response and willingness to help! 
This is very valuable for us! 
I'm not big expert  for Mac, but I'm a big fan of your program! 
And I really want to continue working with her as I see great potential in her! 
And so what can I say! 
When launching the latest beta version, the program "jumps" in the dock and even starts (a black dot appears that the program is running) 
At the same time, there are no mistakes! 
It's just that the program windows don't appear! 
If you click on the program icon in the dock, you can only close the program "forcibly"
I'll try to run it through the terminal tomorrow and let you know! 
By the way, I also can not launch the previous two beta versions!

Sincerely Sergey
+31617349421
Catharina van Rennesstraat 8,
2511AT, 's-Gravenhage
https://www.linkedin.com/in/sergey-kasilov
https://www.innerquill.com <https://www.innerquill.com/>
https://www.instagram.com/innerquill.project

jpirnay:
Just for the avoidance of  doubt: which was the last release that worked for you?

joerlane:
I uploaded Beta 6 Mac builds just for your testing. Not likely to make a difference if you say the last version working is Beta 4; but worth a try to narrow things down. 

Curious about launching from the terminal; but not positive it will provide an error if you say MeerK40t is actually opening crashed and not showing any windows. Still worth a try though as I don't have any other quick ideas.

Unfortunately I don't know anyone off hand with an Intel machine running that version of macOS (or even anything close). I moved to apple silicon during macOS 11. 

[...]

PR #2633, opened 2024-10-10
Foolproof column width

jpirnay:
## Summary by Sourcery

Improve column width management by using GetColumnCount() instead of ColumnCount and add load/save functionality for column widths in various GUI components.

Enhancements:
- Replace direct access to ColumnCount with GetColumnCount() for improved robustness in column width handling.
- Add exception handling in adjust_last_column to prevent errors during column resizing.


jpirnay:
Even if I did not recognize a real error , I tried to foolproff stuff to address issue #2625

PR #2646, opened 2024-10-15
Improved / more logical drag handling

jpirnay:
Addresses the concerns raised in Discord about the visual (non-)representation of the drag and drop operation:
- By default drag and drop within the same hierarchy will insert the dragged node *below* the drop node
- A drag to the root branches will insert the dragged node on top of the list, if it's a a top-level node (before it was added to the end)
- Visual indicator if you can drop the dragged nodes on the target node


jpirnay:
@sourcery-ai review

PR #2651, opened 2024-10-18
Fix UI theming for skinned desktops

Laserology:
This PR adds proper foreground and background theming to the program to fit in with an existing system's theme, and also fixes an issue with dark theme systems. That also includes icons, which now properly contrast with their backgrounds.

This does not cover everything, but it does fix most windows and panels.

This PR also changes the default color pallet to be more appealing. I suggest the devs build this, run it, and see what they think before accepting it. It is subjective, but overal improves readability and lessens the harsh contrasts.

Here is a screenshot of the main window + the about page on a dark theme system after my changes:
screenshot

## Summary by Sourcery

Improve UI theming to align with system themes, particularly for dark mode, and update the default color palette for better readability.

Bug Fixes:
- Fix UI theming issues for dark theme systems by adjusting foreground and background colors to ensure proper contrast.

Enhancements:
- Update the default color palette to improve readability and reduce harsh contrasts.


jpirnay:
I probably would have chosen a more central location to apply the color changes, but I appreciate any effort to help address our Achilles heel = UI

jpirnay (approved):

Laserology:
> I probably would have chosen a more central location to apply the color changes, but I appreciate any effort to help address our Achilles heel = UI

Where would such setting be applied? To be honest, i have never worked with wxpython a single time prior to this, and rarely ever with python in general. Only reason i could navigate well was due to prior experience in other languages. Does WX python have a 'default' property i can set somewhere? I looked once before but i couldn't find any reference about it.

PR #2656, opened 2024-10-22
Simplify device listings and add longer ray5

Laserology:
This PR simplifies the device listings to separate types better, and also adds the Longer Ray5 to the list - making it easier for newcomers to get set up. Most types of devices not under a specific product or category have been moved into a generic or firmware named category.

I've mostly removed the redundant "Family Priority" entry as we can sort by name instead to form more concise grouped listings.

## Summary by Sourcery

Simplify device listings by reorganizing categories and removing redundant attributes, and add support for the Longer Ray5 device.

New Features:
- Add support for the Longer Ray5 device in the GRBL plugin, allowing users to configure and use this specific diode laser model.

Enhancements:
- Simplify device listings by removing the 'family_priority' attribute and sorting devices by 'family' and 'priority' instead.
- Reorganize device categories to better reflect their types, such as renaming 'Older CO2-Laser' to 'Newly CO2-Laser' and 'Generic' to 'Generic Diode-Laser'.


jpirnay (changes requested):
Your method will create a different sequence of devices if a different language than English is chosen. That was the reason why we added a specific sort key to ensure a defined sequence regardless of the langage settings. So I would advise to keep the family_priority key.

Laserology:
> Your method will create a different sequence of devices if a different language than English is chosen. That was the reason why we added a specific sort key to ensure a defined sequence regardless of the langage settings. So I would advise to keep the family_priority key.

Not that we have translations now - but how about we sort names first, and then apply translations afterwards, effectively keeping the ordering? Re-sorting only gets triggered manually it appears.

jpirnay:
- Probably a bit of a personal preference - Labels are not necessarily providing a good sorting criteria - so is a **G**RBL better placed before a **K**40 or a **L**ihuiyu device? 
- A translation afterwards is a bit messy, as we have a couple of external tools for the translators that look for a ``_("Something to translate")`` occurence  (specifically for the ``_(`` invocation).
- We have translations for the family types:
grafik
grafik
grafik

Laserology:
> - Probably a bit of a personal preference - Labels are not necessarily providing a good sorting criteria - so is a **G**RBL better placed before a **K**40 or a **L**ihuiyu device? 
> - A translation afterwards is a bit messy, as we have a couple of external tools for the translators that look for a ``_("Something to translate")`` occurence  (specifically for the ``_(`` invocation).
> - We have translations for the family types:
> grafik
> grafik
> grafik
> 

I was not aware, thanks for pointing that out.

Family priority still feels quite clunky.

After translating, do the internal names still remain the same? Or do they change?

How about, when translating, adding an extra field 'family_translated', so that the normal family name can still be used for sorting? For displaying, just check if translated exists for that device, if so, use that for display name, if not, then use the normal family name.

Laserology:
As for order of names, it's reverse alphabetical, so grbl/generic goes AFTER K-series and others

Laserology:
> > - Probably a bit of a personal preference - Labels are not necessarily providing a good sorting criteria - so is a **G**RBL better placed before a **K**40 or a **L**ihuiyu device? 
> > - A translation afterwards is a bit messy, as we have a couple of external tools for the translators that look for a ``_("Something to translate")`` occurence  (specifically for the ``_(`` invocation).
> > - We have translations for the family types:
> > grafik
> > grafik
> > grafik
> > 
> 
> I was not aware, thanks for pointing that out.
> 
> Family priority still feels quite clunky.
> 
> After translating, do the internal names still remain the same? Or do they change?
> 
> How about, when translating, adding an extra field 'family_translated', so that the normal family name can still be used for sorting? For displaying, just check if translated exists for that device, if so, use that for display name, if not, then use the normal family name.

What about a mapping function for priority? Instead of defining family priority in the device entry, it could be sorted based on family name inside of a map, like so:

Family 1 -> priority 5
Family 2 -> priority 30
[ Etc... ]

jpirnay:
> What about a mapping function for priority? Instead of defining family priority in the device entry, it could be sorted based on family name inside of a map, like so:
> 
> Family 1 -> priority 5 Family 2 -> priority 30 [ Etc... ]

I think that could make sense.

Laserology:
> > What about a mapping function for priority? Instead of defining family priority in the device entry, it could be sorted based on family name inside of a map, like so:
> > Family 1 -> priority 5 Family 2 -> priority 30 [ Etc... ]
> 
> I think that could make sense.

That should do it.

meerk40t/grbl/plugin.py
```diff
@@ -175,6 +170,36 @@ def plugin(kernel, lifecycle=None):
                 ],
             },
         )
+        kernel.register(
+            "dev_info/grbl-longer-ray5",
+            {
+                "provider": "provider/device/grbl",
+                "friendly_name": _("Longer Ray5 (GRBL)"),
+                "extended_info": _("Longer-branded 5w/10w/20w grbl diode laser.\nMake sure you verify your bed size! This machine has several upgrade kits."),
+                "priority": 21,
```
jpirnay, line 179: That should be removed, shouldn't it?

jpirnay (approved):

meerk40t/grbl/plugin.py
Laserology: If you're talking about the preset for the Longer Ray5, no, i had intentionally added it as a preset. It's basic but helps people getting set up, some might not necessarily know what "GRBL" means in the context of laser engravers, and will assume their device isn't supported. (Example being my business partner)

Do you personally think it's a bad idea to include presets?

PR #2697, opened 2024-10-30
Update version information

jpirnay:
## Summary by Sourcery

Chores:
- Update the application version number from 0.9.5.1 to 0.9.5100.


jpirnay:
Unittests failed because I  added too many parallel open pull request and that will crash the tests if they are already running for another pr.

Issue #2330, opened 2023-12-07
[0.8.12] Raster at DPI 1000 shows ridges


Sophist-UK:
Burning an acrylic raster at DPI 1000 I get ridges.

I tried a burn with the laser off and my finger on the Y-rail to sense the steps and it seemed to be doing 3 sweeps with a step and then 1 without.

Is this a quirk of the m2 nano?

At DPI 500, it steps every time but I cannot tell whether they are all step 2 or not.

tatarize:
The movement is very small, but it should be moving. If you're worried you can dump the LHYMICRO-GL code and look at it. Should tell you either by specifying `G001` at the start or having movements `La` or `Ra`.

Sophist-UK:
I have no idea how to "dump the lhymicro-gl code", nor how to interpret it.

I am also unsure about whether this is a MK or a controller issue (not the least because I have no idea what the control codes are for rastering on the m2 nano). The raster doesn't end up 3/4 of the size it should be in the Y direction, so overall the number of steps seems right and the missing steps are made up somehow.

So it feels either like some sort of rounding error where the y-position results in 0 steps and then 2 steps or a some sort of in-baked hardware issue.

The m2 nano is "dumb" so it seems unlikely that it is a rounding error inside the controller. It might be a rounding error inside MK, but that also seems unlikely (though in the past I have diagnosed similar rastering issues in #1921 that were rounding/offset issues in MK, so I guess this is still possible).

Looking at hardware possibilities, when I unlock the rails, and move the head by hand there are also slight but extremely regular ups and downs in resistance. I am unsure whether these are magnetic effects inside the stepper motors (the regularity is of the order of (maybe) 0.5mm (I haven't attempted to measure it accurately) but it seems too big for steps (which are c. 0.025mm), or something to do with the belt ridges but it seems too small for these (which are c. 2mm apart), or it could be something else entirely. But they feel far too regular to be something sticking and in any case it is the same on both axes which also rules out a hardware glitch IMO. 

[...]

Issue #1921, opened 2023-05-19
[0.9x][0.8.4] Image raster at v. low DPI burns alternate lines offset on Lhy but not in Simulation 


Sophist-UK:
As testing the focus for a new head / lens, I rastered an image at v. low DPI (50). The raster lines do not overlap at this DPI so it is easy to see what is happening.

Here is the simulation:
![image](https://github.com/meerk40t/meerk40t/assets/3001893/2c0cac34-26ec-4e45-9980-ac5acc7d4778)

Here is an image of the actual burn:
![1684501207688](https://github.com/meerk40t/meerk40t/assets/3001893/2791a85a-b1ec-45e7-8e55-3b395a4847da)

It isn't consistently the case, but it appears that lines are sometimes longer in the direction of the head movement. Most notable on the vertical lines.

I am not sure whether or not this is the same as #1900 (i.e. this is not a crosshatch burn nor is it Lightburn+Ruida Emulation).

P.S. It might be an improvement i.e. be more realistic if the simulation picture used a fixed physical height for raster rows when DPI is (say) below 150.

Sophist-UK:
As a comparison, here is the same image burned with uni-direction set:
![1684501759554](https://github.com/meerk40t/meerk40t/assets/3001893/3fa725cc-a200-4536-8ff7-c239bdb607e6)

As you can see, the difference in quality is noticeable.

tatarize:
Though I actually expect focus to be such that you more-or-less can see the overlap until 100 DPI or so. 
Even if it did convert to individual lines I would still generally overlap them in simulation there. However, this is an image and remains an image the entire time. The RasterCut actually keeps an image inside a bit of cutcode.

I would have previously agreed that maybe we should convert this over to just a bunch of lines, but newlydraw, for example has a mechanism whereby it converts the data during writing to basically a bitmap. You give it a bitdepth, step-amount, and then a stream of binary and it cuts that in a horizontal or vertical cut. I don't know if this works as nicely if it's already lines.

But, mostly lines are too slow there, and it might need some additional work converting that to geomstr data in 0.9.x to see if it can't be made fast enough to actually render as cuts instead of as a proper raster.

tatarize:
Tiger noted you may be highlighting an off-by-one error. Since you're right that every other pixel seems to go further. This was/is also the case at 1900. And it would be fairly easy to do. Since I'm treating pixels as discrete locations and more properly they should be treated as the inter-pixel locations between the pixels.

[...]

PR #2719, opened 2024-11-07
Simplified raster plotter fixes

jpirnay:
Issue #1921 (and possibly #2330, too) have led to a deep-dive analysis of the RasterOp operation and the RasterPlotter implementation. During this some need for enhancements / clarifications have been identified and implemented:
# GUI-fixes / changes
## Simulation window
The simulation window has received a couple of updates (that were merged earlier to help identify certain things). The context menu allows to deep dive into an image / raster and show the cutcode that MeerK40t has creted for the image. Additionally you can see how wide the to be burned path will become by looking at the device laserspot diameter (a new property that has been introduced):

simulation

You can now see if and how much lines are going to overlap.

## Property Panel
Some inconsistencies around defining the burn direction and start position have been addressed and a new cutoff value has been introduced (see methodology description below).

# The Raster algorithm
- You look at all to be rastered objects and create an image around these, that will have a certain size according to the indicated dpi.
- So a 1in x 1in rectangle with a dpi of 10 will create a 11x11 pixel representation.
- A dpi of 500 will create a 508x508 image
- This image will be scanned by the rasterplotter - So a low dpi is a very bad idea as you are creating an enormous amount of artifacts
- You can see the consequences of a low dpi value inside the preview window:

- By default the created image will be antialiased (so some grayscales will be used where you have a partial overlap of elements with that pixel) - this will create unwanted artifacts at lower dpi values. The newly introduced threshold parameter will let you define 'how black a pixel needs to be ' to be considered. You can leave at the default of zero for any sensible dpi value.
- Have a look at the preview window to see when lines are going to overlap: a too low value will leave gaps, a too high value will become too dark!
- But never ever choose a ridiculously low value - unless you know what are you doing!

## Fixes
- The implementation of the rasterplotter had an issue with bidirectional movement, when an overscan value was present


meerk40t/core/cutcode/rastercut.py
```diff
@@ -122,8 +123,8 @@ def length(self):
             scan_distance = self.width * scan_stride
         else:
             scanlines = self.width
-            scan_stride = self.step_x
-            scan_step = self.step_y
+            scan_stride = self.step_y
+            scan_step = self.step_x
             scan_distance = self.height * scan_stride
```
Sophist-UK, line 128: Personally I think that these lines should be in the opposite sequence so that it makes it easier to compare with lines 121-122 i.e. left side is in the same sequence and right side is reversed for a raster in the opposite direction.

PR #2724, opened 2024-11-08
Fix performance issue at delete many items

pkochubey:
Often, deleting multiple items can cause the app to freeze, and sometimes it takes a long time. With this fix, deletion now works quickly. File for test in attachment. 
heads_all

## Summary by Sourcery

Bug Fixes:
- Resolve performance issue when deleting multiple items by preventing unnecessary selection operations.


jpirnay (approved):
Seems to do the job - thank you :-)

PR #2739, opened 2024-11-18
Fix textctrl colors for overloaded light on dark systems

jpirnay:
## Summary by Sourcery

Bug Fixes:
- Fix text control colors to respect system color settings for better visibility on light-on-dark themes.


jpirnay:
Unittests fail as test rasterplotter change has been committed before

Issue #2762, opened 2024-12-12
[Bug] 0.9.5.3: Inkscape path saved with non-zero fill is loaded as Odd-Even fill


Sophist-UK:
Inscape as saved
![image](https://github.com/user-attachments/assets/eb9b77bb-edb3-473f-adff-b24c1cc17766)

Meerk40t as loaded
![image](https://github.com/user-attachments/assets/9d78c14b-94ee-43b5-a46b-ef9810dbf872)

Meerk40t as it should be loaded
![image](https://github.com/user-attachments/assets/10d03a40-9e31-4f59-8f97-64a77a59a94f)

Note where the i/b/r overlaps. (This is a combined path with individual letters as subpaths. Note the counter direction inner paths for e.g. the "a" are working as intended.) 

When saved by MK and reloaded, Non-Zero is loaded as saved - it seems to be an Inkscape issue.

[Libra inkscape.svg](https://github.com/user-attachments/assets/bdf9d806-b0dc-46db-8f71-44cc5276358a)

jpirnay:
Resolved in PR 2763

PR #2763, opened 2024-12-13
Set proper SVG defaults on svg load

jpirnay:
Revise the default values for fill and stroke attributes according to SVG defalult definitions.
Addresses #2762

## Summary by Sourcery

Bug Fixes:
- Correct default values for fill and stroke attributes in SVG files to align with SVG specifications.


Sophist-UK:
Sorry Jens but you missed the unindenting of what is now line 893.

PR #2777, opened 2025-01-04
Custom commands bugfix

FredrikDietmar:
In the on_text_commands function custom_commands is set to None if text is empty. So the get function will return None instead of using the default value.
Therefore the value has to be checked if it is None before usage.
Another solution could be to instead let it get set to an empty string,

## Summary by Sourcery

Bug Fixes:
- Handle the case where the "custom_commands" setting is empty, preventing potential issues.


jpirnay (changes requested):
Thank you for your contribution. One minor change to be made, please: the walrus operator if a := b : 
was introduced with python 3.8. We intentionally try to remain on the python 3.6 level as this is the last supported version for a couple of relevenat older systems. So please split the assigment & if statement

PR #2905, opened 2025-06-08
Add layer name match for engrave/cut (like whisperer)

jpirnay:
## Summary by Sourcery

Add support for classifying DXF entities into engrave/cut operations based on their layer names and emit an update signal after classification.

New Features:
- Automatically assign DXF entities to engrave or cut operations when their layer name contains “engrave” or “cut”.

Enhancements:
- Emit an element_property_update signal after new elements are classified.


jpirnay:
@sourcery-ai review

PR #2956, opened 2025-07-31
PT_BR and ES translation updates

ristraus:
Added 15 new missing translation entries to messages.po.

I noticed they are missed in all languages.

Minor corrections plus added new missing translations to PT_BR plus ES


jpirnay (commented):
Ok

PR #2963, opened 2025-08-07
Aggiornamento it75a traduzione italiana e aggiunta file spec/pip

Betaeta84:


listPip.txt
jpirnay, line 1: I am not sure what purpose this file will serve - involuntary addition?

meerk40t.spec
jpirnay, line 1: this file is not fit for purpose for anyone else:
a) it contains fixed paths to your environments (d:\utenti\EtaBeta...)
b) run meerk40t.exe and look in "About" which libraries are included and which are missing - I assume there are several.

jpirnay (changes requested):
- Invalid file added (?)
- Fixed paths in meerk40t.spec
- Library imports unclear

PR #2981, opened 2025-08-24
Updated Translations ES,NL and pt_BR

ristraus:
Updated files:
messages.po with the new lines from DE
meerk40t.po and meerk40t.mo for ES, NL and pt_BR
all other languages meerk40t.po with the missing translation lines from messages.po, but NOT updated meerk40t.mo, since they need to be translated.


jpirnay:
I recommend using either the ``translate.py`` tool in meerk40ts root directory or at least ``.github\workflows\translate-error-check.py`` (this will be executed as test during .po commits and has reported this error, hence the failed status of your PR)

PR #3000, opened 2025-09-01
Angle div fix for /=

jpirnay:
## Summary by Sourcery

Enable robust in-place division and comparison for Angle objects

Bug Fixes:
- Add __itruediv__ to support true in-place division ('/=') for Angle instances
- Modify __eq__ to catch invalid type conversions and return False instead of raising errors


jpirnay:
Woohoo - 3.000 PRs !

tiger12506:
Github uses the same number pool for both issues and pull requests.... love the enthusiasm though!

PR #3032, opened 2025-10-01
Ruida - File rejection fixes.

StevenIsaacs:
This PR corrects two problems with the Ruida driver which were causing the Ruida controller to reject files:
  - The "en_laser_tube_start" method now accepts a parameter to indicate whether the tube should be enabled. A new encoder, "encode_switch", was added to support this.
  - The file EOF marker is now being included in the file checksum.

A version requirement for scipy was removed because of problems when installing on Fedora 41.

## Summary by Sourcery

Fix Ruida driver file rejection issues by updating the laser tube start command and including the EOF marker in the checksum, and relax the SciPy version requirement on Linux.

Bug Fixes:
- Ensure the EOF marker is included in the file checksum to prevent Ruida file rejections
- Allow the en_laser_tube_start command to accept an explicit on/off parameter

Enhancements:
- Add a new encode_switch encoder function to support switch commands

Build:
- Remove the SciPy version constraint for better compatibility on Fedora 41


requirements-optional-linux.txt
```diff
@@ -3,6 +3,6 @@ pypotrace
 pyclipr
 meerk40t-barcodes
 numba
-scipy>=1.12.0,<=1.14.0
```
jpirnay, line 6: This restriction is intentional as higher versions of scipy cause other problems. Please reinstate the restrictions

jpirnay (changes requested):
Please resolve the optional scipy import statement.

requirements-optional-linux.txt
StevenIsaacs: Will do.

jpirnay (approved):

PR #3037, opened 2025-10-04
Correct detection of run environment.

StevenIsaacs:
Corrected the path for detection of the run environment so will be displayed correctly on the title bar. The sys.path[0] is typically equal to an empty string which previously made the tested paths relative to the root directory.  This change makes the paths relative to the current directory.

## Summary by Sourcery

Bug Fixes:
- Use os.path.join instead of string concatenation when checking for .git and .github directories to avoid root-relative path errors


StevenIsaacs:
I did consider using abspath but shy away from it unless absolutely necessary because I've seen instances where the results find their way into config files and limiting portability as a result. Yes, not likely in this case but best to minimize risk (imho). 

On the other hand, the pathlib suggestion may make sense. I don't have time at the moment to experiment with it to see what the results would be. At first look it appears to me to be more code to accomplish the same thing in this case.

jpirnay (approved):

PR #3105, opened 2025-11-23
GRBL: Added config class for USB which allows direct entry of any serial port device.

StevenIsaacs:
This PR adds the ConfigurationUsb class to allow direct entry of any serial port device. This helps avoid problems where some Linux distros create different ttyUSBx devices for the same device which frequently happens when turning a laser off and then back on.  UDEV rules can now be used to create symlinks which have a consistent and meaningful name. This also enables use of serial port sniffers which create virtual serial ports.

NOTE: This is largely a cut and paste from the ruida gui. Perhaps the class can be added to the core for common use?

## Summary by Sourcery

Introduce a dedicated USB configuration panel for GRBL to allow direct entry of serial port devices and integrate it into the GRBL configuration interface.

New Features:
- Add a ConfigurationUsb panel that lets users specify the serial port device directly, supporting stable device naming via symlinks on Linux.

Enhancements:
- Replace the generic serial ChoicePropertyPanel in the GRBL interface configuration with the new USB configuration panel for a more focused serial setup experience.


jpirnay:
This is the 'wrong' approach - the code already has a linux logic - should provide a dropdown combox that allows to enter arbitray texinput to define the comport, too.

jpirnay:
There was a bug preventing this, but that should be fine now

StevenIsaacs:
> There was a bug preventing this, but that should be fine now

You've rendered this PR useless. Your fix works.

StevenIsaacs:
No longer needed after jpirny fix.

PR #3134, opened 2025-12-28
Safeguard undo index logic

jpirnay:
## Summary by Sourcery

Tighten undo/redo index handling and thread safety in the core undo service and add regression tests for the new behavior.

Enhancements:
- Make the undo service lock re-entrant and make validate() thread-safe while ensuring LAST_STATE snapshots are only added when there is real history and stack size limits are enforced.
- Adjust undo(), redo(), redo_string(), and states() to compute target indices safely around validate(), avoid off-by-one errors, and keep menu state lists consistent with the current index.

Tests:
- Add comprehensive tests for undo/redo index management across marking, branching history, direct index operations, and validate() behavior.
- Add tests for the states() helper to ensure undo and redo state listings correctly exclude the current state and align with redo() behavior.


jpirnay:
@sourcery-ai dismiss

PR #3158, opened 2026-01-09
Make locale inclusion more flexible

jpirnay:
## Summary by Sourcery

Make locale inclusion for packaged builds dynamic based on available translation files instead of hardcoded paths.

Enhancements:
- Generate PyInstaller locale data entries programmatically in Windows and Linux spec files.
- Construct macOS PyInstaller --add-data arguments at build time by scanning existing locale directories.


jpirnay:
Too much black magic, which paths to take, as the path base generated in the spec files is not identical to the one used by pyinstaller?! For a rainy day

PR #3221, opened 2026-04-29
Working fix for GetTextExtent issue that was affecting layout of ribbon

tiger12506:
I can't really explain why this code works better than what we had previously.
For whatever reason, when we used the dc derived from the Bitmap from _set_buffer() in layout(), it would provide inconsistent results for calls to GetTextExtent().

It's not even necessary for the Bitmap to be anything but wx.Bitmap(1, 1) if we're just calling GetTextExtent on it.

Stored the text_bmp = wx.Bitmap(1, 1) as a class variable so that there is only one copy for all ribbon, and used SelectObjectAsSource to indicate to wxwidgets that this is a readonly binding of the dc to the bitmap (I presume this is more efficient gdi object count/memory-wise)

## Summary by Sourcery

Improve ribbon text measurement to avoid inconsistent GetTextExtent results during layout.

Enhancements:
- Use a shared 1x1 bitmap and dedicated helper method to measure text extents for ribbon labels instead of relying on the layout DC.
- Prepare ribbon buffer creation for future DPI-aware rendering by routing bitmap construction through a local variable.
- Reduce layout noise by commenting out a debug print of the ribbon size.

### DIFF
--- a/meerk40t/gui/ribbon.py
+++ b/meerk40t/gui/ribbon.py
@@ -1162,7 +1162,12 @@ class RibbonBarPanel(wx.Control):
                 width = 1
             if height <= 0:
                 height = 1
-            self._ribbon_buffer = wx.Bitmap(width, height)
+            bmp = wx.Bitmap(width, height);
+
+# Tiger note: This will be helpful when making ribbon DPI aware but Meerk40t isn't ready for that quite yet.
+#            bmp.CreateWithDIPSize(width, height, self.GetDPIScaleFactor())
+
+            self._ribbon_buffer = bmp
         return self._ribbon_buffer
 
     def toggle_show_labels(self, v):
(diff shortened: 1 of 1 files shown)


tiger12506:
> ## Overall Comments
> - Since `get_text_extent` may be called frequently during layout, consider reusing a single `wx.MemoryDC` instance instead of creating a new one on each call to reduce overhead and GDI object churn.
> - Using `Art.text_bmp` as a class-level bitmap from an instance `__init__` is a bit surprising; either initialize it in a clear class-level initialization path or make it an instance attribute to clarify ownership and lifetime.

Ignore these Sourcery suggestions, they don't work because the AI doesn't understand that the Art object class variables would be created outside of the GUI thread and therefore immediately crash Meerk40t.

PR #3227, opened 2026-05-14
Fix segfault in tree widget during file load

jondale:
Meerk40t would occasionally segfault when loading a file.  The issue was a call to GetNextChild at the same time the tree was changing.  This fix is to create a list and then loop it instead of GetNextChild.

## Summary by Sourcery

Bug Fixes:
- Avoid a segmentation fault during file load by iterating over a snapshot of tree children instead of traversing a mutating tree.


jondale:
This was already addressed by another pr

PR #3230, opened 2026-05-16
refactor wordlist component for DRY

Johann150:
Refactors the code in accordance with DRY principle and while doing that removes an undefined reference to the `brackets` variable which should probably have been `_BRACKETS` instead.


Johann150:
Rebased and altered my PR since the wrong variable name was already fixed in #3233.

PR #3250, opened 2026-07-11
Fix numpy 2.x removal of 2D np.cross in geomstr (Outline button crash)

George-RD:
## What

`np.cross` for 2D vectors was removed in numpy 2.0 ([numpy/numpy#26443](https://github.com/numpy/numpy/releases/tag/v2.0.0) deprecation completed). `geomstr.py` has four 2D call sites that now raise `ValueError: Both input arrays must be (arrays of) 3-dimensional vectors`:

- quickhull signed distance (`convex_hull`) — crashes the GUI **Outline** button for any device without a native `outline()` (e.g. GRBL), via `element* trace hull`
- point-line distance 2D branch (`_douglas_peucker`)
- parallel-side check
- polygon angle check

All four compute the scalar z-component, so they are replaced with an explicit `_cross2d` helper (`a[...,0]*b[...,1] - a[...,1]*b[...,0]`), which is also what numpy's own release notes recommend.

## Why

With numpy 2.x installed (the default for any fresh `pip install meerk40t` environment today), pressing **Outline** in the Laser panel with a GRBL device crashes. Found live-driving a Sculpfun S9 on macOS / Python 3.12 / numpy 2.x.

## Testing

`python -m unittest test.test_geomstr` on current main with numpy 2.x: **10 errors** (all `np.cross` 2D).
With this patch: **61 tests, OK** — zero errors, no other behaviour change.

Also verified the exact crash reproduction (convex hull over three rects from the GUI trace path) returns a correct 5-point hull.

## Summary by Sourcery

Replace deprecated 2D np.cross usage in geometric computations with a dedicated helper to restore compatibility with NumPy 2.x and prevent GUI outline crashes.

Bug Fixes:
- Fix convex hull signed-distance calculation to avoid runtime errors with NumPy 2.x when computing outlines.
- Fix 2D point-to-line distance calculation to work correctly without np.cross for 2D vectors.
- Fix parallel-side detection and regular polygon angle checks that relied on 2D np.cross, eliminating NumPy 2.x ValueErrors.

Enhancements:
- Introduce a reusable _cross2d helper for computing the 2D cross-product z-component in geometric routines.

### DIFF
--- a/meerk40t/core/geomstr.py
+++ b/meerk40t/core/geomstr.py
@@ -72,6 +72,15 @@ from copy import copy
 
 import numpy as np
 
+
+def _cross2d(a, b):
+    """z-component of the 2D cross product (np.cross for 2D vectors was
+    removed in numpy 2.0)."""
+    a = np.asarray(a)
+    b = np.asarray(b)
+    return a[..., 0] * b[..., 1] - a[..., 1] * b[..., 0]
+
+
 # Import numba for JIT compilation optimization
 try:
     from numba import njit
@@ -7175,7 +7184,7 @@ class Geomstr:
         """
 
         def process(S, P, a, b):
-            signed_dist = np.cross(S[P] - S[a], S[b] - S[a])
+            signed_dist = _cross2d(S[P] - S[a], S[b] - S[a])
             K = [i for s, i in zip(signed_dist, P) if s > 0 and i != a and i != b]
 
             if len(K) == 0:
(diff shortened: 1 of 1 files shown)


jpirnay:
Thank you for your contribution. Probably already resolved since the integration of https://github.com/meerk40t/meerk40t/pull/3247 that had the same intent.

George-RD:
Closing as superseded by #3247, which has now merged the equivalent NumPy 2.x compatibility fix covering all affected geomstr 2D np.cross call sites. Thanks for pointing this out.

PR #3260, opened 2026-08-12
Fix newly driver bed dimension type

### LINKED ISSUE #3259: newly driver crashes on jog near bed edge - bedwidth/bedheight typed as str instead of Length
Jogging near the far edge of the bed crashes with:

TypeError: unsupported operand type(s) for -: 'str' and 'float'
  File "meerk40t/newly/driver.py", line 395, in move_rel
    self.service.view.height

Locals: dx -10.0mm, dy 10.0mm, confined True, new_x 50.0456mm,
new_y 200.1444mm. Bed set to 220x200mm. 0.9.9100 on Ubuntu,
Python 3.12.3, wxPython 4.2.1, Artsign JSM-40U over USB.

newly/device.py declares bedwidth and bedheight with "type": str.
grbl, ruida, moshi and lihuiyu all use "type": Length. View stores
width/height raw with no coercion, so on newly the value stays a
string all the way to the clamp in move_rel, and
view.height - native_y*scale is str - float.

The comparison a few lines above it works because Length.__gt__
coerces the string, so it only fails on the assignment. The < 0
branches never touch width/height, which is why it only crashes
near the far edge and not the near one.

Editing the field to remove "mm" does not help, the value is still
a str.

Same clamp code exists in the other four drivers but they are
immune because their bed dims are Length.

Still present on main.

KenTread:
Fixes #3259

newly/device.py declared bedwidth and bedheight as "type": str.
Every other driver uses "type": Length. View assigns width and
height without coercion, so on newly the raw string reached the
confined-jog clamp in driver.py move_rel and blew up on
str - float.

Changed the two declarations to Length to match grbl, ruida, moshi
and lihuiyu. No driver.py change needed, and it fixes any other
consumer of view.width/view.height in this driver rather than just
the one call site.

Root cause was confirmed by reading the source at 0.9.9100. The
fix itself is not yet hardware-tested. I have the affected machine
(Artsign JSM-40U over USB) and will confirm shortly that jogging
into the far edge clamps instead of crashing, and that an existing
config storing the value as a bare string still loads. I will
follow up on this PR with the result.

## Summary by Sourcery

Align newly driver bed dimension types with other drivers and bump application version.

Bug Fixes:
- Prevent crashes in the newly driver by declaring bedwidth and bedheight as Length instead of raw strings.

Build:
- Update application and package version metadata from 0.9.9000 to 0.9.9100, including Windows file version.

### DIFF
--- a/meerk40t/newly/device.py
+++ b/meerk40t/newly/device.py
@@ -138,7 +138,7 @@ class NewlyDevice(Service, Status):
                 "attr": "bedwidth",
                 "object": self,
                 "default": "310mm",
-                "type": str,
+                "type": Length,
                 "label": _("Width"),
                 "tip": _("Width of the laser bed."),
                 # Hint for translation _("General")
@@ -152,7 +152,7 @@ class NewlyDevice(Service, Status):
                 "attr": "bedheight",
                 "object": self,
                 "default": "210mm",
-                "type": str,
+                "type": Length,
                 "label": _("Height"),
                 "tip": _("Height of the laser bed."),
                 # Hint for translation _("General")


jpirnay (approved):

Sophist-UK:
Welcome back Jens - everyone has missed you!!!

Issue #3268, opened 2026-08-26
[Bug]: the ubuntu version and the linux mint version both give an error when launching


muezabdalla:
### Summary Description

when I try to run the linux mint version I get this error:
```
(Meerk40-linux-mint20:23047): dbind-WARNING **: 16:03:03.098: Couldn't register with accessibility bus: Did not receive a reply. Possible causes include: the remote application did not send a 
reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
Gdk-Message: 16:03:03.307: Unable to load hand2 from the cursor theme

(Meerk40-linux-mint20:23047): Gtk-WARNING **: 16:03:03.352: Could not load a pixbuf from /org/gtk/libgtk/icons/16x16/status/image-missing.png.
This may indicate that pixbuf loaders or the mime database could not be found.
**
Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecognized imag
e file format (gdk-pixbuf-error-quark, 3)
Bail out! Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecog
nized image file format (gdk-pixbuf-error-quark, 3)
Aborted./Meerk40-linux-mint20
```
and I also tried to use the ubuntu version but got this error:
```
Traceback (most recent call last):
  File "mk40t.py", line 11, in <module>
  File "meerk40t/main.py", line 201, in run
  File "meerk40t/main.py", line 252, in _exe
  File "meerk40t/kernel/kernel.py", line 1171, in __call__
  File "meerk40t/kernel/kernel.py", line 694, in set_kernel_lifecycle
  File "meerk40t/gui/plugin.py", line 40, in plugin
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "PyInstaller/loader/pyimod02_importers.py", line 457, in exec_module
  File "wx/__init__.py", line 12, in <module>
AttributeError: partially initialized module 'wx' has no attribute '__version__' (most likely due to a circular import)
[...]

PR #3270, opened 2026-09-05
wx: Fix linux reparent errors with AddPage calls

frmdstryr:
This fixes the various errors seen in #3268 by calling `panel.Reparent(notebook)` for every usage of `AuiNotebook.AddPage`. 

I'm not sure if this is linux only or something with a more recent wx (I currently have 3.3.3). This makes the main UI and all the dialogs work for me.

## Summary by Sourcery

Fix AUI notebook initialization by reparenting all page panels before adding them.

Bug Fixes:
- Fix Linux and wxPython compatibility errors when adding panels to AUI notebooks.

Enhancements:
- Ensure notebook page panels are explicitly parented to their target notebooks throughout the application.

### DIFF
--- a/meerk40t/balormk/gui/balorconfig.py
+++ b/meerk40t/balormk/gui/balorconfig.py
@@ -118,6 +118,7 @@ class BalorConfiguration(MWindow):
                     injector=injection,
                 )
                 self.panels.append(newpanel)
+                newpanel.Reparent(self.notebook_main)
                 self.notebook_main.AddPage(newpanel, pagetitle)
 
         # newpanel = EffectsPanel(self, id=wx.ID_ANY, context=self.context)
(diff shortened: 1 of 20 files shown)


tiger12506 (approved):
This bug only affects wxPython >= 4.3.x (~wxWidgets 3.3.3) on both Linux and Windows.

I don't know why wx now requires this reparenting, but I confirm the bug, and I confirm the fix. The original issue #3268 has OP confirming this as a fix as well.

I tested against wxPython 4.3.1 (as in the bug report), and also wxPython 4.2.5 (not affected by bug) and both open and panes seem to operate correctly in this branch in both Linux and Windows.

Curiously, on Windows it outputs messages to stderr in this branch (including attached), but meerk40t running correctly with warnings is better than meerk40t crashing.

[error.txt](https://github.com/user-attachments/files/32149125/error.txt)

